PR #1743, opened 2015-01-04
Update cl_init.lua

cardentity:
the same effect in "for"


FPtje:
Negligible optimisation

PR #1747, opened 2015-01-07
Converted Server Settings to use the MySQLite Library

TheEMP:


FiG-Scorn:
Hmm, I can see some potential harm from this, especially for server communities that run multiple servers connected to one MySQL database.

TheEMP:
I can kinda see where your coming from but I think on a whole most server owners would expect this type of thing to be synced cross server. The only other thing that is not synced is the restrict weapon settings, Bans, Ranks, Groups everything in the darkrp is sycned. I really dont see a reason why Server settings wouldn't be.

FPtje:
The reason to have multiple servers is to run them with different settings. This change makes it impossible to have servers run with different settings. I specifically decided this when making it.

PR #1784, opened 2015-02-15
Added link to darkrp forum in README.MD

HDG-MrDeeJay:
I wonder if this may help reduce the number of times Falco has to reply
with "Not a help forum".

- Added a link to the forum to the README.MD file, as well as a reminder
that the github is NOT a place to request help, whereas the forum IS.


FPtje:
Good idea!

PR #1810, opened 2015-03-04
FAdmin/ServerGuard user group compatibility.

impulsh:


HDG-MrDeeJay:
Mind telling us /how/ this adds compatibility?

alexgrist:
The founder rank is ServerGuard's highest rank and is assigned to server owners. When trying to use admin commands in DarkRP with a founder rank it does not work correctly. So FAdmin needs to have the founder rank added to it when ServerGuard is installed unless FAdmin switches over to using the IsSuperadmin and IsAdmin functions.

HDG-MrDeeJay:
Fair enough, seems reasonable

FPtje:
Sorry man, but I don't agree with this solution. It's too specific for ServerGuard.   Besides, when fixing compatibility problems I don't want references to other addons to be spread across arbitrary core files. That's because it really hurts coupling.

alexgrist:
I understand that but what other solution is there? The issue isn't with ServerGuard's rank setting, it's with your FAdmin not using the correct IsSuperadmin or IsAdmin functions. You're relying on specific user group names instead of relying on a general permission set.

alexgrist:
Could we at least get this implemented until something like CAMI is completed as a temporary solution? We have customers to use DarkRP with ServerGuard and us having to somehow interfere with FAdmin feels a lot worse than an if statement in FAdmin.

FPtje:
If those leaders are admins, then that is used by FAdmin. That shouldn't
cause too much trouble.

Adding the leader rank to FAdmin is a guess solution in that I still don't
know what the problem is precisely and what exactly causes it.

impulsh:
Owners of the server have a founder rank - IsAdmin and IsSuperAdmin returns true, but IsUserGroup returns "founder". I'm assuming this issue occurs because FAdmin doesn't know that the founder usergroup is worth anything, so it doesn't allow certain admin actions. An example; setting a door group using F2.

HDG-MrDeeJay:
I guess as an ultimate solution, you could just have 
IsSuperAdmin/IsAdmin override FAdmin, unless a suitable usergroup can be 
found in FAdmin.... Which means on a server-by-server case, you CAN 
create this founder rank for better compatibility, but it would still 
work out of the box... and would provide adequate compatibility with 
most administrative addons.

Bo98:
IsAdmin and IsSuperAdmin isn't the issue here. It's FAdmin.Access.PlayerHasPrivilege which does not fallback to ply:IsAdmin() when the group does not exist.

HDG-MrDeeJay:
Which is exactly what I was saying (but from another perspective), put 
that fallback in, and insta-compatibility everywhere.

Neosinos:
I've always wondered why FAdmin sometimes uses the isAdmin Fallback and
sometimes doesnt.

Am Mittwoch, 4. März 2015 schrieb Nathan Davies-Lee :

alexgrist:
In ServerGuard, we have a "Superadmin" permission which makes the rank return true with IsSuperadmin, same with admin of course. So yes, falling back to those functions is pretty much all that would be required.

FPtje:
I'll do that. Until then, this is not a solution.

PR #1832, opened 2015-03-18
Allow multiple chat prefixes

Kefta:
Both "!dropmoney" and "/dropmoney" work

### DIFF
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -266,7 +266,7 @@ GM.Config.moneyModel = "models/props/cs_assault/money.mdl"
 -- The skin DarkRP uses. Set to "default" to use the GMod default derma theme
 GM.Config.DarkRPSkin = "DarkRP"
 GM.Config.currency = "$"
-GM.Config.chatCommandPrefix = "/"
+GM.Config.chatCommandPrefix = { "/", "!" }
 GM.Config.F1MenuHelpPage = "http://wiki.darkrp.com/index.php/Main_Page"
 GM.Config.F1MenuHelpPageTitle = "Wiki page"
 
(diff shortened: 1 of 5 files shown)


HDG-MrDeeJay:
Um, we already have a means to change chat prefixes in the modification 
file, and wouldn't this affect other addons that use commands of the 
same name but use !.

I think your pull request is harmful to DarkRP.

Kefta:
Changing != multiple prefixes. This allows for unified ! and / commands. Also, if an addon is using a same command name with just a different prefix, then that's just bad coding.

Kefta:
Changing != multiple prefixes. This allows for unified ! and / commands. Also, if an addon is using a same command name with just a different prefix, then that's just bad coding.

HDG-MrDeeJay:
no, because FAdmin uses /ban. ULX uses !ban. What you propose would make 
FAdmin use both !ban and /ban.

FPtje:
This breaks DarkRP on servers that don't immediately update darkrpmod, which is basically every server. 

It breaks every addon that uses this config option. 

Behaviour is not defined when the chat prefix table is empty. End users won't know this and they will have no way of figuring out what's wrong when they see the Lua errors. 

This pr can only cause trouble and has no real gain. 


Having multiple chat command prefixes has a negative effect on usability. There can be confusion about the multiple chat commands, spawning the question "which is better, /dropmoney or !dropmoney?". That while it doesn't make anything easier for anyone. Everyone is already used to just /.

Kefta:
If I remove ! from the default table, can support for multiple chat prefixes still be added?

Bo98:
Still need to make it backwards compatible with old configs that set a string and not a table.

FPtje:
It still wouldn't solve most of the problems I mentioned. Why do you want this anyway?

PR #1839, opened 2015-03-22
Added support for destroying func_breakable_surf

deinemudda32:
support for destroying func_breakable_surf for probably police raids :)

### DIFF
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -102,6 +102,7 @@ function SWEP:DoFlash(ply)
 end
 
 local entMeta = FindMetaTable("Entity")
+
 function SWEP:DoAttack(dmg)
 	if CurTime() < self.NextStrike then return end
 
@@ -117,7 +118,7 @@ function SWEP:DoAttack(dmg)
 	if IsValid(vm) then
 		vm:ResetSequence(vm:LookupSequence("idle01"))
 		timer.Simple(0, function()
-			if not IsValid(self) or not IsValid(self.Owner) or not IsValid(self.Owner:GetActiveWeapon()) or self.Owner:GetActiveWeapon() ~= self then return end
+			if !IsValid(self) or !IsValid(self.Owner) or !IsValid(self.Owner:GetActiveWeapon()) or self.Owner:GetActiveWeapon() ~= self then return end
 			self.Owner:SetAnimation(PLAYER_ATTACK1)
 
 			if IsValid(self.Weapon) then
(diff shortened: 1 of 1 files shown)


HDG-MrDeeJay:
so stuff like custom(shitty) windows that take more than a love-tap to 
break?

deinemudda32:
what do you mean?
func_breakable_surf are standard half life 2 windows, seen in alot of RP maps.

FPtje:
Why the C syntax and the mix between spaces and tabs though?

deinemudda32:
I like it tidy.

FPtje:
Well I like it consistent with the rest of DarkRP.

deinemudda32:
But how is the idea to make func_breakable_surf destroyable? ._.

FPtje:
Is it possible in another way than bullets?

HDG-MrDeeJay:
Terrible. If its standard HL2 windows, they should be breakable anyway 
just by tapping it. They are brittle as fuck.

deinemudda32:
I tried alot in singleplayer but nothing seemed to work to get it breakable.
If you got another way, tell me :)

HDG-MrDeeJay:
Yes. Props, a fast moving player, vehicles.

deinemudda32:
I did mean with the weapon?
Should the weapon throw a washing machine or what?
Or a moving player or a jeep?

FPtje:
You'd expect a crowbar to hit a hole in the glass, but not a stunstick. The
stunstick is electricity based. There should be some way to break the glass
without poking a hole in it.

HDG-MrDeeJay:
I'd still expect a stunstick to break the glass

You gotta remember, Valve designed breakable glass to not only represent the sheer power that the HEV suit can output in strength (in comparison to a standard human), but also to represent old, brittle glass. This is why I'm more partial to map developers making their own (transparent) entities that are breakable, to act as glass. Can be taken out one hit without fracturing, can be moved around without hassle, and is easier to work with from the lua end.

deinemudda32:
Maybe a stunning impact?

HDG-MrDeeJay:
Wait just to be clear, we are talking about the same windows right? Those ones that are used in Downtown_v4c all the time, that respawn (even if not needed) every 5 minutes?

FPtje:
Break the glass with its electricity, not with its impact.

Sometimes when you throw a prop near a window, the glass shatters, but it's
still in the window without holes. Kind of like fractured car glass. That
would look a lot cooler than a hole in the glass.

HDG-MrDeeJay:
>Maybe a stunning impact?

Nah, think more like you take the handle of your batton, and apply direct force through the hilt into the center of the window.

HDG-MrDeeJay:
I don't think I'm on the same page as everybody

FPtje:
There's no animation for that.

deinemudda32:
Maybe not skin it blue.
I personally think, thats the main case that it doesnt look and feel like a stun baton

FPtje:
We weren't discussing whether it looks like a stunstick.

deinemudda32:
Yeah right, sorry but I hate that.
Im trying in singleplayer to get this solved.

Bo98:
func_breakabale_surf has an input called "Shatter" that takes a Vector.

FPtje:
Nice

deinemudda32:
Got it working:
So how do I post code?

FPtje:
This is what I meant

FPtje:
Also, you're changing *my* code in the pull request. Get your damn act together.

PR #1847, opened 2015-03-24
Fix for #job.model == 1 showing model choice

Partixel:
If the job.model is a table with only 1 model the model choice gui part thing won't show.


Partixel:
Reasons: Making a job maker gui thing which takes models as a list of strings so it always is a table.

When the table only contains one model, the F4 menu still shows the model choice part when selecting the job because it is a table.

Partixel:
In case it's not clear what the model choice part thing is:
http://i.imgur.com/vJ3cZYI.png

PR #1872, opened 2015-04-01
Added 'LockpickTime' hook to Lockpick

smithy285:
This allows developers to easily modify the lockpick time.


FPtje:
- Hooks start lowercase.
- "The time an entity can be lockpicked" makes it sound like people are supposed to return a clock time.
- The override return value documentation is wrong

smithy285:
@FPtje Updated the pull request. My apologies for not getting it right the first time

Bo98:
The return val description still isn't right.

smithy285:
@Bo98 Third time lucky?

PR #1873, opened 2015-04-02
Syntax fix

aStonedPenguin:
uniformityIs aGreatThing

### DIFF
--- a/gamemode/cl_init.lua
+++ b/gamemode/cl_init.lua
@@ -44,4 +44,4 @@ DarkRP.DARKRP_LOADING = nil
 
 DarkRP.finish()
 
-hook.Call("DarkRPFinishedLoading", GM)
+hook.Run("DarkRPFinishedLoading", GM)


Kefta:
Remove the GM argument

aStonedPenguin:
ohShit iForgot

HDG-MrDeeJay:
gg

Bo98:
If you want to do this then at least do it in init.lua too.

Bo98:
Doesn't this break the GM function though? While hook.Run DOES call the GM function, I believe it doesn't do this until GAMEMODE exists.

Partixel:
hook.Run != hook.Call, how is this a syntax fix?

FPtje:
Hook.Run somehow didn't work. It should, but it didn't. I don't know why, but I went "fuck it" and used hook.Call.

Bo98:
Pretty sure it's because hook.Run uses GAMEMODE which is nil here.

FPtje:
Yeah gmod.GetGamemode or something. Hook.Call it is.

PR #1879, opened 2015-04-05
Set default player scale to 72, not 64

DaaOtt:
This is GMod's default, prediction will work now

### DIFF
--- a/gamemode/modules/playerscale/cl_playerscale.lua
+++ b/gamemode/modules/playerscale/cl_playerscale.lua
@@ -4,6 +4,6 @@ local function doScale(um)
 
 	if not IsValid(ply) then return end
 	ply:SetModelScale(scale, 1)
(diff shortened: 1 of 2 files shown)


FPtje:
I remember it always being 64. Where are you getting this from?

DaaOtt:
https://developer.valvesoftware.com/wiki/Dimensions#Overhead_Obstacle_Height

further proof incoming

FPtje:
Second question, why does prediction break when it's not on the default value?

DaaOtt:
https://a.pomf.se/gisfob.mp4

HDG-MrDeeJay:
oh boy, this bug.

DaaOtt:
Sandbox: https://a.pomf.se/rnnimg.mp4

FPtje:
w/e

PR #1884, opened 2015-04-07
SWEP changes

Bo98:
* Lots of prediction fixes.
  - The UpdatePlayerSpeed, canDoorRam, canLockpick and lockpickTime hooks are now shared.
  - The recoil is now less jerky and comes back down after a short period.
  - The only prediction error now is when door ramming a vehicle with a driver as Vehicle:GetDriver() unfortunately only exists serverside.
* Refactored arrest, unarrest and stunsticks to a common stick_base.
  - A lot less code duplication now.
* Added stunstick killicon.
* Relaxed SWEP.IconLetter restrictions to any valid character in the font.
  - Used to be restricted to weapons that exist by default in DarkRP.
  - Now is expanded to [0-9a-wA-Z].
  - Lowercase x-z are not valid characters in the font.
* Increased Deagle and MP5 recoil; decreased MP5 damage.
  - The Deagle had next to no recoil because of the large delay between shots. This didn't matter too much before since the recoil did not come back down but since it does now then this is in need of an increase.
  - The MP5 had its next-to-zero recoil increased slightly. The damage was decreased slightly as it's still one of the most accurate weapons in DarkRP.
* Simplified ls_sniper HUD code.
  - A lot less code duplication.

### DIFF
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -4,86 +4,18 @@ if CLIENT then
 	SWEP.PrintName = "Arrest Baton"
 	SWEP.Slot = 1
 	SWEP.SlotPos = 3
-	SWEP.DrawAmmo = false
-	SWEP.DrawCrosshair = false
 end
 
-SWEP.Base = "weapon_cs_base2"
+DEFINE_BASECLASS("stick_base")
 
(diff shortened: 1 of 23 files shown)


Bo98:
These diffs are not going to be very easy to look at.

PR #1888, opened 2015-04-08
Update sv_database.lua

Acecool:
Removes Vector.__tostring requirement so the Vector tostring function can be altered to output more friendly output instead of "0 0 0" which could be confused with an angle which outputs the same way...


FPtje:
Could you fix the style inconsistencies? I don't put spaces between parentheses.

PR #1905, opened 2015-04-20
Change several libraries to git submodules

Partixel:
Not necessary but nice to have, no longer have to manually merge changes.'


Bo98:
Still have to update the commit ID though.

Partixel:
Nevermind, apparently github doesn't support submodules when downloading the zip, nice github, nice :(

Bo98:
Yeah that was going to be my other question.

Partixel:
I could've sworn it worked in the past with addons like evolve but apparently not.

FPtje:
There are more reasons why I avoid submodules. Subtree merging of fpp is
impossible with it and you have to give a special parameter to git when
cloning.

Subtree merging all the way bitches.

Partixel:
All very good points, now excuse me while I go change all of my projects away from submodules...

-cries-

HDG-MrDeeJay:
drama queen

Partixel:
That may or may not be true, @HDG-MrDeeJay , I guess we'll never know.

PR #1906, opened 2015-04-20
Fixed some singleplayer issues with the batons

Bo98:
Singleplayer is a horrible, horrible place. I would add:

```lua
if game.SinglePlayer() then error("Use a listen server.") end
```

to init.lua if I had my own way.

Fixes http://forum.darkrp.com/threads/baton-glitch.4924/.

### DIFF
--- a/entities/weapons/stick_base/shared.lua
+++ b/entities/weapons/stick_base/shared.lua
@@ -63,7 +63,7 @@ function SWEP:Deploy()
 	BaseClass.Deploy(self)
 	if SERVER then
 		self:HookStartCommand()
-		self:CallOnClient("HookStartCommand")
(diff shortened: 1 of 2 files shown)


Bo98:
The unarrest_stick part might fix similar reports on servers but god knows how that code is even allowed to run when it is removed on Holster.

But trust someone to find a case that should only happen when the stars align. The stars must align every day for them.

HDG-MrDeeJay:
reminds me of a friend who could find a way to break any game.

FPtje:
Honestly, I don't care much for SinglePlayer either, but completely disabling it is a step too far.

Bo98:
Yeah, I wasn't serious there.

PR #1911, opened 2015-04-22
Update sh_init.lua

XRAYBlake:
Add reasons for Hits.


FPtje:
So you add a darkRPVar for a hit reason but no way to choose that reason?

FPtje:
Besides, requesting a hit is an interaction between two players. Having to type a reason makes this hit more annoying and defeats the purpose of "*the hitman asks no questions*" roleplay. 

It would have been a decent idea if it actually prevented abuse, but it doesn't. Forcing people to type a reason might work when everyone can see it, but not when it's a private interaction.

XRAYBlake:
Thank you for the feedback. Since I didn't really know how to make it work, I referenced the police module regarding wanted and warrants.

PR #1926, opened 2015-05-11
Now *AvatarLarge* does not exist.

cardentity:


Colefrick:
Does this patch fix the over sized avatars in the scoreboard?

FPtje:
No

Colefrick:
Ok thanks.

Is the over sized avatars a server side error or a darkrp error?

FPtje:
It should be your error, that was fixed ages ago. If I'm not mistaken this
pull request removes dead code that resulted from that fix.

Colefrick:
Ok thanks.

PR #1927, opened 2015-05-13
Update sv_gamemode_functions.lua

cardentity:
+Performance

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -536,6 +536,10 @@ local function initPlayer(ply)
 	end
 end
 
+
+local DisconnectList = {}
+
+
 function GM:PlayerInitialSpawn(ply)
 	self.BaseClass:PlayerInitialSpawn(ply)
 	DarkRP.log(ply:Nick().." ("..ply:SteamID()..") has joined the game", Color(0, 130, 255))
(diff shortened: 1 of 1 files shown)


FPtje:
Hmm, I get the idea of the pr: DarkRP shouldn't try to regain ownership if the player is not "rejoining". My gut says there should be a more elegant way to solve, this, though.

cardentity:
Good idea. But now it being better than nothing.

FPtje:
I'm looking at it right now. Why should it loop over all entities in the
first place?

cardentity:
It set own "disconnected player" entities. I moved it into "if"

FPtje:
It still loops over all entities. Besides, one timer per entity is not clever performance wise. Like I said, I'm looking at it.

cardentity:
Yep.

FPtje:
Alright, sorry for not merging your thing, but thanks a lot for the pr anyway. Bringing this up was important. 

Here's what I've improved:
- On disconnected, it loops over all entities once instead of *once for every custom entity*
- Instead of a timer per entity (and player), there is now one timer per player. More timers = less performance. 
- on PlayerInitialSpawn it only loops over the entities that were queued to be removed. It also skips this if the player hasn't played recently, thanks to your pr. 

Thanks a lot for this pr!

cardentity:
No reason to apologize :)

PR #1929, opened 2015-05-14
Made DarkRPVars global to stop memory leak of valueable player inform…

ChampRay:
…ation when auto refreshing.


FPtje:
Makes no sense at all.

PR #1931, opened 2015-05-14
Fixed error that pops up when lua autorefreshing.

ChampRay:
`[ERROR] gamemodes/darkrp/gamemode/modules/base/sh_entityvars.lua:80: attempt to index global 'DarkRPVar' (a nil value)
  1. readNetDarkRPVar - gamemodes/darkrp/gamemode/modules/base/sh_entityvars.lua:80
   2. func - gamemodes/darkrp/gamemode/modules/base/cl_entityvars.lua:34
    3. unknown - lua/includes/modules/net.lua:32

### DIFF
--- a/gamemode/modules/base/sh_entityvars.lua
+++ b/gamemode/modules/base/sh_entityvars.lua
@@ -71,7 +71,7 @@ function DarkRP.readNetDarkRPVar()
 	local DarkRPVarId = net.ReadUInt(DARKRP_ID_BITS)
 	local DarkRPVar = DarkRPVarById[DarkRPVarId]
 
-	if DarkRPVarId == UNKNOWN_DARKRPVAR then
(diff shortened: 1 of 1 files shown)


FPtje:
This error happens when a third party mod registers a DarkRPVar on the server, but not the client. Making the error go away won't solve the problem, it will make it worse. Shit will go wrong and no one will know why. 

The problem of this error is that the client doesn't know *which* DarkRPVar isn't registered on the client. I should ping pong the code back to the server to have the server throw a proper error.

HDG-MrDeeJay:
Should be called "Hid error that indicated when lua autorefreshing."

FPtje:
Oh wait, my proposed solution would rely on the client to tell the server which DarkRPVars have been registered clientside. This would open up an exploit.

Sorry, can't be done.

PR #1932, opened 2015-05-15
Made chat commands table local

fruitwasp:


FPtje:
Breaks backwards compatibility with very little reason

fruitwasp:
Isn't is faster to call a local table instead of a global? I could add backwards compatibility.

```DarkRP.chatCommands = chatCommands```

FPtje:
You won't notice the difference

PR #1939, opened 2015-05-24
Update sv_jobs.lua

cardentity:


FPtje:
What are you fixing here?

cardentity:
earlier: if limit is one of the teams you cannot swap

cardentity:
e.g, earlier, you can't swap mayor and citizen, because mayor limit is 1

TheEMP:
But then it would allow you to bypass all other checks that the job may have. Your fixing one problem and creating a lot more.

FPtje:
Yup, this bypasses a ton of checks.

PR #1941, opened 2015-05-28
Possible fix for the rare case of sticks not resetting its view model colour; fixed colour affecting whole arm; added stunstick spark effect

Bo98:
Still couldn't reproduce it but this will make what should be impossible even more impossible.

Instead of calling SetColor on the view model, it now sets the colour of the render using render.SetColorModulation. This only applies to the current frame so it should be impossible for it to carry over to other weapons. The shiny material is also now reset on ViewModelDrawn rather than in Holster so that we are handling everything on a per frame basis.

### DIFF
--- a/entities/weapons/stick_base/shared.lua
+++ b/entities/weapons/stick_base/shared.lua
@@ -42,6 +42,10 @@ function SWEP:Initialize()
 
 	CreateMaterial("darkrp/"..self:GetClass(), "VertexLitGeneric", {
 		["$basetexture"] = "models/debug/debugwhite",
+		["$surfaceprop"] = "metal",
+		["$envmap"] = "env_cubemap",
+		["$envmaptint"] = "[ .5 .5 .5 ]",
+		["$selfillum"] = 0,
 		["$model"] = 1
 	}):SetVector("$color2", self.StickColor:ToVector())
 end
--- a/entities/weapons/stick_base/shared.lua
+++ b/entities/weapons/stick_base/shared.lua
@@ -74,10 +74,16 @@ function SWEP:Deploy()
 	return true
 end
 
-function SWEP:PreDrawViewModel()
-	if SERVER or not IsValid(self:GetOwner()) or not IsValid(self:GetOwner():GetViewModel()) then return end
-	self:GetOwner():GetViewModel():SetColor(self.StickColor)
-	self:GetOwner():GetViewModel():SetMaterial("models/shiny")
+function SWEP:PreDrawViewModel(vm)
+	if not IsValid(vm) then return end
+	local colVec = self.StickColor:ToVector()
+	render.SetColorModulation(colVec.x, colVec.y, colVec.z)
+	vm:SetMaterial("models/shiny")
+end
+
+function SWEP:ViewModelDrawn(vm)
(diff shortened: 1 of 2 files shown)


HDG-MrDeeJay:
Well if we are handling it per frame then it should now be 110% impossible.

But then again, it was also impossible to have a computer in your own home.

FPtje:
That looks really finicky. Any performence effect on resetting the materials all the time?

Bo98:
It was already setting the shiny material every frame for some reason so there was already some material setting going before. Now it just clears it too. I didn't notice any FPS drop in my testing.

What might be a better idea for the colour part is to roll that into a custom material using $color2. Will be a bit cleaner too.

Bo98:
I moved the colour setting to a material in the last commit. It's also less glossy/metallic than `models/shiny`. By moving colour setting to materials, we can now avoid colouring the arm by using SetSubMaterial, which I have done:

pic

FPtje:
But I like it shiny D:

Bo98:
Alright I'll see if I can get it back. `CreateMaterial` doesn't work well with some things for some reason.

```lua
	["$basetexture"] = "models/debug/debugwhite",
	["$surfaceprop"] = "metal",
```

```lua
] lua_run_cl print(Material("!darkrp/arrest_stick"):GetString("$surfaceprop"))

] lua_run_cl print(Material("!darkrp/arrest_stick"):GetString("$basetexture"))
models/debug/debugwhite
```

HDG-MrDeeJay:
You like cops walking around looking like they shoved an omni-coloured 
candycane into a large jar of lube?

FPtje:
Yes, that's exactly what I like.

HDG-MrDeeJay:
God damn it Falco, I find myself questioning your sexuality more and 
more by the day.

In all seriousness though, I prefer this candy-pain style.

Bo98:
Huh, apparently I needed to restart the game to get any small material change to work.

HDG-MrDeeJay:
Welcome to garry's mod. Where the most simplest of things require the 
most stupidest of tasks to be performed.

FPtje:
There was a command mat_reloadallmaterials or something

Bo98:
Would've been nice if CreateMaterial was smart enough to do that itself.

Bo98:
Have a stunstick that sparks electricity as a bonus:

https://www.youtube.com/watch?v=jey1KIfEWQ0

pic

Bo98:
Now the threaten function actually looks threatening.

HDG-MrDeeJay:
does the third person model sorta flick his baton like CP do in HL2?

that should happen

Bo98:
Right now it's more like a really violent whack.

HDG-MrDeeJay:
i suppose its threatening....

HDG-MrDeeJay:
but not classy....

Bo98:
Oh you mean the threaten, just holds up the baton in the air.

Bo98:
But now it sparks.

HDG-MrDeeJay:
At this stage the pull request title reminds me of a story book about an 
autistic kid.

It was called "The Curious Incident of the Dog in the Night-Time".

PR #1958, opened 2015-06-04
Fixed #1957

### LINKED ISSUE #1957: Error stuntick
[ERROR] gamemodes/darkrp/entities/weapons/stunstick/shared.lua:92: attempt to index local 'attachment' (a nil value)
  1. unknown - gamemodes/darkrp/entities/weapons/stunstick/shared.lua:92

TylerGeorges:
Credit to @00875 for helping fix it.

### DIFF
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -86,15 +86,17 @@ function SWEP:PostDrawViewModel(vm)
 end
 
 function SWEP:DrawWorldModel()
-	self:DrawModel()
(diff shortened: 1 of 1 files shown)


FPtje:
Why did you replace all tabs with spaces?

00875:
This is not a solution, but workaround

fruitwasp:
Why don't you ```return``` if ```achievements == nil```, instead of that tree?

00875:
If code was more longer, i will use return

FPtje:
#1963 
Sorry you got overruled, but @Bo98 has the better solution.

PR #1962, opened 2015-06-04
Fixes allowed = nil in custom shipments with table syntax

Partixel:
Previously, when using table syntax, having allowed as nil would not allowed everyone to use the shipment

### DIFF
--- a/gamemode/modules/base/sh_createitems.lua
+++ b/gamemode/modules/base/sh_createitems.lua
@@ -698,26 +698,27 @@ end
 function DarkRP.createShipment(name, model, entity, price, Amount_of_guns_in_one_shipment, Sold_seperately, price_seperately, noshipment, classes, shipmodel, CustomCheck)
 	local tableSyntaxUsed = type(model) == "table"
 
-	local AllowedClasses = classes or {}
-	if not classes then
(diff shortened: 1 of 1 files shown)


Partixel:
```lua
local price = tonumber(price)
 local shipmentmodel = shipmodel or "models/Items/item_item_crate.mdl"
```

These should probably also be changed as they only apply to non-table syntax, although because it works as it is I haven't changed it.

Partixel:
Also, why do this exist?
```lua
	if customShipment.separate ~= nil then
		customShipment.seperate = customShipment.separate
	end
```

Bo98:
Notice the spelling. It was sep**e**rate for many years until it was noticed that it was spelt wrong so sep**a**rate was added while keeping backwards compatibility with sep**e**rate.

Partixel:
Yep, just noticed it

Bo98:
The default shipments should probably be using sep**a**rate so the incorrect spelling isn't being encouraged.

Partixel:
http://wiki.darkrp.com/index.php/DarkRP:CustomShipmentFields should probably be updated to reflect the change to "separate" as well

Partixel:
I'll make a pr for them now.

PR #1963, opened 2015-06-04
Properly fixed #1957

### LINKED ISSUE #1957: Error stuntick

Bo98:
Instead of hiding the problem like #1958 did, this actually fixes the problem by relying on the Bip01_R_Hand bone which every (working) player model should have. The previously used anim_attachment_rh attachment seems to not be required in a player model and thus some don't have it as #1957 has revealed. Also added a nil check in case someone decides that it's a good idea to set the player model to something that isn't a player model.

### DIFF
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -4,6 +4,7 @@ if CLIENT then
 	SWEP.PrintName = "Stun Stick"
 	SWEP.Slot = 0
 	SWEP.SlotPos = 5
+	SWEP.RenderGroup = RENDERGROUP_BOTH
 
 	killicon.AddAlias("stunstick", "weapon_stunstick")
 end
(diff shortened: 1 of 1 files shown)


FPtje:
Nice work.

00875:
Hahaha! DrawModelTraslucent conflicts with many addons. For example: PAC. If set draw weapons to false, stunstick will still drawned. Also in some cases there are two playermodels. One is real, second is colored as stick and 'follows' player at minimal distance as trail.

Bo98:
That sounds more like an issue with PAC. How does it disable the drawing of weapons? The playermodel thing sounds really odd as the stunstick only touches the world model of the weapon.

Bo98:
Do you have a way to reproduce the playermodel thing?

Bo98:
Couldn't reproduce the draw weapons thing but I could reproduce the playermodels part:

```lua
player.GetByID(1).RenderOverride = function(ent) ent:DrawModel() end
```

Though it doesn't seem to be anything to do with `DrawWorldModelTranslucent`. Setting the render group to `RENDERGROUP_OPAQUE` still seems to make the colouring of the playermodel appear when attacking while moving as well as during babygod so the translucent drawing isn't the sole problem.

HDG-MrDeeJay:
I believe that Swep Constructor Kit has the same issue, could be wrong.

00875:
Now i not at home, so i can't give you any pics of this. When i will be at home, i will upload pics

00875:
Overriding weapons by PAC are successful, but there still problem with second player model.

00875:
So...

Bo98:
Still not convinced that is is a DarkRP bug. Yes DarkRP might trigger it, but it does sound like either:
* RenderOverride is broken for players and does not take materials into account.
* PAC is drawing it wrong.

You can reproduce it without PAC with:

```lua
lua_run_cl Entity(1).RenderOverride = function(ent) ent:DrawModel() end
```

The Source engine must do something extra with handling of materials when it draws players normally. I'll poke around the Source SDK in case PAC is missing something.

HDG-MrDeeJay:
I know what PAC is missing, but you wont find it on the SDK.

Hint: It's a developer that actually gives a shit CC @Hazbell

00875:
Confirmed to working. RenderOverride is rendering wrongly.
``` Entity(1).RenderOverride = function(ent) ent:DrawModel() end ```

PR #1965, opened 2015-06-05
Stunstick Fix

00875:
This can be merged

### DIFF
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -87,7 +87,6 @@ function SWEP:PostDrawViewModel(vm)
 end
 
 function SWEP:DrawWorldModelTranslucent()
-	self:DrawModel()
 	if CurTime() <= self:GetLastReload() + 0.1 then
 		local bone = self:GetOwner():LookupBone("ValveBiped.Bip01_R_Hand")
 		local bonePos, boneAng
@@ -105,6 +104,7 @@ function SWEP:DrawWorldModelTranslucent()
 			cam.End3D()
 		end
 	end
+	self:DrawModel()
 end
 
 local entMeta = FindMetaTable("Entity")
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -4,9 +4,9 @@ if CLIENT then
 	SWEP.PrintName = "Stun Stick"
 	SWEP.Slot = 0
 	SWEP.SlotPos = 5
-	SWEP.RenderGroup = RENDERGROUP_BOTH
 
 	killicon.AddAlias("stunstick", "weapon_stunstick")
+	SWEP.RenderGroup = RENDERGROUP_BOTH
 end
 
 DEFINE_BASECLASS("stick_base")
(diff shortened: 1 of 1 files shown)


FPtje:
Why did you change SWEP:DrawWorldModelTranslucent to  SWEP:DrawWorldModel

00875:
@FPtje What changed if call DrawModel() before and after?

FPtje:
It's your pull request, you tell me! Do you even know what you're doing?

00875:
I'll see nothing changed. So thats good!

Bo98:
I moved it after for a reason. It's to do with Z ordering.

Anyway why do we need the attachment bit? Did you find a model that does not have a right hand but has a right hand animation attachment?

Partixel:
I believe he said it was to prevent erroring if the player has the "ERROR" playermodel

Bo98:
The fix for that is a simple nil check for LookupBone, nothing else here is necessary as I am aware of.

Partixel:
Well, he's clearly not the smartest of the bunch, it would probably be best if someone helped him along because this gets drawn out over several days.

Partixel:
```lua
function SWEP:DrawWorldModelTranslucent()
	if CurTime() <= self:GetLastReload() + 0.1 then
		if !IsValid( self:GetOwner():LookupBone("ValveBiped.Bip01_R_Hand") ) then return end
		local bonePos, boneAng = self:GetOwner():GetBonePosition(self:GetOwner():LookupBone("ValveBiped.Bip01_R_Hand"))
		if bonePos then
			local pos = bonePos + (boneAng:Up() * -16) + (boneAng:Right() * 3) + (boneAng:Forward() * 6.5)
			render.SetMaterial(Material("sprites/light_glow02_add"))
			render.DrawSprite(pos, 32, 32, Color(255, 255, 255))
		end
	end
	self:DrawModel()
end
```
Would that do?

Bo98:
Bones don't have an IsValid method so just a simple nil check will be what's needed here.

While returning would be nicer here too, it prevents DrawModel being called which is why I avoided it for bonePos. Does lead to the [pyramid problem](https://github.com/FPtje/DarkRP/commit/81039b7b388ba4c53ce37767e57c076836ff54a1) though.

```lua
if a then
	if n then
		if d then
			if _ then
				if p then
					if y then
						if r then
							if a then
								if m then
									if i then
										if d then
											if s then
												if _ then
													if a then
														if r then
															if e then
																if _ then
																	if n then
																		if o then
																			if t then
																				if _ then
																					if n then
																						if i then
																							if c then
																								if e then
																									if _ then
																										if c then
																											if o then
																												if d then
																													if e then
																														-------|>-|o
																													end
																												end
																											end
																										end
																									end
																								end
																							end
																						end
																					end
																				end
																			end
																		end
																	end
																end
															end
														end
													end
												end
											end
										end
									end
								end
							end
						end
					end
				end
			end
		end
	end
end
```

Partixel:
Updated my original comment with a nil check + DrawModel(), will that do?

Bo98:
Sure, don't need the `var == nil` though, `nor var` is the normal way to do it. If you'd like me being picky you can make the LookupBone line store to a variable (eg. boneIdx) for less code duplication.

Partixel:
```nor var``` Righty

Partixel:
I'll make a pr with the change

Partixel:
#1966

PR #1966, opened 2015-06-05
Fixes stunstick error when bone is nil

Partixel:


Bo98:
How such a small fix took 3 pull requests is beyond me.

Partixel:
Does this need to be applied to any of the other sticks or not?

Bo98:
No, I only added the glow/spark effect to the stunstick.

Partixel:
Alrighty.

PR #1970, opened 2015-06-06
Fixed Player:getEyeSightHitEntity returning entities behind the player

Bo98:
And thus fixes this happening with the sticks.


Partixel:
What's "this"?

Bo98:
You previously could arrest/unarrest/stun people behind you.

Partixel:
Ah, I see. The way you said it sounded like you were referring to something I was missing.

PR #1996, opened 2015-06-21
Moneypickup

fruitwasp:
I think this is useful.


FPtje:
I know there already are `on...` hooks in DarkRP, but I'd like to keep that to a minimum. Could you change it to `playerPickedUpMoney`?

PR #2020, opened 2015-07-14
Added spawned entity class vars

fruitwasp:
Cleaner than Entity:GetClass() == "x"

### DIFF
--- a/entities/entities/spawned_money/init.lua
+++ b/entities/entities/spawned_money/init.lua
@@ -28,7 +28,7 @@ function ENT:Use(activator,caller)
 	self:Remove()
 end
 
-function ENT:Touch(ent)
+function ENT:StartTouch(ent)
 	-- the .USED var is also used in other mods for the same purpose
 	if ent:GetClass() ~= "spawned_money" or self.USED or ent.USED or self.hasMerged or ent.hasMerged then return end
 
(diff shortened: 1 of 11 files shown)


FPtje:
Good idea, but shouldn't those changes resonate in the rest of DarkRP's code?

fruitwasp:
Will do.

fruitwasp:
Commit squash failed. Sorry about that.

entities/entities/spawned_money/init.lua
```diff
@@ -30,7 +30,7 @@ end
 
 function ENT:Touch(ent)
 	-- the .USED var is also used in other mods for the same purpose
-	if ent:GetClass() ~= "spawned_money" or self.USED or ent.USED or self.hasMerged or ent.hasMerged then return end
+	if not ent.IsSpawnedMoney or self.USED or ent.USED or self.hasMerged or ent.hasMerged then return end
```
FPtje: We might want to enforce that only two spawned_money can merge. That makes most sense and doesn't have the risk of errors with other money kinds that use different variables.

fruitwasp: Now, we're talking about that function anyway. Change the function to [ENT:StartTouch](http://wiki.garrysmod.com/page/ENTITY/StartTouch), so it's not called continuously?

FPtje: sure

PR #2050, opened 2015-08-05
Potentially fixed random rank resets.

00875:
Thats worked on my server for ages, so i create a pull request


FPtje:
I won't put arbitrary retries in functions that store something in the database.

00875:
Basicly: Something changes rank quicker than FAdmin to user and back to admin, and that writes to DB faster than FAdmin reads.

FPtje:
This pr won't solve that. CAMI will.

Partixel:
CAMI will solve all of our problems

PR #2052, opened 2015-08-06
Wake up player if he is disconnected

00875:
I think this is needed


HDG-MrDeeJay:
I don't think it is. Any issues that could occur from a dc'd player being asleep could easily be resolved with validation of the player entity.

00875:
No. If player disconnected, he's ragdoll does not deleted

Bo98:
It's worth noting here that the PlayerDiaconnected hook doesn't catch everything last time I checked. If you must use something like it then try the player_disconnect gameevent.

Unless it was fixed but I don't think it was.

FPtje:
Ragdolls are being removed on disconnect just fine for me.

PR #2054, opened 2015-08-08
Added lockpickStarted hook

fruitwasp:
Also fixed lockpickTime hook ply call


entities/weapons/lockpick/shared.lua
```diff
@@ -215,6 +222,30 @@ DarkRP.hookStub{
...
+			description = "The player that is about to pick a lock.",
+			type = "Player"
+		},
+		{
+			name = "ent",
+			description = "The entity being lockpicked.",
+			type = "Entity"
+		},
+		{
+			name = "trace",
```
TheEMP: Unless I'm missing something, the hook doesn't return the trace...

PR #2066, opened 2015-08-13
Logging changes

fruitwasp:


Bo98:
Problem about the writetofile parameter is that old calls (like existing ones in DarkRP) will suddenly not write to file. Might have to invert that.

gamemode/modules/logging/cl_init.lua
```diff
@@ -3,7 +3,32 @@ Log a message to console
...
+	
+	hook.Call("darkRPLogPrinted", nil, text, colour)
 end
 usermessage.Hook("DRPLogMsg", AdminLog)
+
+/*---------------------------------------------------------------------------
+Interface
+---------------------------------------------------------------------------*/
+DarkRP.hookStub{
+	name = "darkRPLogPrinted",
```
Bo98: Should the D in `darkRP` be capitalised here? Existing hooks beginning with DarkRP have a capital D.

http://wiki.darkrp.com/index.php/Category:Hooks

LetsRaceBwoi: Yeah it should really have a big D

PR #2078, opened 2015-08-18
Added ability to give a reason when a player cannot change their job

fruitwasp:


HDG-MrDeeJay:
What happened to customCheck and CustomCheckFailMsg?

fruitwasp:
> What happened to customCheck and CustomCheckFailMsg?

That's a check to use per job.

PR #2079, opened 2015-08-18
Change timer.Destroy to timer.Remove

Fuzzik:
Did a quick find an replace, let me know if I missed something


NamefuI:
Isn't timer.Destroy() and timer.Remove() basically the same thing? Or is there any difference?

NamefuI:
I've just looked it up on the wiki and I can see why you want it changed. I agree with you.

PR #2093, opened 2015-08-24
After Cleanup setup doors

00875:
After that fix:


fruitwasp:
I think it belongs in the doorsystem module.

Bo98:
It's calling local functions so it has to be there.

FPtje:
What happens if you remove the timer?

NamefuI:
"was puted on your bank account" ayy lmao

00875:
@FPtje for some reasons it sometimes don't work without timer.

PR #2108, opened 2015-08-28
Cleaned up deprecated crap

Bo98:
* ValidPanel -> IsValid
* SetKeyBoardInputEnabled -> SetKeyboardInputEnabled
* Cleaned up ConVar changes in other PR - the point of deprecating GetConVar\<Type\>() was to encourage people to cache the ConVar object (either through the return value of the creation function or through a one-time GetConVar call) because fetching it is slow. This is what now happens in most places that would need the ConVar potentially more than once.
* Fixed regression in the other PR which broke the FAdmin scoreboard
  - https://github.com/FPtje/DarkRP/commit/69e26130351ae20b43e1c14b763c000bda6ced77#commitcomment-12956649

### DIFF
--- a/gamemode/modules/fadmin/fadmin/logging/sv_init.lua
+++ b/gamemode/modules/fadmin/fadmin/logging/sv_init.lua
@@ -1,3 +1,4 @@
+local logging
 FAdmin.StartHooks["Logging"] = function()
 	FAdmin.Access.AddPrivilege("Logging", 3)
 	FAdmin.Commands.AddCommand("Logging", function(ply, cmd, args)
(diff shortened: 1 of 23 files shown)


00875:
This is not only clean up, this is also optimizatoions. Nice work!

gamemode/modules/fadmin/fadmin/logging/sv_init.lua
```diff
@@ -11,12 +12,13 @@ FAdmin.StartHooks["Logging"] = function()
...
 		return true, OnOff
 	end)
+	logging = GetConVar("FAdmin_logging")
 end
 
 local LogFile
 function FAdmin.Log(text, preventServerLog)
 	if not text or text == "" then return end
-	if not GetConVar("FAdmin_logging"):GetBool() then return end
+	if not logging:GetBool() then return end
```
FPtje: Is it possible that any hook gets called before logging is set? Might want to be sure of that.

FPtje: Added an extra check

FPtje:
Nice work.

gamemode/modules/fadmin/fadmin/logging/sv_init.lua
Bo98: I've not seen it happen but you could add a sanity check if you want to cover any weird and wonderful cases people seem to encounter:

```lua
if not logging then logging = GetConVar("FAdmin_logging") end
```

Bo98: Sounds good

PR #2112, opened 2015-08-30
Refactored labs (drug_lab/gunlab/microwave) to a common lab_base

Bo98:
A LOT less code duplication now. One or two of the labs have gained features and fixes that the other(s) had.


00875:
NICE. I already done that, but i thought what nobody needs this

entities/entities/gunlab/shared.lua
```diff
@@ -1,11 +1,9 @@
...
+ENT.Base = "lab_base"
 ENT.PrintName = "Gun Lab"
-ENT.Author = "Pcwizdan"
-ENT.Spawnable = false
-ENT.CanSetPrice = true
 
-function ENT:SetupDataTables()
-    self:NetworkVar("Int", 0, "price")
-    self:NetworkVar("Entity", 1, "owning_ent")
+function ENT:initVars()
```
00875: Maybe set vars directly? Like ENT.model in shipment, etc.

Bo98: I did it this way because I think DarkRPMod loads language files after entities are loaded which would cause issues with the DarkRP.getPhrase call. Same with the config.

Bo98:
Having a base is helpful for addons that want to make their own lab.

00875:
Yeah

PR #2114, opened 2015-08-30
Improved/fixed player class support

Bo98:
* All player class hooks are now called properly.
  - Perhaps fixes some addons that used player classes.
* Deprecated UpdatePlayerSpeed hook in favour of GMod's SetupMove/Move hooks.
  - Still called on spawn and job change for backwards compatibility.
    * No longer called in the weapons and police module as they only existed because DarkRP needed them itself.
  - Should a deprecated field be added to stubs? I changed the description but there already is a deprecated field in the wiki templates that the wiki generator could use to make it much more noticeable that it is deprecated.
* Player classes variables now respect the config if set to -1.
* Added playerClass job field.
* Added ply:applyPlayerClassVars(applyHealth).
* Added playerClassVarsApplied hook.
* DarkRP's player class now bases from Sandbox rather than the base gamemode.
  - I removed the bhopping thing though and crap we don't want like HL2 weapon loadout.


FPtje:
Deprecated field could be somewhat useful, I guess.

PR #2140, opened 2015-09-05
Make the lockdown more customizable

NamefuI:


FPtje:
Because global variables like that are very bad practice.
Also, I don't really like that this waitLock code is all back. Other functions handle that in a much cleaner way.

NamefuI:
What do you want me to do?

gamemode/config/config.lua
```diff
@@ -107,6 +107,12 @@ GM.Config.lockpickfading                = true
 GM.Config.logging                       = true
 -- lottery - Enable/disable creating lotteries for mayors
 GM.Config.lottery                       = true
+-- lockdown - Enable/Disable initiating lockdowns for mayors.
+GM.Config.lockdown                      = true
+-- lockdownMaxDuration - Whether you want the lockdown to end automatically after a certain amount of time (the amount of time can be configured below).
+GM.Config.lockdownMaxDuration           = true
+-- preventLockdownSpam - Whether the Mayor needs to wait a certain amount of time before initiating another lockdown (the amount of time can be configured below).
+GM.Config.preventLockdownSpam           = true
```
Bo98: These two configs probably would just complicate things. How about just using the timer configs and taking 0 to be disabled?

NamefuI:
I wouldn't say they would complicate things. There are a few different configs which use this practice.

gamemode/config/config.lua
NamefuI: I wouldn't say they would complicate things. It's pretty clear unless someone's half asleep or just retarded. It's up to Falco though.

FPtje:
Partially merged. Thanks for your input. It's not bad, but I decided to do it a bit differently.

NamefuI:
Now I see what you mean. Glad to be of service.

PR #2160, opened 2015-09-15
Fixed FAdmin Chat Auto-complete and Moved FAdmin Messaging to net

TheEMP:


Bo98:
You're using tabs in some places here. The DarkRP codebase uses spaces for indentation now.

FPtje:
Are we sure that third party addons don't hook into the usermessages?

FPtje:
The chat tab fix has been cherry-picked and fixed in style. I'm not just going to allow umsgs being changed to net messages, because there's a reasonable chance that it breaks backwards compatibility with third party addons.

TheEMP:
Thats fine but then we still need to address the other problem that changing to net message fixed, https://github.com/TheEMP/DarkRP/commit/5bbb91fc572be6470fd20b54784979246cb148ae FAdmin.Messages.ConsoleNotify(ply, message) sends a usermesssage to all players regardless of if they have the SeeAdmins privilege.

FPtje:
done

PR #2169, opened 2015-09-22
Added hook onPlayerSay

TheEMP:
As it stands right now there is no way to to alter messages the player
sends in chat. Since DarkRP rips out all the other PlayerSay hooks, this hook is
needed as an alternative.

### DIFF
--- a/gamemode/modules/chat/sv_chat.lua
+++ b/gamemode/modules/chat/sv_chat.lua
@@ -89,6 +89,10 @@ local function RP_PlayerChat(ply, text, teamonly)
         callback = callback or "" .. " "
     end
 
+    local command = tblCmd and tblCmd.command or ""
+    local overrideTxt = hook.Run("onPlayerSay", ply, command, text, teamonly)
+    if overrideTxt then return overrideTxt, callback, DoSayFunc end
+
     return text, callback, DoSayFunc;
 end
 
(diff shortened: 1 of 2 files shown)


Velkon:
I would shitpost, but please add this

FPtje:
All the old chat hooks are called at the start of playerSay, what are you talking about?

00875:
If override the gmod default one hook, chat will be global, etc.
Without overriding hook:
2015-09-23_16-19_ponyrp r 1 0 10 6 t
With override:
2015-09-23_16-20_garry s mod
2015-09-23_16-20_ponyrp r 1 0 10 6 t
2015-09-23_16-20_garry s mod 2

FPtje:
That doesn't happen to hooks that are added before `InitPostEntity`. Those hooks go through the player chat process. Any string that they return won't be global.

It's probably better to streamline this behaviour and make it that way for hooks added *after* `InitPostEntity` as well.

TheEMP:
Then that's what the problem is, I was adding the PlayerSay hook in the InitPostEntity hook so I could detect if the game mode was DarkRP. Looking more in depth now there are probably better hooks I could use, so I'll use those instead.

Velkon:
I WANT A CLIENTSIDE WAY TO EDIT THE CHAT MESSAGES WITHOUT FUCKING UP THE (OOC) TAGS

HDG-MrDeeJay:
CAPS LOCK IS NOT CRUISE CONTROL FOR COOL

Velkon:
But I'm still cooler than you 
Cooler

HDG-MrDeeJay:
you desperately wish.

PR #2198, opened 2015-10-19
removed useless globals moo and tmsql

Noiwex:


FPtje:
mySQLite has its own repository

Noiwex:
D:<

PR #2214, opened 2015-11-07
Allow players to drop $1

immibis:
It's pointless to prevent players from dropping $1 but allow them to drop $2.

Note: Since this is a trivial change, I've made it with GitHub's online editor and haven't tested it in-game.

### DIFF
--- a/gamemode/modules/money/sv_money.lua
+++ b/gamemode/modules/money/sv_money.lua
@@ -126,8 +126,8 @@ local function DropMoney(ply, args)
     end
     local amount = math.floor(tonumber(args))
 
-    if amount <= 1 then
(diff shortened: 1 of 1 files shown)


HDG-MrDeeJay:
given that all you are changing is a number, it hardly needs testing.

deinemudda32:
Its pointless a player would drop 1$.

immibis:
@deinemudda32 Then it's pointless a player would drop $2.

FPtje:
It's not up to decide that dropping $1 is pointless in every and all
roleplay situations.

HDG-MrDeeJay:
The number of times I've died due to some kid saying "drop $1 or die"... 
well now I can drop $1.

PR #2217, opened 2015-11-09
Hitmenu gethits

fruitwasp:


fruitwasp:
Eh, whoops.

FPtje:
Nice use of fp though.

PR #2232, opened 2015-12-04
fn module optimization (Composition function)

thelastpenguin:
speed improvement by a factor of 1715.2 according to my benchmarks.

I've optimized the composition function by removing the need for temporary tables when passing values between functions. My new implementation keeps all of the values in the stack.

I've optimized the composition function by removing the need for temporary tables as seen in the old implementation
```lua
local res = {...}
for i = #funcs, 1, -1 do
	res = {funcs[i](unpack(res))}
end
return unpack(res)
```

The new implementation keeps all of the work on the stack by doing function compositions in what I feel is a slightly more intuitive way and one that makes Lua a lot happier.

See my benchmarks http://hastebin.com/jizuvizuba.lua to confirm speed improvement.
I've tested this and it doesn't appear to conflict with the current implementation in any ways

### DIFF
--- a/gamemode/libraries/fn.lua
+++ b/gamemode/libraries/fn.lua
@@ -75,7 +75,7 @@ end
 Misc functions
 ---------------------------------------------------------------------------*/
 -- function composition
-Compose = function(funcs, ...)
+do
 	local function comp_h(a, b, ...)
 		if b == nil then return a end
 		b = comp_h(b, ...)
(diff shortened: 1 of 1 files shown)


aStonedPenguin:
thisIsGood

00875:
I think comp_h needs to be outside of function

aStonedPenguin:
no it doesn't

thelastpenguin:
@00875 I put it in there since it's a bit cleaner to bundle them. I'd hope that Lua wouldn't create a new instance of the function unless it uses upvalues but I'm not really sure. I can move it out though if there is concensus that it's better.

00875:
Every time when function Compose called, it creates new local function in memory. So, i would put this outside of Compose function

thelastpenguin:
@00875 would it be better written like this?

```lua
do
	local function comp_h(a, b, ...)
		if b == nil then return a end
		b = comp_h(b, ...)
		return function(...)
			return a(b(...))
		end
	end
	Compose = function(funcs, ...)
		if type(funcs) == 'table' then
			return comp_h(unpack(funcs))
		else
			return comp_h(funcs, ...)
		end
	end
end
```

00875:
Of course

FPtje:
Nice work. Could you make the PR to https://github.com/FPtje/GModFunctional, though? Thanks!

PR #2235, opened 2015-12-05
Added optional message if entity cannot be lockpicked

fruitwasp:


FPtje:
I'd rather have the hook writer call `DarkRP.notify` before returning false.

PR #2239, opened 2015-12-07
Entity prices

fruitwasp:


FPtje:
canBuyX returns price, having yet another hook for price is nonsense.

PR #2280, opened 2016-01-06
Fix FireNotification breaking FAdmin_OnCommandExecuted

TheEMP:
So I noticed that the new Notification system at some point changes the targets variable in commands from this:
1	=	Player [1][|FP| TheEMP]
to this:
1	=	Brother Gustro(STEAM_0:1:55485548)
This interferes with with the res arguments passed to FAdmin_OnCommandExecuted as it now doesn't pass the player objects.


FPtje:
I don't see where this supposed change is happening. targets is a table of player objects. It remains that way throughout that function, they are sent as-is to the client.

FPtje:
FAdmin.TargetsToString does it, and it's called indirectly from the lognotification function. That's where it should be fixed.

FPtje:
FAdmin.TargetsToString that is, not the log

PR #2281, opened 2016-01-07
Fix TargetsToString breaking FAdmin_OnCommandExecuted

TheEMP:
#2280


FPtje:
Why are you creating a local variable with the same name as the parameter?

FPtje:
Ffs, trailing whitespace. 

Install glualint, seriously.

PR #2292, opened 2016-01-14
custom spawn positions 

kulcris:
Custom spawn positions not based on job or rank


FPtje:
There's already a hook for that

PR #2304, opened 2016-01-23
Add "weaponIllegal" hook

AMD-NICK:
Some scripts looks like this:
```lua
function SomeFunction(ply,wep)
	ply:Give(wep):SetNWBool("from_dispenser",true)
end

hook.Add("canDropWeapon","name",function(ply, wep)
	if wep:GetNWBool("from_dispenser") then return false end
end)
```
But after strip weapons by weaponchecher weapon data just reset and code above does not work.
I propose to add the following hook to fix the problem

```lua
hook.Add("weaponIllegal","name",function(wep)
	if wep:GetNWBool("from_dispenser") then return true end
end)
```

### DIFF
--- a/entities/weapons/weaponchecker/shared.lua
+++ b/entities/weapons/weaponchecker/shared.lua
@@ -127,19 +127,20 @@ end
 
 function SWEP:GetStrippableWeapons(ent, callback)
     CAMI.PlayerHasAccess(ent, "DarkRP_GetAdminWeapons", function(access)
-        for k,v in pairs(ent:GetWeapons()) do
-            if not v:IsValid() then continue end
-            local class = v:GetClass()
+        for _,w in pairs(ent:GetWeapons()) do
+            if not w:IsValid() then continue end
+            local class = w:GetClass()
+            local illegal = hook.Call("weaponIllegal", nil, w)
 
             if GAMEMODE.Config.weaponCheckerHideDefault and (table.HasValue(GAMEMODE.Config.DefaultWeapons, class) or
                 access and table.HasValue(GAMEMODE.Config.AdminWeapons, class) or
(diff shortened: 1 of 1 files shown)


entities/weapons/weaponchecker/shared.lua
```diff
@@ -127,19 +127,20 @@ end
...
-            local class = v:GetClass()
+        for _,w in pairs(ent:GetWeapons()) do
+            if not w:IsValid() then continue end
+            local class = w:GetClass()
+            local illegal = hook.Call("weaponIllegal", nil, w)
 
             if GAMEMODE.Config.weaponCheckerHideDefault and (table.HasValue(GAMEMODE.Config.DefaultWeapons, class) or
                 access and table.HasValue(GAMEMODE.Config.AdminWeapons, class) or
-                ent:getJobTable() and ent:getJobTable().weapons and table.HasValue(ent:getJobTable().weapons, class)) then
+                ent:getJobTable() and ent:getJobTable().weapons and table.HasValue(ent:getJobTable().weapons, class) or illegal) then
```
Bo98: This should probably be in its own `if` check as `GAMEMODE.Config.weaponCheckerHideDefault` should have no affect on the hook.

Also, would it not make more sense if a return value of `true` meant that it _can_ be stripped since the hook is called `weaponIllegal`?

Bo98:
IMO the hook would be a nice one to have but I don't think it's the proper solution for _this_ problem. Right now you're just preventing any stripping of weapons with that NWBool because DarkRP breaks it and doesn't fix it. What if someone wanted to have it strippable but still keep the NWBool intact? A better solution here would be for DarkRP to save the NWVars table (`wep:GetNWVarTable()`) and set it back when the weapons are returned but I think that function only works with the experimental NWVars2.

AMD-NICK:
Ok. Thanks for answer. I will alter my method tomorrow

PR #2312, opened 2016-01-30
PlayerSayEx hook

MrBrax:
advanced linguistics


MrBrax:
FUCKING github why do you have to be so god damn complicated, i don't even know why that other commit is there

FPtje:
What does this solve what onChatCommand doesn't?

MrBrax:
more specific hooks for every type of say command, useful for logging to i suppose
i'm gonna use it on my end anyway

BillyOnWiiUIn144p:
```
if (DarkRP) then

	local function chathook(ply,txt,team)
		if (!IsValid(ply)) then return end
		if (!ply:IsPlayer()) then return end
		if (txt == "") then return end
		local teams = ""
		if (team == true) then
			teams = "(TEAM) "
		end
		bLogs.Log({
			module = "Chat",
			log = bLogs.GetName(ply) .. ": " .. teams .. txt,
			involved = {ply},
		})
	end

	bLogs.AddHook("onChatCommand","Chat",function(ply,cmd,args)
		if (!IsValid(ply)) then return end
		if (!ply:IsPlayer()) then return end
		if (type(cmd) == "table") then cmd = table.concat(cmd) end
		if (type(args) == "table") then args = table.concat(args," ") end
		if (string.lower(cmd) == "advert") then
			bLogs.Log({
				module = "Adverts",
				log = bLogs.GetName(ply) .. ": " .. args,
				involved = {ply},
			})
		else
			bLogs.Log({
				module = "Chat",
				log = bLogs.GetName(ply) .. ": /" .. cmd .. " " .. args,
				involved = {ply},
			})
		end
	end)
	bLogs.AddHook("PostPlayerSay","Chat",chathook)

else

	bLogs.AddHook("PlayerSay","Chat",chathook)

end
```

Let me save you some time?

-1

FPtje:
@BillyOnWiiUIn144p shows that the hook is redundant. I get that it may be easier to do precisely what you want, but I don't want to add hooks for things you can already do with other hooks. It would cause bloat confusion and general code terribleness.

MrBrax:
it won't correctly show targets for every action and individual parameters however

PR #2338, opened 2016-02-23
Add new option to hitman module

AMD-NICK:
My knowledge of English isn't good enough to describe the need for this change


FPtje:
I'd rather you'd have that config option prevent the setting of the DarkRPVar. Now clients still have the information.

FPtje:
I looked into it a bit. The problem is that clients need the knowledge about whether a hitman has a hit for other reasons. One of them is when you press E on a hitman. When the hitman has a hit, you get a message that you cannot request a hit. 

So, players can still see whether the hitman has a hit, even when the client isn't specifically sent that variable. The option has no use and would just disable the drawing of that one text. People would use it to "hide" hit information, which it won't do.

The PR doesn't do what one would expect it to do.

PR #2347, opened 2016-02-26
playerPickedUpCheque, playerToreUpCheque, playerDroppedCheque, playerDroppedMoney

BillyOnWiiUIn144p:
- playerPickedUpCheque
- playerToreUpCheque
- playerDroppedCheque
- playerDroppedMoney
- Added money entity to playerPickedUpMoney
- Moved some hookstubs to where they're called
- Added hookstubs for new hooks

For logging.

### DIFF
--- a/entities/entities/darkrp_cheque/init.lua
+++ b/entities/entities/darkrp_cheque/init.lua
@@ -28,12 +28,15 @@ function ENT:Use(activator, caller)
         owner = (IsValid(owner) and owner:Nick()) or DarkRP.getPhrase("disconnected_player")
         DarkRP.notify(activator, 0, 4, DarkRP.getPhrase("found_cheque", DarkRP.formatMoney(amount), "", owner))
         activator:addMoney(amount)
+        hook.Call("playerPickedUpCheque", nil, activator, recipient, amount or 0, true, self)
         self:Remove()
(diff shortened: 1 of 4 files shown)


HDG-MrDeeJay:
Would it not be a million times better to just have a hook whenever the 
wallet changes, and specify what the source or reason for this change is.

This way you would not only log Cheques and Cash, you would also log any 
addon, plugin or script that made any adjustment to the players wallet, 
which would be far more effective than just logging cheques and money. 
The only time I'd see that as unnecessary is when someone uses darkrp 
add/setmoney <player> <amount>... as that is logged anyway.

BillyOnWiiUIn144p:
Yes, you could do that but you'd have to keep the old hooks to keep older addons working.

I would definitely like that way but legacy seems best from here.

HDG-MrDeeJay:
Perhaps in 2.7

FPtje:
You would have to make a reason system. How would you represent a reason? Some string? An enumeration? Just a number? What about addons that change wallet count, should they keep telling DarkRP why they changed wallet counts? If you use enums or numbers, you'd have to make some function to register enums or get some unique number. 

This gets really ugly really fast.

BillyOnWiiUIn144p:
Agreed, especially for older addons.

PR #2363, opened 2016-03-10
Fix potential error

AMD-NICK:
(not tested)

### DIFF
--- a/gamemode/modules/police/sv_commands.lua
+++ b/gamemode/modules/police/sv_commands.lua
@@ -140,7 +140,7 @@ function DarkRP.lockdown(ply)
     end
 
     if not GAMEMODE.Config.lockdown then
-        show(ply, 1, 4, DarkRP.getPhrase("disabled", "lockdown", ""))
+        show(DarkRP.getPhrase("disabled", "lockdown", ""))
         return ""
     end
 


fruitwasp:
That ```show``` function doesn't even exist.

AMD-NICK:
image

gamemode/modules/police/sv_commands.lua
```diff
@@ -157,7 +157,7 @@ function DarkRP.lockdown(ply)
     SetGlobalBool("DarkRP_LockDown", true)
     DarkRP.notifyAll(0, 3, DarkRP.getPhrase("lockdown_started"))
 
-    return ""
+    return true
```
AMD-NICK: if DarkRP.lockdown(ply) then

FPtje:
I don't see any ability here, and the only thing I see you doing here is returning no value or a boolean in a chat command, which should always see a string being returned.

Bo98:
The first commit was useful. The second reverted it for some reason.

FPtje:
I don't see any commit reverting anything. I see one commit changing the return values and another changing a show call.

Bo98:
Probably hadn't pulled the other commit before committing.

Bo98:
Apparently it shows up now. No idea why it didn't for me before.

AMD-NICK:
> in a chat command

As I understand, the DarkRP.lockdown function is global
I want to make module which override the default /lockdown command and allows to set lockdown reason in arguments, but without return value I should write shit-code or edit core files.

There is about I wanna do:
```lua
DarkRP.defineChatCommand("lockdown", function(ply,args)

	if args == "" then
		DarkRP.notify(ply, 1, 4, "*russian text. Don't want to transtate*")
		return
	end

	if DarkRP.lockdown(ply) then
		ULib.tsayColor( nil, false, Color(230,200,30), "Reason: " .. tostring(args) )
	end

end)
```

AMD-NICK:
...

AMD-NICK:
@FPtje, please, merge request (or hint me how to do my module it in other way). DarkRP.lockdown function is GLOBAL. Not only a chat command
I do not want to edit core files

FPtje:
The chat command should always return a string. The bool return is made redundant by the global bool that's set in the function. 

It would be nicer if  `DarkRP.lockdown` didn't return strings, though. I guess it could still return bools.

PR #2367, opened 2016-03-16
Minor fixes for non-DarkRP Skins

Hazbelll:
- Hidden the minimise/maximise buttons on vote/question windows.
- Removed the hardcoded text and cursor colour used in the search box of the F1 help menu when not using the official DarkRP skin.
- Corrected a typo in a comment.

### DIFF
--- a/gamemode/modules/f1menu/cl_searchbox.lua
+++ b/gamemode/modules/f1menu/cl_searchbox.lua
@@ -6,15 +6,18 @@ function PANEL:Init()
     self.BaseClass.Init(self)
     self.F1Down = true
     self:SetFont("DarkRPHUD2")
-    self:SetTextColor(Color(255,255,255,255))
-    self:SetCursorColor(Color(255,255,255,255))
 
     self.lblSearch = vgui.Create("DLabel", self)
     self.lblSearch:SetFont("DarkRPHUD2")
-    self.lblSearch:SetColor(Color(200, 200, 200, 200))
     self.lblSearch:SetText(DarkRP.getPhrase("f1Search"))
     self.lblSearch:SizeToContents()
     self.lblSearch:SetPos(5)
+
+    if GAMEMODE.Config.DarkRPSkin == "DarkRP" then
+        self:SetTextColor(Color(255,255,255,255))
+        self:SetCursorColor(Color(255,255,255,255))
+        self.lblSearch:SetColor(Color(200, 200, 200, 200))
+    end
 end
 
 function PANEL:OnLoseFocus()
--- a/gamemode/modules/f1menu/cl_searchbox.lua
+++ b/gamemode/modules/f1menu/cl_searchbox.lua
@@ -6,18 +6,15 @@ function PANEL:Init()
     self.BaseClass.Init(self)
     self.F1Down = true
     self:SetFont("DarkRPHUD2")
+    self:SetTextColor(Color(255,255,255,255))
+    self:SetCursorColor(Color(255,255,255,255))
 
     self.lblSearch = vgui.Create("DLabel", self)
     self.lblSearch:SetFont("DarkRPHUD2")
+    self.lblSearch:SetColor(Color(200, 200, 200, 200))
     self.lblSearch:SetText(DarkRP.getPhrase("f1Search"))
     self.lblSearch:SizeToContents()
     self.lblSearch:SetPos(5)
-
-    if GAMEMODE.Config.DarkRPSkin == "DarkRP" then
-        self:SetTextColor(Color(255,255,255,255))
-        self:SetCursorColor(Color(255,255,255,255))
-        self.lblSearch:SetColor(Color(200, 200, 200, 200))
-    end
 end
 
 function PANEL:OnLoseFocus()
--- a/entities/weapons/keys/cl_menu.lua
+++ b/entities/weapons/keys/cl_menu.lua
@@ -50,7 +50,9 @@ local function openMenu(setDoorOwnerAccess, doorSettingsAccess)
 
     KeyFrameVisible = true
     local Frame = vgui.Create("DFrame")
-    Frame:SetSize(200, 30) -- base size
+    Frame:SetSize(200, 30) -- Base size
+    Frame.btnMaxim:SetVisible(false)
+    Frame.btnMinim:SetVisible(false)
     Frame:SetVisible(true)
     Frame:MakePopup()
 
--- a/entities/weapons/pocket/cl_menu.lua
+++ b/entities/weapons/pocket/cl_menu.lua
@@ -34,6 +34,8 @@ function DarkRP.openPocketMenu()
     frame = vgui.Create("DFrame")
 
     frame:SetTitle(DarkRP.getPhrase("drop_item"))
+    frame.btnMaxim:SetVisible(false)
+    frame.btnMinim:SetVisible(false)
     frame:SetVisible(true)
     frame:MakePopup()
 
--- a/gamemode/modules/animations/sh_animations.lua
+++ b/gamemode/modules/animations/sh_animations.lua
@@ -180,6 +180,8 @@ local function AnimationMenu()
     AnimFrame:SetSize(130, Height)
     AnimFrame:SetPos(ScrW() / 2 + ScrW() * 0.1, ScrH() / 2 - (Height / 2))
     AnimFrame:SetTitle(DarkRP.getPhrase("custom_animation"))
+    AnimFrame.btnMaxim:SetVisible(false)
+    AnimFrame.btnMinim:SetVisible(false)
     AnimFrame:SetVisible(true)
     AnimFrame:MakePopup()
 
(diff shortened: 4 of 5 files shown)


gamemode/modules/f1menu/cl_searchbox.lua
```diff
@@ -6,15 +6,18 @@ function PANEL:Init()
...
-    self.lblSearch:SetColor(Color(200, 200, 200, 200))
     self.lblSearch:SetText(DarkRP.getPhrase("f1Search"))
     self.lblSearch:SizeToContents()
     self.lblSearch:SetPos(5)
+
+    if GAMEMODE.Config.DarkRPSkin == "DarkRP" then
+        self:SetTextColor(Color(255,255,255,255))
+        self:SetCursorColor(Color(255,255,255,255))
+        self.lblSearch:SetColor(Color(200, 200, 200, 200))
+    end
```
Bo98: There is a much better was to do this. I have it in my copy but haven't had the chance to send a pull request for it.

You could remove the first two colour setting lines and replace them with

```lua
SKIN.colTextEntryText       = Color(255, 255, 255)
SKIN.colTextEntryTextCursor = Color(255, 255, 255)
```

in the skin.

As for the `lblSearch`, because it only applies to _this_ label then it's a bit more awkward. There's a GMod pull request open https://github.com/garrynewman/garrysmod/pull/972 that standardises placeholders (I've requested changes for proper SKIN support) but don't expect it to be merged in the next couple years. For now, I'd probably just create a custom SKIN field for this and call `SetColor` if it exists though it will unfortunately be something we'd need to provide backwards compatibility for if placeholders did arrive properly in GMod.

Hazbelll: I did originally have a look in `cl_dermaskin.lua`, which I found hadn't defined the colours for anything like had the default derma skin, so figured this was probably by choice and wanted to be left alone (I've noticed myself that changing the properties of things in a SKIN table away from what the default derma skin has set can cause elements not using that skin to also be affected - can make things look very odd).

The F1 menu's search box was the only thing I found when using both the default derma skin and my own custom-made skin where colour conflictions occurred with very little you could do to otherwise prevent it (changing the colour of your search boxes for *one* element is a tad extreme), and in the case of someone using the default derma theme, nothing at all.

But that's why I opted for a more local solution than a global one, because as far as I can tell, it's only this which is troublesome; but I do agree your approach to it would be better.

Bo98: > (I've noticed myself that changing the properties of things in a SKIN table away from what the default derma skin has set can cause elements not using that skin to also be affected - can make things look very odd)

That should never happen. If it did then existing DarkRP SKIN fields would also have the issue and I haven't seen a complaint about it. If you found a way to reproduce this bug however, then it would definitely be worth reporting it.

> (changing the colour of your search boxes for one element is a tad extreme)

Addons, future DarkRP updates etc. could change this. The current transparent background for text entries currently is applied to _all_ DTextEntry elements with the DarkRP skin which is why a global colour change makes sense.

The only case I see it being a problem, because of the transparent background of the element, is if someone used a light background in the parent but then they are no longer using the DarkRP colour scheme (ie. they are using something like a hardcoded `SetBackgroundColor`/`SetBGColor`) so should be making their own skin (or even continue their hardcoded colour-setting decision by using `SetTextColor`). There would likely be conflicts anyway with a blue or red background like some other DarkRP elements used.

Hazbelll: > That should never happen. If it did then existing DarkRP SKIN fields would also have the issue and I haven't seen a complaint about it. If you found a way to reproduce this bug however, then it would definitely be worth reporting it.

I know it shouldn't, but it seems to in some cases. I believe it's to do with [ForceDermaSkin](http://wiki.garrysmod.com/page/GM/ForceDermaSkin) being used, but the affected panels using [SetSkin](http://wiki.garrysmod.com/page/Panel/SetSkin), to something else. I can't confirm though, not looked into it properly.

> Addons, future DarkRP updates etc. could change this.

In the case of addons, that's their decision, and the Server Operator's too if they wish for that. In the case of future DarkRP updates, that's not for me to decide. My intention here was to keep changes to a minimum whilst providing a little more flexibility for others wishing to use their own skin, without having to compromise too much due to hard coding.

> The current transparent background for text entries currently is applied to all DTextEntry elements with the DarkRP skin which is why a global colour change makes sense.

Fair enough, that makes sense. My knowledge on such within DarkRP is very limited, especially compared with yours. If you'd like, I can revert my change to the F1 menu so you can instead make a pull with yours instead.

> The only case I see it being a problem, because of the transparent background of the element, is if someone used a light background in the parent but then they are no longer using the DarkRP colour scheme (ie. they are using something like a hardcoded SetBackgroundColor/SetBGColor)

I don't quite understand. Are you referring to people who actually modify how the panel looks via modifying the F1 menu itself, such as making it a module of theirs?

Bo98:
> Yeah, it shouldn't do. I believe it's to do with SetSkin being used on panels that change the skin to something different than what ForceDermaSkin is set to use.

Ah, `SetSkin` will have not yet applied when `PANEL:Init` is called while `ForceDermaSkin` will have. It should be fine later on though so it should have no visual effect unless the control is broken and doesn't fetch the SKIN fields after `PANEL:Init`.

But you did remind me that it was DTextEntry that was broken. See https://github.com/garrynewman/garrysmod/pull/1141. ETA on fix is probably 2018-19 I'm afraid though.

Luckily, there is a very simply workaround which would be to apply the `SKIN:PaintTextEntry` changes into the DarkRP skin. It is forwards compatible, requires a minuscule amount of effort, and will have the exact same effect as if it was incorporated in the default GMod skin. I wouldn't even call it "hacky" either. There are much hackier workarounds in DarkRP currently. Of course, it all comes down to what @FPtje prefers though.

---

To clarify the other points:

> whilst providing a little more flexibility for others wishing to use their own skin

Changes to the DarkRP skin does not affect this in any way. Baseclasses don't exist in the SKIN world.

> The change made here shouldn't cause conflictions with anything at least

Neither solution should cause conflicts.

> although a global variable could be defined which holds the name instead (which the skin itself uses for the name), but I doubt it's worth it

Definitely not.

> the whole "staying local" idea

Addressed in my previous comment. It's not a bad mindset though in some cases. I just disagree in it applying here.

gamemode/modules/f1menu/cl_searchbox.lua
Bo98: > Yeah, it shouldn't do. I believe it's to do with SetSkin being used on panels that change the skin to something different than what ForceDermaSkin is set to use.

Ah, `SetSkin` will have not yet applied when `PANEL:Init` is called while `ForceDermaSkin` will have. It should be fine later on though so it should have no visual effect unless the control is broken and doesn't fetch the SKIN fields after `PANEL:Init`.

But you did remind me that it was DTextEntry that was broken. See https://github.com/garrynewman/garrysmod/pull/1141. ETA on fix is probably 2018-19 I'm afraid though.

Luckily, there is a very simply workaround which would be to apply the `SKIN:PaintTextEntry` changes into the DarkRP skin. It is forwards compatible, requires a minuscule amount of effort, and will have the exact same effect as if it was incorporated in the default GMod skin. I wouldn't even call it "hacky" either. There are much hackier workarounds in DarkRP currently. Of course, it all comes down to what @FPtje prefers though.

---

To clarify the other points:

> whilst providing a little more flexibility for others wishing to use their own skin

Changes to the DarkRP skin does not affect this in any way. Baseclasses don't exist in the SKIN world.

> The change made here shouldn't cause conflictions with anything at least

Neither solution should cause conflicts.

> although a global variable could be defined which holds the name instead (which the skin itself uses for the name), but I doubt it's worth it

Definitely not.

> the whole "staying local" idea

Addressed in my previous comment. It's not a bad mindset though in some cases. I just disagree in it applying here.

Hazbelll: > Ah, SetSkin will have not yet applied when PANEL:Init is called while ForceDermaSkin will have. It should be fine later on though so it should have no visual effect unless the control is broken and doesn't fetch the SKIN fields after PANEL:Init

Yeah, makes sense that. You see it now and again within various addon's GUIs still.

> Changes to the DarkRP skin does not affect this in any way. Baseclasses don't exist in the SKIN world.

Oh, that's good then. I don't normally work with skins at all anyway. Decided not to be daring with what was done due to that, haha. Again, more than happy to revert my change so you can instead make a pull with yours if you so wish.

> Definitely not.
Addressed in my previous comment.

Indeed:
> (also ironically goes against the whole "staying local" idea)

Bo98: > You see it now and again within various addon's GUIs still.

Unless they are using a custom control with a broken implementation, it should only be broken for DTextEntry. Or are the GUIs you are mentioning all involving DTextEntry?

Hazbelll: Yeah, although I think a few do use their own custom controls. Not sure, never looked into it properly to be honest.

FPtje:
I generally agree with Bo's approach. It allows for more customisation while providing the same benefits.

PR #2382, opened 2016-03-25
Improved DarkRP.isEmpty

fruitwasp:


FPtje:
There's only one line change, why is the entire function weirdly indented?

fruitwasp:
Linter warnings

FPtje:
So the linter forces you to change 4 spaces into 2?

fruitwasp:
Not forced, just a warning

FPtje:
glualint only warns when you use both tabs and spaces for indentation. Not when you use two spaces and/or four.

fruitwasp:
Believe me or not. Anyway, reverted

FPtje:
I'm the author of glualint. It only warns for inconsistent use of tabs and spaces.

fruitwasp:
I know you are. As I said, believe me or not. Leave it, please.

PR #2415, opened 2016-04-15
Allowing Admins Lockdown/UnLockdown

MaDmaxwell311:
Granting ability to Admins to lockdown and unlockdown.
I would add if to the CAMI privilege of DarkRP_AdminCommands but I was unsure if what I have will work. What I had is this     CAMI.PlayerHasAccess(ply, "DarkRP_AdminCommands", function(access) 
     if not access then return end 
)
If it does, I will change the Commit (if possible).

### DIFF
--- a/gamemode/modules/police/sv_commands.lua
+++ b/gamemode/modules/police/sv_commands.lua
@@ -134,7 +134,7 @@ function DarkRP.lockdown(ply)
         return ""
     end
 
-    if ply:EntIndex() ~= 0 and not ply:isMayor() then
+    if ply:EntIndex() ~= 0 and not ply:isMayor() or ply:IsAdmin() then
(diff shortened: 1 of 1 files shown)


Shigbeard:
I'd argue that admins should have the ability to remove lockdowns and not start them (say, a mayor sets a lockdown for an RP reason, but goes AFK before the lockdown is due to end).. however the demotion of a mayor is supposed to end a lockdown isn't it?

IMO admin's don't need lockdown control, this is a feature better suited to a custom admin mod and not to a core darkrp change, especially one such as this.

MaDmaxwell311:
Good point, I should have changed the unLockdown.

FPtje:
Leave it under control of the mayors. The admins are powerful enough as it is.

PR #2427, opened 2016-04-26
Hook on payday

a1steaksa:
I'd like to keep idle players from collecting a paycheck, but there isn't a hook for it that I can see.

Ideally the hook would return the player and the amount of money they're getting, with the option to return a new amount of money to override.


Bo98:
You're supposed to create an issue, not a pull request.

Bo98:
Anyway, the hook already exists:

http://wiki.darkrp.com/index.php/Hooks/Server/playerGetSalary

PR #2445, opened 2016-05-19
Added universal playerCanPickUp hook

fruitwasp:
This hook can be used for every entity. It should stop the proliferation of  `canPickUp` hooks.

I also think `spawned_weapon` should use this too, but replacing it with the existing `shouldntPickUpDarkRPWeapon` will definitely break addons.

### DIFF
--- a/entities/entities/darkrp_cheque/init.lua
+++ b/entities/entities/darkrp_cheque/init.lua
@@ -20,6 +20,12 @@ end
 
 
 function ENT:Use(activator, caller)
+    local canUse, reason = hook.Call("canDarkRPUse", nil, activator, self)
+    if canUse == false then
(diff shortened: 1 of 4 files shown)


FPtje:
`playerCanPickup` sounds like what happens when you press E on a small prop: something similar to a gravgun pickup.

Can a better name be thought of?

fruitwasp:
`playerCanUse` perhaps?

FPtje:
Perhaps too generic. Maybe something that entails the fact that it's about picking up DarkRP items.

fruitwasp:
`playerCanDarkRPUse`?

fruitwasp:
updated & squashed

PR #2448, opened 2016-05-24
added chat command alias sellalldoors

fruitwasp:
unownalldoors is existing command

### DIFF
--- a/gamemode/modules/doorsystem/sh_doors.lua
+++ b/gamemode/modules/doorsystem/sh_doors.lua
@@ -243,6 +243,8 @@ DarkRP.declareChatCommand{
     delay = 1.5
 }
 
+DarkRP.chatCommandAlias("unownalldoors", "sellalldoors")
+
 DarkRP.declareChatCommand{
     command = "title",
     description = "Set the title of the door you're looking at.",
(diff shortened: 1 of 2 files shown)


FPtje:
You don't need to re-declare the command if you use `chatCommandAlias`

fruitwasp:
removed that and squashed commits

Bo98:
Shouldn't the `DarkRP.chatCommandAlias` call be shared rather than just serverside? Like it is now looks like you would not get the clientside benefits from declaring chat commands (F1 menu).

fruitwasp:
you're right, I will move it.

I doubt my decision to use `chatCommandAlias` is the right one though, as most command aliases are defined and declared as new chat commands. This change makes the code inconsistent.

Example:

jailpos: https://github.com/FPtje/DarkRP/blob/a20464ac789824b26df1aad2c9c3834d6f702b30/gamemode/modules/positions/sh_commands.lua#L22
setjailpos: https://github.com/FPtje/DarkRP/blob/a20464ac789824b26df1aad2c9c3834d6f702b30/gamemode/modules/positions/sh_commands.lua#L28

FPtje:
If it prevents duplicate code, why not

fruitwasp:
moved chat command alias to shared

PR #2458, opened 2016-05-28
Only create property tax timer if property tax enabled

fruitwasp:


FPtje:
Have you tested your shit this time?

PR #2476, opened 2016-06-15
Add health and armor entities

sweepyoface:
Can change whatever, but I think these are useful


FPtje:
For a third party addon, maybe, but not in DarkRP by default.

MrBrax:
either way, naming the entities so loosely isn't a very good idea

PR #2483, opened 2016-06-21
Update init.lua

kiruthekiller:
Because billboard should be placed as it is necessary, and not stand in front of his eyes like a pile of shit


FPtje:
No unfreezing. That only kills people behind them.

PR #2494, opened 2016-06-26
Nice looking vote VGUI

kiruthekiller:
Now 20% cooler than old!


FPtje:
It doesn't work properly:
http://i.imgur.com/KQtbAww.png

It's not in the style of DarkRP

This should be an addon.

Shigbeard:
The only update the vote vgui needs is for consecutive votes to be 
positioned right. When you accept 1, all consecutive votes get placed in 
a slightly different position. Hard to explain this early in the 
morning, but get like two votes going (say, a vote for mayor and a vote 
to demote) and you will see what I mean.

FPtje:
I know what you mean

PR #2495, opened 2016-06-26
Human hobo sounds

kiruthekiller:


FPtje:
If I wanted hobos to make human sounds I'd have done that myself. The disgusting zombie sounds is what makes the hobos good.

PR #2496, opened 2016-06-26
Different money models depending on the amount

kiruthekiller:
If u cancel this, i'll find your home and poo under the door >:C


FPtje:
What about these things being like workshop addons?
Why must all DarkRP servers have this?

PR #2508, opened 2016-06-29
made easier to derive entity from darkrp billboard

fruitwasp:


entities/entities/darkrp_billboard/cl_init.lua
```diff
@@ -1,39 +1,55 @@
...
+    self.DrawPos = Vector(1, -111, 58)
+
+    self.Width = 558
+    self.Height = 290
+
+    self.HeaderMargin = 10
+    self.BodyMargin = 10
+
+    self.HeaderFont = "Trebuchet48"
+    self.HeaderFontSize = 48
```
Bo98: Would `draw.GetFontHeight(self.HeaderFont)` work? (not tested)

fruitwasp: didn't know about that one. nice one!

fruitwasp: I case of an invalid font, it will spam errors though.

Bo98: Does the drawing not do that anyway?

fruitwasp: yah, true.

PR #2523, opened 2016-07-16
Distance -> DistToSqr

KingofBeast:
I didn't know how you'd want to prepare minHitDistance so I have it preparing at the bottom of config.lua with a nice big banner. That should be the only questionable thing here.

### DIFF
--- a/entities/entities/money_printer/init.lua
+++ b/entities/entities/money_printer/init.lua
@@ -75,13 +75,13 @@ end
 
 function ENT:Fireball()
     if not self:IsOnFire() then self.burningup = false return end
-    local dist = math.random(20, 280) -- Explosion radius
+    local dist = math.random(400, 78400) -- Explosion radius (20, 280) ^ 2
     self:Destruct()
     for k, v in pairs(ents.FindInSphere(self:GetPos(), dist)) do
         if not v:IsPlayer() and not v:IsWeapon() and v:GetClass() ~= "predicted_viewmodel" and not v.IsMoneyPrinter then
             v:Ignite(math.random(5, 22), 0)
         elseif v:IsPlayer() then
-            local distance = v:GetPos():Distance(self:GetPos())
+            local distance = v:GetPos():DistToSqr(self:GetPos())
             v:TakeDamage(distance / dist * 100, self, self)
         end
     end
--- a/entities/entities/money_printer/init.lua
+++ b/entities/entities/money_printer/init.lua
@@ -75,13 +75,13 @@ end
 
 function ENT:Fireball()
     if not self:IsOnFire() then self.burningup = false return end
-    local dist = math.random(400, 78400) -- Explosion radius (20, 280) ^ 2
+    local dist = math.random(20, 280) -- Explosion radius
     self:Destruct()
     for k, v in pairs(ents.FindInSphere(self:GetPos(), dist)) do
         if not v:IsPlayer() and not v:IsWeapon() and v:GetClass() ~= "predicted_viewmodel" and not v.IsMoneyPrinter then
             v:Ignite(math.random(5, 22), 0)
         elseif v:IsPlayer() then
-            local distance = v:GetPos():DistToSqr(self:GetPos())
+            local distance = v:GetPos():Distance(self:GetPos())
             v:TakeDamage(distance / dist * 100, self, self)
         end
     end
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -500,9 +500,3 @@ Hungermod module
 GM.Config.hungerspeed = 2
 -- starverate <Amount> - How much health that is taken away every second the player is starving  (3 is the default).
 GM.Config.starverate = 3
-
---[[---------------------------------------------------------------------------
-DO NOT TOUCH THIS (unless you want broken shit!)
----------------------------------------------------------------------------]]
--- Prepare minHitDistance for DistToSqr
-GM.Config.minHitDistance = GM.Config.minHitDistance ^ 2
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -500,3 +500,9 @@ Hungermod module
 GM.Config.hungerspeed = 2
 -- starverate <Amount> - How much health that is taken away every second the player is starving  (3 is the default).
 GM.Config.starverate = 3
+
+--[[---------------------------------------------------------------------------
+DO NOT TOUCH THIS (unless you want broken shit!)
+---------------------------------------------------------------------------]]
+-- Prepare minHitDistance for DistToSqr
+GM.Config.minHitDistance = GM.Config.minHitDistance ^ 2
--- a/gamemode/modules/base/sh_util.lua
+++ b/gamemode/modules/base/sh_util.lua
@@ -128,7 +128,7 @@ end
 
 function meta:getEyeSightHitEntity(searchDistance, hitDistance, filter)
     searchDistance = searchDistance or 100
-    hitDistance = hitDistance or 15
+    hitDistance = (hitDistance or 15) ^ 2
     filter = filter or function(p) return p:IsPlayer() and p ~= self end
 
     self:LagCompensation(true)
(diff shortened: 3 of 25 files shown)


entities/entities/money_printer/init.lua
```diff
@@ -75,13 +75,13 @@ end
...
-    local dist = math.random(20, 280) -- Explosion radius
+    local dist = math.random(400, 78400) -- Explosion radius (20, 280) ^ 2
     self:Destruct()
     for k, v in pairs(ents.FindInSphere(self:GetPos(), dist)) do
         if not v:IsPlayer() and not v:IsWeapon() and v:GetClass() ~= "predicted_viewmodel" and not v.IsMoneyPrinter then
             v:Ignite(math.random(5, 22), 0)
         elseif v:IsPlayer() then
-            local distance = v:GetPos():Distance(self:GetPos())
+            local distance = v:GetPos():DistToSqr(self:GetPos())
             v:TakeDamage(distance / dist * 100, self, self)
```
FPtje: This might cause a problem. Dividing two squared numbers does not give the same result as two normal numbers. It's a change of behaviour. `10/2 = 5, 100 / 4 = 25`

gamemode/config/config.lua
```diff
@@ -500,3 +500,9 @@ Hungermod module
 GM.Config.hungerspeed = 2
 -- starverate <Amount> - How much health that is taken away every second the player is starving  (3 is the default).
 GM.Config.starverate = 3
+
+--[[---------------------------------------------------------------------------
+DO NOT TOUCH THIS (unless you want broken shit!)
+---------------------------------------------------------------------------]]
+-- Prepare minHitDistance for DistToSqr
+GM.Config.minHitDistance = GM.Config.minHitDistance ^ 2
```
FPtje: No. No, no, no, no, no, no, no. 

Don't make config options for things you don't want people to configure. That should be a local variable if used in one place, and some function if used in more than one file. 

Also, never write `X ^ 2`.It's really bad practice. Write  `X * X` instead. It's about the same performance difference as square root, no square root. Powers take more cycles than multiplications. 

No real noticeable difference in this case, but bad practice nonetheless. I'm not contradicting my own arguments here.

gamemode/modules/base/sh_util.lua
```diff
@@ -128,7 +128,7 @@ end
 
 function meta:getEyeSightHitEntity(searchDistance, hitDistance, filter)
     searchDistance = searchDistance or 100
-    hitDistance = hitDistance or 15
+    hitDistance = (hitDistance or 15) ^ 2
```
FPtje: Same here

entities/entities/money_printer/init.lua
KingofBeast: True, I didn't put a lot of thought into that. This may be one that needs to remain pure Distance.

KingofBeast: Keeping DistToSqr would mean adding math.sqrt to TakeDamage and nullify the gains. Will revert.

gamemode/modules/base/sh_util.lua
KingofBeast: Review now

FPtje:
You missed the config comment. 

Also, I won't be able to merge this pr for a while, even after you fixed everything. The reason for that is unrelated to you, the pr or the discussion we've had.

KingofBeast:
I localized it in all 3 places to prevent function calls despite the comment, only because I don't see it changing in the foreseeable future. Where would you want that sort of util function at? The hitmenu module doesn't appear to have any other helper functions so it'd look a little out of place.

FPtje:
That looks good actually, though I hope no external addons use that config option.

FPtje:
Never mind that last comment. I'm tired.

gamemode/config/config.lua
wyozi: Presumably LuaJIT (or even the compiler?) would easily optimize a case as simple as `x^2`. In my opinion `x^2` conveys what you're trying to do better than `x*x` especially if `x` was a more complex variable name. To me this falls under category of premature optimization albeit it is a very small one.

Of course I come without benchmarks or any proof so happy to be proven wrong etc.

Bo98: Unless I did something wrong, it is about the same:

```
] profile_distance 
Starting profiling...
Distance:	Total = 0.012140291107926s	Average = 1.2140291107926e-007s
DistToSqr:	Total = 0.011352322906305s	Average = 1.1352322906305e-007s
DistToSqr with var ^ 2:	Total = 0.011833091029303s	Average = 1.1833091029303e-007s
DistToSqr with var * var:	Total = 0.011812867023991s	Average = 1.1812867023991e-007s
Profiling finished.
] profile_distance 
Starting profiling...
Distance:	Total = 0.012508931202319s	Average = 1.2508931202319e-007s
DistToSqr:	Total = 0.011544066955139s	Average = 1.1544066955139e-007s
DistToSqr with var ^ 2:	Total = 0.011952387059864s	Average = 1.1952387059864e-007s
DistToSqr with var * var:	Total = 0.011959299061573s	Average = 1.1959299061573e-007s
Profiling finished.
] profile_distance 
Starting profiling...
Distance:	Total = 0.012022019077676s	Average = 1.2022019077676e-007s
DistToSqr:	Total = 0.011507970946241s	Average = 1.1507970946241e-007s
DistToSqr with var ^ 2:	Total = 0.011819267025885s	Average = 1.1819267025885e-007s
DistToSqr with var * var:	Total = 0.011819011025636s	Average = 1.1819011025636e-007s
Profiling finished.
] profile_distance 
Starting profiling...
Distance:	Total = 0.011998467071635s	Average = 1.1998467071635e-007s
DistToSqr:	Total = 0.01154995495699s	Average = 1.154995495699e-007s
DistToSqr with var ^ 2:	Total = 0.012268035140551s	Average = 1.2268035140551e-007s
DistToSqr with var * var:	Total = 0.011923971052511s	Average = 1.1923971052511e-007s
Profiling finished.
] profile_distance 
Starting profiling...
Distance:	Total = 0.012259331138466s	Average = 1.2259331138466e-007s
DistToSqr:	Total = 0.011466498935306s	Average = 1.1466498935306e-007s
DistToSqr with var ^ 2:	Total = 0.011808259022928s	Average = 1.1808259022928e-007s
DistToSqr with var * var:	Total = 0.012045059083448s	Average = 1.2045059083448e-007s
Profiling finished.
```

```lua
local function Distance(a, b)
	return a:Distance(b) > 400
end

local function DistToSqr(a, b)
	return a:DistToSqr(b) > 160000
end


local var = 400

local function DistToSqrVarPow(a, b)
	return a:DistToSqr(b) > var ^ 2
end

local function DistToSqrVarMul(a, b)
	return a:DistToSqr(b) > var * var
end

------------------------------------------

local function runBenchmark(f, a, b)
	local t = SysTime()
	for i = 1, 100000 do
		f(a, b)
	end
	return SysTime() - t
end

------------------------------------------

local function printResult(name, total)
	print(name .. ":", "Total = " .. total .. "s", "Average = " .. total / 100000 .. "s")
end

concommand.Add("profile_distance", function()
	local a = Vector(math.Rand(-1000, 1000), math.Rand(-1000, 1000), math.Rand(-1000, 1000))
	local b = Vector(math.Rand(-1000, 1000), math.Rand(-1000, 1000), math.Rand(-1000, 1000))

	print("Starting profiling...")
	printResult("Distance", runBenchmark(Distance, a, b))
	printResult("DistToSqr", runBenchmark(DistToSqr, a, b))
	printResult("DistToSqr with var ^ 2", runBenchmark(DistToSqrVarPow, a, b))
	printResult("DistToSqr with var * var", runBenchmark(DistToSqrVarMul, a, b))
	print("Profiling finished.")
end)
```

Bo98:
FYI, you didn't send the FSpectate changes to https://github.com/FPtje/FSpectate so they have now been reverted by a merge.

FPtje:
Oh, should serve the fucker right for changing fspectate files in the
DarkRP repository.

PR #2525, opened 2016-07-22
Updated Stunstick

roboderpy:
Make player force to exit vehicle if player is not covered by vehicle itself.
e.g. when hitting jeep with player driver it do nothing, when hitting player sitting on ground, it forces player to exit vehicle.

### DIFF
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -67,7 +67,7 @@ function SWEP:DoFlash(ply)
     if not IsValid(ply) or not ply:IsPlayer() then return end
 
     ply:ScreenFade(SCREENFADE.IN, color_white, 1.2, 0)
-    ply:ExitVehicle()
+    ply:ExitVehicle() --If player sitting on ground or in uncovered vehicle, get him out (like sitting on ground)
 end
 
 function SWEP:PostDrawViewModel(vm)
--- a/entities/weapons/stunstick/shared.lua
+++ b/entities/weapons/stunstick/shared.lua
@@ -67,24 +67,25 @@ function SWEP:DoFlash(ply)
     if not IsValid(ply) or not ply:IsPlayer() then return end
 
     ply:ScreenFade(SCREENFADE.IN, color_white, 1.2, 0)
+    ply:ExitVehicle()
 end
 
 function SWEP:PostDrawViewModel(vm)
-    if self:GetSeqIdleTime() ~= 0 or self:GetLastReload() >= CurTime() - 0.1 then
-        local attachment = vm:GetAttachment(1)
-        local pos = attachment.Pos
-        cam.Start3D(EyePos(), EyeAngles())
-            render.SetMaterial(Material("effects/stunstick"))
-            render.DrawSprite(pos, 12, 12, Color(180, 180, 180))
-            for i = 1, 3 do
-                local randVec = VectorRand() * 3
-                local offset = (attachment.Ang:Forward() * randVec.x) + (attachment.Ang:Right() * randVec.y) + (attachment.Ang:Up() * randVec.z)
-                render.SetMaterial(Material("!darkrp/stunstick_beam"))
-                render.DrawBeam(pos, pos + offset, 3.25 - i, 1, 1.25, Color(180, 180, 180))
-                pos = pos + offset
-            end
-        cam.End3D()
-    end
+    if not (self:GetSeqIdleTime() ~= 0 or self:GetLastReload() >= CurTime() - 0.1) then return end
+	
+    local attachment = vm:GetAttachment(1)
+    local pos = attachment.Pos
+    cam.Start3D(EyePos(), EyeAngles())
+        render.SetMaterial(Material("effects/stunstick"))
+        render.DrawSprite(pos, 12, 12, Color(180, 180, 180))
+        for i = 1, 3 do
+            local randVec = VectorRand() * 3
+            local offset = (attachment.Ang:Forward() * randVec.x) + (attachment.Ang:Right() * randVec.y) + (attachment.Ang:Up() * randVec.z)
+            render.SetMaterial(Material("!darkrp/stunstick_beam"))
+            render.DrawBeam(pos, pos + offset, 3.25 - i, 1, 1.25, Color(180, 180, 180))
+            pos = pos + offset
+        end
+    cam.End3D()
 end
 
 function SWEP:DrawWorldModelTranslucent()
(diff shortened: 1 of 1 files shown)


mcNuggets1:
Useless..

Shigbeard:
No, what about the sit anywhere mod? Sit anywhere's default settings allow a player to sit down and instantly become invulnerable. They cant be arrested, they cant be shot, etc etc etc.

If I'm not mistaken, this would allow cops to stunstick people out of their seat, countering Sit Anywhere.

Unless I'm wrong @roboderpy

roboderpy:
Yes, you are right. Because of that, in my alternative to sit anywhere that is optimized for DarkRP servers players is hardcoded to receive damage, stick arrests and etc.

mcNuggets1:
You can get arrested and get killed in sit anywhere mod.

roboderpy:
Only if you find out there is a cvar to enable player damage.

mcNuggets1:
That is not that difficult..

roboderpy:
Instead of just saying multiple times "go away" or even killing player cops now can simply force player to stand up and go away using stunstick

roboderpy:
No, my addon is coded from zero and have nothing generic with sit anywhere

Bo98:
Why not use the door ram?

roboderpy:
Using stunstick it would be easier to cops

Shigbeard:
Door rams often require warrants on some servers. Why would you need a warrant to make a person stand up from a seated position, when there is no door between you and them (or an imaginary door, in the hl2 jeep's case).

Bo98:
That's solvable with hooks but I guess your average server owner doesn't use them.

Since it was mentioned, it is worth noting that player damage while in a seat should be fixed in the next GMod update.

entities/weapons/stunstick/shared.lua
```diff
@@ -67,24 +67,25 @@ function SWEP:DoFlash(ply)
...
-            for i = 1, 3 do
-                local randVec = VectorRand() * 3
-                local offset = (attachment.Ang:Forward() * randVec.x) + (attachment.Ang:Right() * randVec.y) + (attachment.Ang:Up() * randVec.z)
-                render.SetMaterial(Material("!darkrp/stunstick_beam"))
-                render.DrawBeam(pos, pos + offset, 3.25 - i, 1, 1.25, Color(180, 180, 180))
-                pos = pos + offset
-            end
-        cam.End3D()
-    end
+    if not (self:GetSeqIdleTime() ~= 0 or self:GetLastReload() >= CurTime() - 0.1) then return end
```
FPtje: This is not how you properly invert compound conditions.

entities/weapons/stunstick/shared.lua
```diff
@@ -67,24 +67,25 @@ function SWEP:DoFlash(ply)
     if not IsValid(ply) or not ply:IsPlayer() then return end
 
     ply:ScreenFade(SCREENFADE.IN, color_white, 1.2, 0)
+    ply:ExitVehicle() --If player sitting on ground or in uncovered vehicle, get him out (like sitting on ground)
```
FPtje: Grammar in comment

FPtje:
I added some line comments, but I some more things to say about this:

This would make the ram out of vehicle feature of the battering rams obsolete. That doesn't sit right. A stunstick is for warnings, threatening and slaps on the wrist. It is not for kicking people out of seats or any vehicle. It makes much more sense for a battering ram to do that.

roboderpy:
But why you need to use battering ram to just make player stand up, when you just wanted to stun him using stunstick.

FPtje:
Because a battering ram is a ram. It rams. A stunstick performs
disciplinary actions. It smacks. Smacking does not necessarily remove one
from a vehicle. A ram on the other hand can.

I'm not making features obsolete because you want it to be slightly easier.

FPtje:
Looks good, will merge later. 

And by later I mean longer than you might expect. Think of like mid August or something.

roboderpy:
Nothing bad with that

lavnart:
Please, add a variable to enable/disable this functionality.

Kefta:
You can fix it yourself. Source chair behaviour is ridiculous.

Shigbeard:
I agree that door rams should be used exclusively for doors and vehicles, but on the edge case where the vehicle is not actually a vehicle and is actually just a seat, I feel a stunstick is appropriate. Add to that the usage of the Sit Anywhere mod, where players can press Alt + E on any flat horizontal surface, you find that using a Stunstick to beat the shit out of a disobedient player (attempting to abuse the fact that Sit Anywhere prevents you from taking damage unless correctly configured by the (often dim witted) server owner) makes much more sense than using a Door Ram to knock their guts into next week.

The point I try to make here is that a stunstick is suitable in the event that the vehicle in question doesn't have a door, imaginary or real (For example: Airboat seat, sit anywhere, jeep seat, etc etc). You can apply damage at the same time as pulling the person up off their ass (And fyi, I have seen people use Sit Anywhere to avoid being arrested or attacked). You cannot expect the developer/s of Sit Anywhere to correct this as they already have a feature built in to circumvent such abuse, however it's default behavior sees it disabled and most server owners are either oblivious or clueless (or in the rare case, ignorant) to the presence of a server-side convar to change this.

As long as this stun stick behavior is applied exclusively to Sit Anywhere, or vehicles that have no means of motion (read: vehicles that are just seats and nothing more), then it would not be rendering the Door Ram obsolete in any way, rather bypassing an annoying system where a police officer would have to request a search warrant and be held behind `PrimaryAttack` and `SecondaryAttack` delays as well as impaired movement speed in order to remove a player from their seat or random ledge, and even after all that, during said delays the player could just as easily return to their seated position starting the entire cycle anew.

Granted, this abuse should be solvable through adequate server moderation, however as we are all well aware, DarkRP admins and moderators are very rarely adequate in any sense of the word.

FPtje:
Arresting seated players should work and the ram should be able to kick
people out without a warrant. If either or both of those are not currently
true I'd rather fix that.

FPtje:
Both are true.

roboderpy:
Instead of switching to door ram and kicking player out from invisible vehicle and switching back to stun stick you can just simply slap player one time with stunstick

FPtje:
Are you seriously saying that the effort of switching weapons is *the* reason why stunsticks should adopt door rams' behaviour?

Kefta:
@FPtje Have a good vacation?

FPtje:
I'm back

roboderpy:
It doesn't. You are still not able to get player out of his vehicle if he is in the jeep or airboat with stunstick

FPtje:
That is true. Neither are you with the unarrest stick, the bugbait swep and the /write chat command. There is no reason for any of those weapons to be able to perform such action. After all, the door ram is best suited for that functionality. We don't need two weapons to do the same fucking thing.

roboderpy:
If stunstick is able to move people when they are standing on ground and not able to move them when the just fucking sitting on ground and that is okay to you, i will stop complaing about it

FPtje:
Something about enlarged resting surface when sitting down.

Shigbeard:
The battering ram usually needs a search warrant for any of its 
functionality to work, unless the server owner is half competent (And we 
are talking about DarkRP here... so no competency is a given). To go 
through all that BS to knock them off their seat, just to have them sit 
back down before you can get your arrest stick out and arrest them... 
thats BS.

+1 this pr

FPtje:
The battering ram can ram players out of seats without a warrant. I've said that earlier today.

Shigbeard:
"If the server owner is half competent".

I've been on literally 2 dozen servers where it doesn't do that Falco.

roboderpy:
And they always run 2.4 for some reason

FPtje:
> "If the server owner is half competent".
> I've been on literally 2 dozen servers where it doesn't do that Falco.

No possible configuration causes door rams not to work without warrants save from returning a value in the `canDoorRam` hook.

roboderpy:
It means that server owners is incridebly dumb and grants sitting players status of unmovable object

FPtje:
It means that someone is doing weird shit in the `canDoorRam` hook, at which point my responsibility is dissolved.

The possible fact that some incompetent server owners might curb the door ram's functionality by writing incompetent code in the provided hook does not fly as a valid reason for another weapon to have the same functionality.

PR #2551, opened 2016-09-23
Foreseeable future

Velkon:
pm me if u wanna know more

### DIFF
--- a/gamemode/modules/workarounds/sh_workarounds.lua
+++ b/gamemode/modules/workarounds/sh_workarounds.lua
@@ -192,6 +192,13 @@ hook.Add("InitPostEntity", "FixDoorTool", function()
             oldFunc(ply, trace, ang, model, open, close, autoclose, closetime, class, hardware, ...)
         end
     end
+	if game.GetMap() == "rp_downtown_v4c_v2" then 
+		for k,v in pairs(ents.GetAll()) do
+			if v:GetClass() == "lua_run" then
+				v:Remove()
+			end
+		end
+	end
 end)
 
 --[[---------------------------------------------------------------------------


Shigbeard:
Sounds like exploits.lua

EstevanTH:
Seriously, this is not really the right place to fix exploits from maps...
Maybe that might be a good idea to create a table of backdoored maps instead, let's wait for Falco's opinion.
Also I would use `ipairs()` instead of `pairs()`, which is faster.

Shigbeard:
`ipairs()` is only faster when talking about hundreds of entities. If we are talking about a fresh map, it may only have to go through 30 entities. It's a minor optimization. Don't get falco started on the minor optimizations, he'll chew your ear off.

FPtje:
What do you mean a table of backdoored maps? Do you mean a table in DarkRP's database with DarkRP code checking that database and doing the exact same thing with the entries as it's doing right now? Do you mean a Lua table?

Also, +1 on @Shigbeard about micro optimisations. It's not worth mentioning here.
Why would you argue about making a table when your point was that this is not the right place to do it in the first place?

Shigbeard:
Huh, this new pull request review shit is rather fancy.

EstevanTH:
Yes a Lua table, like
```lua
backdoored_maps = {
	["rp_downtown_v4c_v2"]=true,
}
```
That would be kind of a good idea to stop exploits.

FPtje:
Not all `lua_run` entities are bad. They were designed for a reason. Seriously, though, don't clients with a different version of the map should disconnect with some kind of map mismatch error? I vaguely remember seeing those errors many years ago.

@EstevanTH There's no need for abstraction like that just yet. Not for one map. Some other maps might have other exploits that require different Lua code to solve. The reason why the code shouldn't account for that is the same as to why it shouldn't account for multiple maps.

Shigbeard:
@FPtje Yes, however I still don't fully understand the nature of the exploit. I'm assuming it's an exploit that can be used to attack players, and so a server host may intentionally take the map, rename it making no changes at all, and force clients to re-fetch it. In hindsight I may be wrong in this assumption. I've asked velkon for an explanation but he appears to be offline at this time.

FPtje:
I don't think there's much a map can do that SendLua can't do. Unless it involves gathering data from clients joining other servers.

FPtje:
I've talked to Roqj, the one who hosts the map on the workshop. Based on that conversation I've decided against merging this PR. The reasons are as follows:

- The map currently doesn't contain any exploits
- Merging this PR would be an implicit endorsement of the accusation that Roqj would backdoor rp_downtown_v4c_v2. I can see a clear possible conflict of interest, but no direct evidence that Roqj would backdoor the map in the near future.
- Roqj can lie about not backdooring the map, but even if the map were to get a backdoor, it can easily be seen in the workshop updates page. That or the map would be re-uploaded, which is also a red flag. 
- The process of reporting the map, having it taken down and having a new version uploaded without the exploit would be much, much faster, more effective and more appropriate than some hack in DarkRP. 
- Roqj could very easily upload a similar map with a different name (`rp_downtown_v4c_v4` or something) with the exploit, easily bypassing this check.

With the conflict of interest I'd recommend keeping a close eye on Roqj's uploads and versions of the downtown map uploaded by possible alt accounts, but so far Roqj has done nothing wrong and I won't act as though he has.

PR #2562, opened 2016-10-10
Make spawned_money work with all models

ARitz-Cracker:
I was creating my own spawned money entity for ARCBank to make the thing cross-gamemode compatible. I decided to look at the DarkRP spawned money to see how that worked. So, I took that entity, added some stuff I did to my entity, and now I am giving back (part) of what I did to make it better!

### DIFF
--- a/entities/entities/spawned_money/cl_init.lua
+++ b/entities/entities/spawned_money/cl_init.lua
@@ -3,27 +3,25 @@ include("shared.lua")
 function ENT:Initialize()
 end
 
+local vector_up = Vector(0,0,1)
 function ENT:Draw()
     self:DrawModel()
+    local offset = (self:OBBMaxs() - self:OBBCenter())*vector_up
 
-    -- Do not draw labels when a different model is used.
-    -- If you want a different model with labels, make your own money entity and use GM.Config.MoneyClass.
-    if self:GetModel() ~= "models/props/cs_assault/money.mdl" then return end
-
-    local Pos = self:GetPos()
+    local Pos = self:OBBCenter() 
     local Ang = self:GetAngles()
 
     surface.SetFont("ChatFont")
     local text = DarkRP.formatMoney(self:Getamount())
     local TextWidth = surface.GetTextSize(text)
 
-    cam.Start3D2D(Pos + Ang:Up() * 0.9, Ang, 0.1)
+    cam.Start3D2D(self:LocalToWorld(Pos+offset), Ang, 0.1)
         draw.WordBox(2, -TextWidth * 0.5, -10, text, "ChatFont", Color(140, 0, 0, 100), Color(255, 255, 255, 255))
     cam.End3D2D()
 
     Ang:RotateAroundAxis(Ang:Right(), 180)
 
-    cam.Start3D2D(Pos, Ang, 0.1)
+    cam.Start3D2D(self:LocalToWorld(Pos-offset), Ang, 0.1)
         draw.WordBox(2, -TextWidth * 0.5, -10, text, "ChatFont", Color(140, 0, 0, 100), Color(255, 255, 255, 255))
     cam.End3D2D()
 end


entities/entities/spawned_money/cl_init.lua
```diff
@@ -3,27 +3,25 @@ include("shared.lua")
 function ENT:Initialize()
 end
 
+local vector_up = Vector(0,0,1)
```
FPtje: - Not the same as it was before (`Pos + Ang:Up() * 0.9`)
- Doesn't make it more modifiable

entities/entities/spawned_money/cl_init.lua
```diff
@@ -3,27 +3,25 @@ include("shared.lua")
...
 function ENT:Draw()
     self:DrawModel()
+    local offset = (self:OBBMaxs() - self:OBBCenter())*vector_up
```
FPtje: Spaces around operators. Install glualint.

entities/entities/spawned_money/cl_init.lua
```diff
@@ -3,27 +3,25 @@ include("shared.lua")
...
-    local Pos = self:GetPos()
+    local Pos = self:OBBCenter() 
     local Ang = self:GetAngles()
 
     surface.SetFont("ChatFont")
     local text = DarkRP.formatMoney(self:Getamount())
     local TextWidth = surface.GetTextSize(text)
 
-    cam.Start3D2D(Pos + Ang:Up() * 0.9, Ang, 0.1)
+    cam.Start3D2D(self:LocalToWorld(Pos+offset), Ang, 0.1)
```
FPtje: Spaces around operators. Have you installed glualint yet?

entities/entities/spawned_money/cl_init.lua
```diff
@@ -3,27 +3,25 @@ include("shared.lua")
...
         draw.WordBox(2, -TextWidth * 0.5, -10, text, "ChatFont", Color(140, 0, 0, 100), Color(255, 255, 255, 255))
     cam.End3D2D()
 
     Ang:RotateAroundAxis(Ang:Right(), 180)
 
-    cam.Start3D2D(Pos, Ang, 0.1)
+    cam.Start3D2D(self:LocalToWorld(Pos-offset), Ang, 0.1)
```
FPtje: Spaces around operators. Why haven't you installed glualint yet!?

entities/entities/spawned_money/cl_init.lua
ARitz-Cracker: That is just to isolate the "z" value I get from the objects bounding box. (Which is done a few lines down)

entities/entities/spawned_money/cl_init.lua
ARitz-Cracker: I use notepad++, if you really want to keep the entire DarkRP gamemode that complaint to the specs you defined in glualint, you can edit it yourself if you'd like.

entities/entities/spawned_money/cl_init.lua
ARitz-Cracker: See previous comment

entities/entities/spawned_money/cl_init.lua
ARitz-Cracker: Why are you so judgemental? I just wanted to help! D:

ARitz-Cracker:
I did notice while I was coding it that it assumes the thinnest part it the "z" axis. I _could_ check each axis and determine which side is the smallest. If you'd like me to do that before implementing something like this, I'd be happy too.

Shigbeard:
Also issues that regard the size of the label could be resolved using a config option for scaling the label.

FPtje:
A config option would then also apply to money packets.

What's wrong with making an entity that has the `spawned_money` as base? That's the most generic and appropriate solution.

JamesxX:
Falco, instead of being a pain to people who write code for a gamemode credited to you, simply because of their style of code, you could instead just lint it yourself. If it's as easy as you say, do it yourself.

FPtje:
People making PRs can be reasonably expected to program in the same style as the rest of the project.

JamesxX:
As with anything, but at the end of the day, if people are doing work for free, you can give them some allowance.

Shigbeard:
Holy shit getting this worked up over gaps or lack thereof, jeez I have two recommendations falco.

The first is to go get a cup of coffee. Like, right this second.

The second is to calm the fuck down, it's just a few lines and doesn't even break anything.

JamesxX:
^

entities/entities/spawned_money/cl_init.lua
```diff
@@ -3,27 +3,25 @@ include("shared.lua")
...
         draw.WordBox(2, -TextWidth * 0.5, -10, text, "ChatFont", Color(140, 0, 0, 100), Color(255, 255, 255, 255))
```
JamesxX: Why *0.5 when you could /2. Seems more intuitive, and easier to understand.

FPtje:
The pushy glualint stuff was meant to be a joke. Sorry if it didn't come across as such.

FPtje:
Still though, if you want text drawn over custom models, wouldn't it be better to just create a derived entity? You'd be able to modify everything: position, angle, scale, colour, the actual text etc.

ARitz-Cracker:
Indeed it is. I thought since there was a config option to have custom models, why not have a (basic) way to support them?

entities/entities/spawned_money/cl_init.lua
FPtje: But 1 vs 0.9 though

entities/entities/spawned_money/cl_init.lua
FPtje: Also, OBBMaxs - OBBCenter shouldn't get you to the top middle of an object, which is probably what you'd want. With the standard money packet, it just gets the Pos (which is either bottom middle or center middle) and adds Ang:Up() * 0.9 to it to go to the top middle of the money packet.

OBBMaxs - OBBCenter is the local vector going from the center to the top right. Surely that shouldn't go straight up for most models.

FPtje:
That's a good point.

entities/entities/spawned_money/cl_init.lua
ARitz-Cracker: It doesn't. That's why I multiply it by vector_up. So it does go straight up.

entities/entities/spawned_money/cl_init.lua
ARitz-Cracker: That's why I use self:OBBMaxs() - self:OBBCenter(), to get the local pos of the "highest" position of the model. Multiplying it by 0,0,1 simply isolates the z value while setting the x and y values to 0.

FPtje:
I tested the solution and found that it would make things look worse.

This is what the text looks like in the current master branch. Note the red slab of text being just slightly above the actual model.
hl2_2016-10-15_14-03-57


This is what the text looks like in the PR. Note that the red slab of text is much further above the model.
hl2_2016-10-15_14-02-32

Here's the cause (done with the `cl_ent_bbox` concommand):
hl2_2016-10-15_14-05-17

The bounding box of the model is bigger than the model itself. Putting the text right on top of this bounding box doesn't work. Not for money packets at least. There has to be a specific offset for money packets. That's where the `0.9` comes from in the current code. It also means that there *cannot* be an acceptable generic solution. There is no acceptable "one size fits all". It's way too different per model.

Also, I've tweaked the offset to `0.82`. That's really tight, and still looks good from some distance (with and without anti-aliasing).

PR #2568, opened 2016-10-21
Freeze glitch

PichotM:
Fix freeze glitch suicide


Bo98:
This will probably be better located in a hook in the FAdmin freeze files since this will break if FAdmin is disabled.

FPtje:
Yeah, that's better. Done.

PR #2569, opened 2016-10-23
Added hook to allow disabling player info for individual players

DylanWalseth:
Can hide hud for individual players if you want either a different hud for certain people.

Or, if you're like me and don't like seeing invisible (using ulib's cloak) admin's hud, can use a hook like so
```lua
hook.Add("ShouldDrawPlayerInfo", "ShouldDrawPlayerInfo", function(ply)
    if ply:GetRenderMode() == RENDERMODE_TRANSALPHA then return false end
end)
```

### DIFF
--- a/gamemode/modules/hud/cl_hud.lua
+++ b/gamemode/modules/hud/cl_hud.lua
@@ -265,8 +265,6 @@ Entity HUDPaint things
 -- This syntax allows for easy overriding
 plyMeta.drawPlayerInfo = plyMeta.drawPlayerInfo or function(self)
     local pos = self:EyePos()
-    local shouldDraw = hook.Call("ShouldDrawPlayerInfo", GAMEMODE, self)
-    if shouldDraw == false then return end
 
     pos.z = pos.z + 10 -- The position we want is a bit above the position of the eyes
     pos = pos:ToScreen()
@@ -334,7 +332,7 @@ local function DrawEntityDisplay()
     local aimVec = localplayer:GetAimVector()
 
     for k, ply in pairs(players or player.GetAll()) do
-        if ply == localplayer or not ply:Alive() or ply:GetNoDraw() then continue end
+        if ply == localplayer or not ply:Alive() or ply:GetNoDraw() or hook.Call("ShouldDrawPlayerInfo", GAMEMODE, ply) == false then continue end
         local hisPos = ply:GetShootPos()
         if ply:getDarkRPVar("wanted") then ply:drawWantedInfo() end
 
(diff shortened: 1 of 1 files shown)


Bo98:
This overlaps in functionality with https://github.com/FPtje/DarkRP/blob/353dc3286092cdb2efade2cc0f2145db5a6c2e69/gamemode/modules/hud/cl_hud.lua#L328-329.

```lua
hook.Add("HUDShouldDraw", "HideForAlphaPlayers", function(name)
    if name == "DarkRP_EntityDisplay" then
        local players = {}
        for _, ply in ipairs(player.GetAll()) do
            if ply:GetRenderMode() == RENDERMODE_TRANSALPHA then continue end
            table.insert(players, ply)
        end
        return true, players
    end
end)
```

But this does raise a point that this will likely conflict between addons. What if addon A wants to block one group of players while addon B wants to block a separate group? It would be impossible under this undocumented extension of HUDShouldDraw. Your hook would help here but I'd move it into loop in the DrawEntityDisplay local function so it also prevents the wanted text from drawing.

As a side note for your particular cloaking example, it's already possible to hide the entity display HUD via `ply:SetNoDraw(true)`, which is what FAdmin uses, though I guess how ULib cloaks players is not in your control.

DylanWalseth:
I moved the location to your recommendation, it makes a lot more sense there. Didn't realize the HUDShouldDraw could be returned with the players to draw. That really does make this pull request redundant.

But yea, I was considering ULib cloak, which a decent number of servers seem to favor for things like cloaking.

FPtje:
Too many hooks spoil the broth. There shouldn't be hooks that overlap in features. I agree that returning players in HUDShouldDraw is bad design, but it's there now. 

Side note, you'll never make players fully invisible with ulib cloak. There's always some way to find them. Chat bubbles, the list of people who can hear you talk when you open chat or microphone, wire finders, clientside scripts and all that shit.

Shigbeard:
Think like Harry Potter's invisiblity cloak incase you don't understand 
why !cloak wont hide you perfectly.

Under his cloak, he is invisible to the naked eye. This doesn't mean he 
is undetectable.

- He can still be smelt
- He can still be heard
- Those with some form of ESP can still detect him
- He still manipulates objects around him
- He leaves tracks

This is the same for an admin who is cloaked: there are ways to find a 
cloaked admin aside from visibly observing them. Better to use FAdmin's 
/spectate command, as it allows the admin to move their perspective to 
the player they want to watch, without moving their entity, thus making 
them undetectable.

PR #2575, opened 2016-10-29
added DarkRP.getJobTable(id)

fruitwasp:


FPtje:
Why?

uRandomAlex:
@FPtje It's very useful in hooks like "playerCanChangeTeam", for example. I can't think of any other way to get job table there.

Bo98:
I suppose you could argue there should've been API to return RPExtraTeam rather than using a global variable but it's far too late for that and this pull request only half finishes the job.

FPtje:
Good point.

PR #2587, opened 2016-11-19
Improving code base

Aka7tsu4:
Using gluaint on all lua files.
remove lua comments
https://fptje.github.io/glualint-web/ 

that's a good idea no? (much server use darkrp they need to be optimized) *-*

### DIFF
--- a/gamemode/modules/base/cl_drawfunctions.lua
+++ b/gamemode/modules/base/cl_drawfunctions.lua
@@ -1,4 +1,3 @@
--- concatenate a space to avoid the text being parsed as valve string
 local function safeText(text)
     return string.match(text, "^#([a-zA-Z_]+)$") and text .. " " or text
 end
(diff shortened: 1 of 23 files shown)


Bo98:
I think you changed the line endings for all the files.

https://github.com/sugarcrm/sugarcrm_dev/wiki/Line-Ending-issues-on-Windows-with-git

(and also don't use the default Notepad since it can't handle some line endings)

Bo98:
That said, I'm not sure how removing _all_ Lua comments is an improvement.

Aka7tsu4:
i also changed umsg with net
and suppressing useless space

FPtje:
I don't see the point of this change. Glualint's pretty printer is great, but not perfect. It cannot get the layout right 100% because many layout things depend on the meaning of code. For example, sometimes you'd want to align code [(e.g. here)](https://github.com/FPtje/DarkRP/blob/master/entities/weapons/door_ram/shared.lua#L183-L190) to make it more clear what it does.

Besides, changes in layout aren't an improvement. In the best case the layout is just different. In the worst case it messes up original layout or even introduces bugs. I'm still not 100% sure that the pretty printer has no edge cases where it changes the semantics of code.

PR #2592, opened 2016-12-05
Update sv_util.lua

PichotM:
I got this error on my server


Bo98:
Could you post the full error just in case there's also a bug elsewhere causing the vector to be nil?

FPtje:
That function should not be defined for a nil vector. Any error it throws because of a nil vector is a correct error. Shutting it up about it is the stupidest thing you can do, because it will make very obvious programmer errors invisible and thus very difficult to debug.

PR #2595, opened 2016-12-08
Removed useless NPC check in arrest baton

Kefta:
The method already returns if the entity is not a player above


FPtje:
Can you arrest NPCs with this commit?

Kefta:
You have never been able to. If you want me to change that, I will, but it might break addons/canArrest hooks that don't account for NPCs

FPtje:
It used to work at some point. DarkRP has no use for it though, so
whatever.

PR #2596, opened 2016-12-08
Remove OnRemove reset in baton base

Kefta:
There's no need to reset the materials and sequences of a weapon that is going to be removed

### DIFF
--- a/entities/weapons/stick_base/shared.lua
+++ b/entities/weapons/stick_base/shared.lua
@@ -105,11 +105,6 @@ function SWEP:Holster()
     return true
 end
 
-function SWEP:OnRemove()
-    BaseClass.OnRemove(self)
-    self:ResetStick(true)
-end
-
 function SWEP:Think()
     if self:GetSeqIdling() then
         self:SetSeqIdling(false)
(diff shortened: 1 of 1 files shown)


Shigbeard:
IIRC, this was done because sometimes the material would apply to other 
sweps.

Kefta:
@Shigbeard That makes sense when it affected the viewmodel in Deploy, but now, that's all handles in clientside VM hooks

Bo98:
OnRemove has had different needs over the years so this is indeed a leftover from that. Before the stick_base refactor it had timers which were removed in OnRemove before I replaced all that in place of proper prediction. And as mentioned the VM tweaks were previously done which also had to be reset.

It's actually possible that you can now remove the `force` parameter entirely from ResetStick. Perhaps even line 93 is not even needed anymore too.

Kefta:
@Bo98 You are correct; thank you for catching that

Bo98:
We don't use the owner the holster so it doesn't really matter.

Kefta:
Now that I'm thinking about it, there's no need to really even reset the world model material -- shouldn't it be set in initialize and left that way? Arrest batons should always be red in the world.

Bo98:
Possibly, yes. Again it could be because it directly replaced some of the view model code. If you try test all scenarios and it works, go ahead.

The only issue I remember about the material is that it gets applied to the entire player if RenderOverride is used on the player, but that still happens and probably will still only happen when deployed regardless of whether it is reset on holster or not.

PR #2618, opened 2017-01-11
Fixes Armor Bar Extending Infinitely

BLEEPBLOOPforLife:
Clamps armor value, so the armor bar does not expand infinitely outwards.

### DIFF
--- a/gamemode/modules/hud/cl_hud.lua
+++ b/gamemode/modules/hud/cl_hud.lua
@@ -95,7 +95,7 @@ local function DrawHealth()
     draw.DrawNonParsedText(math.Max(0, math.Round(myHealth)), "DarkRPHUD2", RelativeX + 4 + (HUDWidth - 8) / 2, RelativeY - 32, ConVars.HealthText, 1)
 
     -- Armor
-    local armor = localplayer:Armor()
+    local armor = math.Clamp(localplayer:Armor(), 0, 100)
     if armor ~= 0 then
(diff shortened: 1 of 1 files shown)


Kefta:
This code should be changed to use MaxArmor instead of 100

BLEEPBLOOPforLife:
Umm... There is no such thing as MaxArmor. I searched the Garry's Mod wiki, old wiki, and the DarkRP wiki.

Shigbeard:
can confirm http://img.ndlee.com/shigbeard/s246y3j.png

roboderpy:
@Kefta fail
2017-01-12_07-59_roboderpy dbot 2

Kefta:
I'm dumb. I thought SetMaxHealth had an armor analogue.

BLEEPBLOOPforLife:
@roboderpy Umm... Even if there were a function for MaxArmor, you would use a colon not a dot. :)

Shigbeard:
@BLEEPBLOOPforLife If you wanted to see if there was a function defined there, you would use a dot and not add brackets on the end. It would return function 0xmemoryaddress.

Anyway, armor uses HL2's old internal maximum health storage of 256 (technically 255, don't know the proper name for it), but it's more logical to assume the max is 100 and anything above 100 is also 100.

BLEEPBLOOPforLife:
@Shigbeard Wow, I learned something new lol. I'm surprised I've actually never heard of that before.

PR #2620, opened 2017-01-19
Customizable arrest stick range.

Icarus5020:
Allows changing the range so a developer can change how far away it can arrest from.

Useful if you want certain jobs to have a higher arrest range than others.

### DIFF
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -10,6 +10,7 @@ DEFINE_BASECLASS("stick_base")
 
 SWEP.Instructions = "Left click to arrest\nRight click to switch batons"
 SWEP.IsDarkRPArrestStick = true
+SWEP.ArrestRange = 8100
 
 SWEP.Spawnable = true
 SWEP.Category = "DarkRP (Utility)"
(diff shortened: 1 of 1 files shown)


FPtje:
The squared distance is an implementation detail and shouldn't show up in that variable. Square it (`self.ArrestRange * self.ArrestRange`)

MrBrax:
that won't really work though, that variable becomes shared on the entity base itself, changing it will change it on all instances of it

Kefta:
@MrBrax What are you talking about? This is config variable that should affect all instances.

MrBrax:
@Kefta you have to add it in Initialize() or similar, otherwise if you change it, it will be changed on all the active/upcoming instances of it
unless that was the point

Kefta:
@MrBrax Config variables like this aren't meant to be changed in real time -- the default implementation doesn't do that.

FPtje:
@Kefta 

> aren't meant to be changed in real time

Why not make this possible? What if someone wants to make some drug to give CPs a larger range temporarily?

MrBrax:
@Kefta 
>Useful if you want certain jobs to have a higher arrest range than others.

Kefta:
@FPtje This should be done in a functional context in my humble opinion; maybe even adding a conditional hook. If you really want to go the variable route, you would have to do:

Icarus5020:
I did as @Kefta suggested.

Kefta:
@Icarus5020 stick_base has an Initialize method as well. You need to call the BaseClass.

FPtje:
Yeah, stick range should be more generic than arrest range.

PR #2621, opened 2017-01-19
Added the viewmodel to PreDrawViewModel Deploy call

Kefta:
Does nothing otherwise. This would have been noticed if there wasn't an IsValid check in PreDrawViewModel since the viewmodel will always be valid when called by the engine, but it masked the problem with the Deploy call. Too many validity checks can be just as bad as having too little!

### DIFF
--- a/entities/weapons/stick_base/shared.lua
+++ b/entities/weapons/stick_base/shared.lua
@@ -71,7 +71,7 @@ function SWEP:Deploy()
 
     local vm = self:GetOwner():GetViewModel()
     if not IsValid(vm) then return true end
-    self:PreDrawViewModel()
+    self:PreDrawViewModel(vm)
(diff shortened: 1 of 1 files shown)


FPtje:
Masked what problem with the deploy call?

Kefta:
Sorry, that was poorly worded. The IsValid check in PreDrawViewModel made it so the problem with calling it with no argument was never noticed.

Bo98:
Well it's obvious the code in SWEP:PreDrawViewModel has no longer been running on deploy for a while. Is this line even needed anymore then?

FPtje:
hl2_2017-01-21_20-04-49

Well it does *something*

FPtje:
Ah, now I see how it works. The function is called by the engine. No need to have it called in Init.

PR #2622, opened 2017-01-24
Fix for rare "compare number with string" error

UAVXP:


FPtje:
These errors are to be fixed, not avoided.

UAVXP:
@FPtje But it's just a regular check for this kind of functions

FPtje:
No, it isn't. It's actually harmful. This function is defined for numbers. This function only makes **sense** for numbers. With that "check", you allow it to do stuff for *other* things than numbers. Your check allow it to do something with strings. The function shouldn't work for strings, because only a tiny subset of strings (namely, ones that contain numbers) would work. 

Surely if some ass was stupid enough to run the function with a string, they'd be stupid enough to run the function with a string that doesn't contain a number. The error would change from comparing a string to a number to `nil` with a number. You would make the error more difficult to understand, just because you think this kind of check is "regular".

UAVXP:
Okay, I get your point of view. Thanks

FPtje:
I shouldn't account for bad calls to my functions. The fact that they error immediately upon bad input is a **good** thing. Functions should not allow input that they don't expect. You're very likely to get unexpected results and weird bugs if your functions are written to try to deal with things they aren't supposed to deal with.

Errors aren't *bad* things. They tell you exactly what's wrong with your script. The more a script complains when you abuse it, the better. After all, errors are *way* easier to solve than weird behaviours.

PR #2632, opened 2017-02-09
strip ammo along with weapons on arrest

fruitwasp:


NemesusRP:
I changed this
And i cant Arrest Player

Bo98:
Works fine for me. Looks like you broke it with the way you have decided to update DarkRP.

PR #2635, opened 2017-02-19
Update link to mysql lib

OctoEnigma:
Because previous thread is closed.

### DIFF
--- a/gamemode/config/_MySQL.lua
+++ b/gamemode/config/_MySQL.lua
@@ -22,7 +22,7 @@ RP_MySQLConfig.MultiStatements = false -- Only available in tmysql4: allow multi
 MANUAL!
 HOW TO USE MySQL FOR DARKRP!
 Download tmysql4 and read the guide here:
-https://facepunch.com/showthread.php?t=1442438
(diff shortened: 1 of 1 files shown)


Bo98:
You should update the text above it to refer to MySQLOO rather than tmysql4 if you're going to do this.

FPtje:
Is that version confirmed to work?

FPtje:
It's not compatible:

> Some important changes between mysqloo9 and mysqloo8 that might cause incompatibilities:
Bigint fields now return as numbers. This might be problematic if you want to use a bigint as the primary key and select it from the server (such as steamid64).
In these cases make sure to cast the bigint to a string in your SQL query like this: SELECT (CAST steamid64 as 'CHAR') as steamid64 FROM ...
Bit fields now return as numbers (0 or 1)

Some work will have to go into making it compatible. I'm not merging this before that's done.

Bo98:
I don't think DarkRP actually retrieves any of the BIGINT fields, but rather just uses them in WHERE clauses. I don't think BIT columns are used either. That's probably why there haven't been any complaints about shit breaking, but testing everything would still be a good idea.

PR #2638, opened 2017-02-25
Remove hook check at admin's SetTeam feature

UAVXP:
Is "playerCanChangeTeam" really needed for this function?


FPtje:
Yes, because otherwise admins can set themselves teams that they can't have.

PR #2652, opened 2017-03-17
Add third argument "order" to DarkRP.addF4MenuTab

### LINKED ISSUE #2651: Add third argument "order" to DarkRP.addF4MenuTab
[//]: # (GO TO THE FORUMS INSTEAD: http://forum.darkrp.com/ )

### Description of the bug

When creating a new F4 tab you can grab the index of the returned panel and switch it's order with another panel, but sometimes you don't want that and you instead just want to place the newly created panel into a certain spot.

In modules/f4menu/cl_init.lua add third argument "order" to DarkRP.addF4MenuTab function.

In modules/f4menu/cl_frame.lua add 7th argument "order" to PANEL:AddSheet function

in the same function add a check if the "order" argument is supplied, if so do
   
               table.insert(self.Items, order, sheet)

In the call to self.tabScroller:AddPanel( sheet.Tab ), have a check for order, if so use this code:
		
                table.insert( self.tabScroller.Panels, order, sheet.Tab )
		sheet.Tab:SetParent( self.tabScroller.pnlCanvas )
		self.tabScroller:InvalidateLayout( true )

 lines 169 through 172 with

               function PANEL:createTab(name, panel, order)
                     local sheet, index = self:AddSheet(name, panel, nil, nil, nil, nil, order)
                     return index, sheet
               end

### Lua errors

I'm sure a lot of developers would love to see this implemented.

DerekSM:
Fixes #2651

### DIFF
--- a/gamemode/modules/f4menu/cl_frame.lua
+++ b/gamemode/modules/f4menu/cl_frame.lua
@@ -68,7 +68,7 @@ function PANEL:SetupCloseButton(func)
     self.CloseButton:SetSize(32, 32)
 end
 
-function PANEL:AddSheet(label, panel, material, NoStretchX, NoStretchY, Tooltip)
+function PANEL:AddSheet(label, panel, material, NoStretchX, NoStretchY, Tooltip, order)
     if not IsValid(panel) then return end
 
     local sheet = {}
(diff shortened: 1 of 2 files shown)


FPtje:
Please use glualint next time. There's inconsistent usage of tabs and spaces

PR #2693, opened 2017-07-13
Take out door classes from IFs

Gyaviste:


FPtje:
why

Gyaviste:
Because it looks better

FPtje:
It doesn't to me. It gives a name to something that is used only once and places it outside the context of where it is used.

Kefta:
You would want to use a hashtable for this anyway

PR #2701, opened 2017-09-10
Changed ents.Create max entity detection to use ents.GetEdictCount

Kefta:
This is how it's done internally, so this will be the most accurate.


FPtje:
The function doesn't exist on the wiki. Was it added in an update? Will it be?

Kefta:
This function exists in the main branch. I have created the wiki page: http://wiki.garrysmod.com/page/ents/GetEdictCount

FPtje:
Then I'll merge when it's live. Thanks!

Kefta:
It is currently live -- it was an undocumented change in the July 2017 update

FPtje:
Super!

PR #2707, opened 2017-10-27
Remove EmitSound override

Kefta:
Cannot replicate the crash anymore

### DIFF
--- a/gamemode/modules/workarounds/sh_workarounds.lua
+++ b/gamemode/modules/workarounds/sh_workarounds.lua
@@ -134,7 +134,7 @@ function ents.Create(name, ...)
 end
 
 if game.SinglePlayer() or GetConVar("sv_lan"):GetBool() then
-	local plyMeta = FindMetaTable("Player")
+    local plyMeta = FindMetaTable("Player")
     local sid64 = plyMeta.SteamID64
 
     function plyMeta:SteamID64(...)
--- a/gamemode/modules/workarounds/sh_workarounds.lua
+++ b/gamemode/modules/workarounds/sh_workarounds.lua
@@ -1,17 +1,4 @@
 -- Shared part
---[[---------------------------------------------------------------------------
-Sound crash glitch
----------------------------------------------------------------------------]]
-
-local entity = FindMetaTable("Entity")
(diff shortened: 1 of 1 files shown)


FPtje:
You mean crash bugs are fixed sometimes?

Kefta:
I never even knew this one existed

FPtje:
I don't remember whether this was a specific sound or just anything with `??` in it

Bo98:
Commit from SVN didn't reveal much: https://github.com/jesserocks3357/darkrp/commit/67851fcd5213f4979a3af2fec50963cd2249b354

But this commit reveals a little more (contains a separate `??` sound check for tools): https://github.com/jesserocks3357/darkrp/commit/a6338056f26060ce99bbf4a19091735d2b7ca6e0

Still don't know the actual crash/whatever but maybe the mention of forline rings a bell.

FPtje:
Forline was a prop kill group iirc. They knew of a bunch of exploits like that and used them to crash people.

Bo98:
If this is considered for removal, it's worth noting that FPP also still contains the check (just slightly different in that it's a single `?`): https://github.com/FPtje/DarkRP/blob/master/gamemode/modules/fpp/pp/server/core.lua#L370-L377

Bo98:
As far as I can gather, the only default GMod tool gun that has "soundname" is the thruster, but the exploit with that was fixed in 2014: https://github.com/Facepunch/garrysmod/commit/284d1569b3835cb169c11c9e8ac46b836bfd2156

I didn't find any with just "sound" (maybe older tools?).

For the broader picture of `EmitSound`, it sounds like issue might still exist, at least in some form if it's not to do with question marks anymore. Though any issue with `EmitSound` looks to also affect `CreateSound` which this workaround does not cover. Does it crash for control characters? Someone try `ent:EmitSound("\n")` unless I get to it first.

Kefta:
Crashes due to bad characters are caught properly

FPtje:
After some questioning around I found the exploit: set the sound of a (wire) turret to `??` and shit will crash. 

I can't reproduce this anymore, though, even with all protections off.

PR #2710, opened 2017-11-14
attempt to optimise darkrp

mcNuggets1:
I found some time to optimise some things in the darkrp core gamemode.
I didnt try optimising the big forfeits, because it would simply need ages and then, it would be declined, because some spaces were missing, so sorry for now :/.

Things I have optimised or fixed:
* Optimised PlayerCanHearPlayers hook in DarkRP.
* Added IsValid checks after ents.Create functions.
* Replaced Distance with DistToSqr checks.
* Optimised sv_chatcommands a lot.
* Optimised a think hook in AFK module. (now a timer)
* Replaced #player.GetAll with player.GetCount.
* Removed unneeded / useless files and functions.
* Removed double font generation. (It just errors and is stupid)
* Some trace entity results are now cached.
* Use ipairs instead of pairs in nearly every numerical loop now.
* Removed lua refresh workaround in hungermod. (It doesnt work for money, why should it work for hunger?)
* Add ParentToHUD Calls added to keysmenu and hitmenu.
* Replace every Nick() call with Name() call to keep it consistent. (sometimes nick is used and sometimes name, but to keep it consistent we are just using Name.)
* Localised a lot of stuff.
* Fixed a few code anomalies.

Things I saw, but I didnt optimise or fix:
* Messy weapon code obviously taken from old TTT code. (Someone might add the updated TTT changes, haha!)
* Missing translations like in sv_chatcommands line 118.
* ents.Create override. (Isn't that bad actually)
* ents.GetAll-loops for little reason. (Doorsystem, I would replace it with doors cached upon OnEntityCreated hook)
* expensive and ineffective tables. (Shipments, CustomEntities, RPExtraTeams)
* workarounds which are to this date useless (sh_workarounds (SKid check etc.)
* the networking sytem (some things there are really weird set up)
* the stupid entity creation of the chat indicator (I might do this in a different pull request)

btw.
Overriding fonts is causing errors on the dev branch, so I removed that.
It's useless however.

Major optimisations:
* Base
* Chatsystem
* Doorsystem
* Loops

### DIFF
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -53,7 +53,6 @@ function SWEP:Deploy()
     return BaseClass.Deploy(self)
 end
 
-local distance = SWEP.stickRange * SWEP.stickRange
 function SWEP:PrimaryAttack()
     BaseClass.PrimaryAttack(self)
 
@@ -71,7 +70,8 @@ function SWEP:PrimaryAttack()
 
     local ent = self:GetOwner():getEyeSightHitEntity(nil, nil, function(p) return p ~= self:GetOwner() and p:IsPlayer() and p:Alive() and p:IsSolid() end)
 
-    if not IsValid(ent) or (self:GetOwner():EyePos():DistToSqr(ent:GetPos()) > distance) or not ent:IsPlayer() then
+    local stickRange = self.stickRange * self.stickRange
+    if not IsValid(ent) or (self:GetOwner():EyePos():DistToSqr(ent:GetPos()) > stickRange * stickRange) or not ent:IsPlayer() then
         return
     end
 
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -53,6 +53,7 @@ function SWEP:Deploy()
     return BaseClass.Deploy(self)
 end
 
+local distance = SWEP.stickRange * SWEP.stickRange
 function SWEP:PrimaryAttack()
     BaseClass.PrimaryAttack(self)
 
@@ -62,14 +63,15 @@ function SWEP:PrimaryAttack()
     local trace = util.QuickTrace(self:GetOwner():EyePos(), self:GetOwner():GetAimVector() * 90, {self:GetOwner()})
     self:GetOwner():LagCompensation(false)
 
-    if IsValid(trace.Entity) and trace.Entity.onArrestStickUsed then
-        trace.Entity:onArrestStickUsed(self:GetOwner())
+    local ent = trace.Entity
+    if IsValid(ent) and ent.onArrestStickUsed then
+        ent:onArrestStickUsed(self:GetOwner())
         return
     end
 
     local ent = self:GetOwner():getEyeSightHitEntity(nil, nil, function(p) return p ~= self:GetOwner() and p:IsPlayer() and p:Alive() and p:IsSolid() end)
 
-    if not IsValid(ent) or (self:GetOwner():EyePos():DistToSqr(ent:GetPos()) > self.stickRange * self.stickRange) or not ent:IsPlayer() then
+    if not IsValid(ent) or (self:GetOwner():EyePos():DistToSqr(ent:GetPos()) > distance) or not ent:IsPlayer() then
         return
     end
 
--- a/entities/weapons/door_ram/shared.lua
+++ b/entities/weapons/door_ram/shared.lua
@@ -87,7 +87,7 @@ local function ramDoor(ply, trace, ent)
     -- Be able to open the door if any member of the door group is warranted
     local keysDoorGroup = ent:getKeysDoorGroup()
     if GAMEMODE.Config.doorwarrants and keysDoorGroup then
-		local teamDoors = RPExtraTeamDoors[keysDoorGroup]
+        local teamDoors = RPExtraTeamDoors[keysDoorGroup]
         if teamDoors then
 			allowed = false
 			for _, v in ipairs(player.GetAll()) do
--- a/entities/weapons/door_ram/shared.lua
+++ b/entities/weapons/door_ram/shared.lua
@@ -73,7 +73,7 @@ local function ramDoor(ply, trace, ent)
     if GAMEMODE.Config.doorwarrants and ent:isKeysOwned() and not ent:isKeysOwnedBy(ply) then
         -- if anyone who owns this door has a warrant for their arrest
         -- allow the police to smash the door in
-        for k, v in pairs(player.GetAll()) do
+        for _, v in ipairs(player.GetAll()) do
             if ent:isKeysOwnedBy(v) and canRam(v) then
                 allowed = true
                 break
@@ -85,14 +85,18 @@ local function ramDoor(ply, trace, ent)
     end
 
     -- Be able to open the door if any member of the door group is warranted
-    if GAMEMODE.Config.doorwarrants and ent:getKeysDoorGroup() and RPExtraTeamDoors[ent:getKeysDoorGroup()] then
-        allowed = false
-        for k,v in pairs(player.GetAll()) do
-            if table.HasValue(RPExtraTeamDoors[ent:getKeysDoorGroup()], v:Team()) and canRam(v) then
-                allowed = true
-                break
-            end
-        end
+    local keysDoorGroup = ent:getKeysDoorGroup()
+    if GAMEMODE.Config.doorwarrants and keysDoorGroup then
+		local teamDoors = RPExtraTeamDoors[keysDoorGroup]
+        if teamDoors then
+			allowed = false
+			for _, v in ipairs(player.GetAll()) do
+				if table.HasValue(teamDoors, v:Team()) and canRam(v) then
+					allowed = true
+					break
+				end
+			end
+		end
     end
 
     if CLIENT then return allowed end
@@ -303,7 +307,7 @@ end
 
 hook.Add("SetupMove", "DarkRP_DoorRamJump", function(ply, mv)
     local wep = ply:GetActiveWeapon()
-    if not IsValid(wep) or not wep.GetIronsights or not wep:GetIronsights() or wep:GetClass() ~= "door_ram" then return end
+    if not wep:IsValid() or not wep.GetIronsights or not wep:GetIronsights() or wep:GetClass() ~= "door_ram" then return end
 
     mv:SetButtons(bit.band(mv:GetButtons(), bit.bnot(IN_JUMP)))
 end)
--- a/entities/weapons/door_ram/shared.lua
+++ b/entities/weapons/door_ram/shared.lua
@@ -89,14 +89,14 @@ local function ramDoor(ply, trace, ent)
     if GAMEMODE.Config.doorwarrants and keysDoorGroup then
         local teamDoors = RPExtraTeamDoors[keysDoorGroup]
         if teamDoors then
-			allowed = false
-			for _, v in ipairs(player.GetAll()) do
-				if table.HasValue(teamDoors, v:Team()) and canRam(v) then
-					allowed = true
-					break
-				end
-			end
-		end
+            allowed = false
+            for _, v in ipairs(player.GetAll()) do
+                if table.HasValue(teamDoors, v:Team()) and canRam(v) then
+                    allowed = true
+                    break
+                end
+            end
+        end
     end
 
     if CLIENT then return allowed end
--- a/gamemode/modules/afk/sv_afk.lua
+++ b/gamemode/modules/afk/sv_afk.lua
@@ -77,11 +77,8 @@ local function AFKTimer(ply, key)
 end
 hook.Add("KeyPress", "DarkRPKeyReleasedCheck", AFKTimer)
 
-local next_think = 0
 local function KillAFKTimer()
-    if next_think > CurTime() then return end
-    next_think = CurTime() + 1
-    for id, ply in ipairs(player.GetAll()) do
+    for _, ply in ipairs(player.GetAll()) do
         if ply.AFKDemote and CurTime() > ply.AFKDemote and not ply:getDarkRPVar("AFK") then
             SetAFK(ply)
             AFKDemote(ply)
@@ -89,7 +86,9 @@ local function KillAFKTimer()
         end
     end
 end
-hook.Add("Think", "DarkRPKeyPressedCheck", KillAFKTimer)
+timer.Create("DarkRPKeyPressedCheck",  1, 0, function()
+    KillAFKTimer()
+end)
 
 local function BlockAFKTeamChange(ply, t, force)
     if ply:getDarkRPVar("AFK") and (not force or t ~= GAMEMODE.DefaultTeam) then
--- a/gamemode/modules/afk/sv_afk.lua
+++ b/gamemode/modules/afk/sv_afk.lua
@@ -77,8 +77,11 @@ local function AFKTimer(ply, key)
 end
 hook.Add("KeyPress", "DarkRPKeyReleasedCheck", AFKTimer)
 
+local next_think = 0
 local function KillAFKTimer()
-    for id, ply in pairs(player.GetAll()) do
+    if next_think > CurTime() then return end
+    next_think = CurTime() + 1
+    for id, ply in ipairs(player.GetAll()) do
         if ply.AFKDemote and CurTime() > ply.AFKDemote and not ply:getDarkRPVar("AFK") then
             SetAFK(ply)
             AFKDemote(ply)
--- a/gamemode/modules/animations/sh_animations.lua
+++ b/gamemode/modules/animations/sh_animations.lua
@@ -195,7 +195,7 @@ local function AnimationMenu()
     end
 
     local i = 0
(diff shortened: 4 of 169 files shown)


mcNuggets1:
I tested everything.

gamemode/modules/afk/sv_afk.lua
```diff
@@ -77,8 +77,11 @@ local function AFKTimer(ply, key)
 end
 hook.Add("KeyPress", "DarkRPKeyReleasedCheck", AFKTimer)
 
+local next_think = 0
 local function KillAFKTimer()
-    for id, ply in pairs(player.GetAll()) do
+    if next_think > CurTime() then return end
+    next_think = CurTime() + 1
```
Bo98: Maybe a preference thing, but this would probably be cleaner as a `timer.Create`.

gamemode/modules/animations/sh_animations.lua
```diff
@@ -185,6 +185,7 @@ local function AnimationMenu()
     AnimFrame.btnMinim:SetVisible(false)
     AnimFrame:SetVisible(true)
     AnimFrame:MakePopup()
+    AnimFrame:ParentToHUD()
```
Bo98: This previously parented to `Panel` above. Are you trying to remove the need for the panel?

gamemode/modules/base/cl_fonts.lua
```diff
@@ -118,5 +118,3 @@ local function loadFonts()
         font = "ChatFont"})
 end
 loadFonts()
--- Load twice because apparently once is not enough
-hook.Add("InitPostEntity", "DarkRP_LoadFonts", loadFonts)
```
Bo98: Not sure how this "errors". Do test the keypad checker. It was the reason this was added.

gamemode/modules/chat/sv_chat.lua
```diff
@@ -96,15 +96,16 @@ local function RP_ActualDoSay(ply, text, callback)
...
+        col2 = Color(255, 200, 200, 255)
         col = col2
     end
 
     if GAMEMODE.Config.alltalk then
-        for k, v in pairs(player.GetAll()) do
-            DarkRP.talkToPerson(v, col, callback .. ply:Name(), col2, text, ply)
+        local name = ply:Name()
+        for k, v in ipairs(player.GetAll()) do
+            DarkRP.talkToPerson(v, col, callback .. name, col2, text, ply)
```
Bo98: meh, if you were going to optimise to this level, you could've at least pulled the concatenation out of the loop (or even the if statement). This isn't the only place. I don't think it's worth the effort going back now though.

gamemode/modules/base/cl_fonts.lua
mcNuggets1: It does error on dev branch

gamemode/modules/chat/sv_chat.lua
mcNuggets1: I just corrected what I saw, sorry. :P

mcNuggets1:
That would be nice indeed

gamemode/modules/animations/sh_animations.lua
mcNuggets1: I had some weird shit happen with ParentToHUD when not called behind MakePopup.

FPtje: > I had some weird shit happen

This is more superstition than a reason. Why don't you run `ParentToHUD` on the panel?

gamemode/modules/animations/sh_animations.lua
```diff
@@ -63,7 +63,7 @@ hook.Add("CalcMainActivity", "darkrp_animations", function(ply, velocity) -- Usi
 
     -- Hobo throwing poop!
     local Weapon = ply:GetActiveWeapon()
-    if RPExtraTeams[ply:Team()] and RPExtraTeams[ply:Team()].hobo and not ply.ThrewPoop and IsValid(Weapon) and Weapon:GetClass() == "weapon_bugbait" and ply:KeyDown(IN_ATTACK) then
+    if RPExtraTeams[ply:Team()] and RPExtraTeams[ply:Team()].hobo and not ply.ThrewPoop and Weapon:IsValid() and Weapon:GetClass() == "weapon_bugbait" and ply:KeyDown(IN_ATTACK) then
```
FPtje: Are you 100% sure `ply:GetActiveWeapon()` won't return `nil` when the player has no weapons? Shit like that is why I use `IsValid`. I hate micro-optimisations like this because they don't improve performance and are a potential for much bigger problems: error messages about shit being nil.

gamemode/modules/base/sh_createitems.lua
```diff
@@ -5,7 +5,7 @@ local plyMeta = FindMetaTable("Player")
 -----------------------------------------------------------
 local function declareTeamCommands(CTeam)
     local k = 0
-    for num,v in pairs(RPExtraTeams) do
+    for num,v in ipairs(RPExtraTeams) do
```
FPtje: `ipairs` can't deal with holes in tables. Again a micro optimisation that you'll have to be **VERY** careful with. Depending on how the disabling of default job is implemented, there might be holes in `RPExtraTeams`. There may also be addons that just set some index of `RPExtraTeams` to `nil`, which would break this code.

mcNuggets1: You already use ipairs in a lot of situations especially for this table.

gamemode/modules/animations/sh_animations.lua
mcNuggets1: I can do a quick test.

mcNuggets1: ] lua_run print(Entity(1):GetActiveWeapon())
> print(Entity(1):GetActiveWeapon())...
[NULL Entity]

Doesn't seem to.

mcNuggets1:
When calling ParentToHUD before MakePopup mouse input is disabled.
I tested it with your animations menu and it works just fine.

mcNuggets1:
player.GetAll and ents.GetAll loops shouldnt have IsValid checks in them, because it is useless.
players can be null (removing player entity), but entities shouldnt be able to.
but removing a player screws up the entire game and got disabled in a gmod update as far as I know

mcNuggets1:
I made this pull request, because a lot of people already have optimised DarkRP, but keep it their own to have a advantage comparing to other servers.
I however like sharing stuff, rather than being a bitch about literally anything I created in lua.

If you need help fo me in any form, I am glad to help you improve this awesome gamemode ;)

mcNuggets1:
I went through everything and now it should fit perfectly.

mcNuggets1:
If this gets merged I would continue the optimisations with the door system.

Kefta:
@mcNuggets1 LocalPlayer's return can be invalid if it is ran too early or the player entity is removed by a module/the engine. Obviously, the latter case is extremely improbable, but I still check LocalPlayer validity everywhere.

mcNuggets1:
I was talking about a invalid player inside player.GetAll :P

mcNuggets1:
If I removed it clientside, I will add it again.

Kefta:
@FPtje Optimising in Lua(JIT) really does come down to small changes like from my experience. It is just so easy to do things the most efficient way the first time in Lua in regards to time complexity. Internal arrays and hashtables are inherent structures that beat any kind of pure-Lua structural implementation, so having to balance and choose data structures happens much less often. Efficient table sorting with custom comparisons is already implemented in table.sort, which eliminates almost all need for Lua table sorting functions. With structures and sorting out of the way, the only main component left of time complexity analysis are individual algorithms, which turn out to be pretty simple to analyse and optimise in an environment where a game engine is doing all of the complex tasks (audio, physics, rendering, etc.).

In full consideration, the majority of optimising work in Garry's Mod comes down to constant time reduction, whether that be through changing a loop iterator, removing unnecessary operations, reusing constant object, and many more small changes. Note that I am not saying that actual optimisation in the traditional sense (time complexity reduction, algorithmic analysis) doesn't happen, but it is just much less common than other programming applications.

mcNuggets1:
Well, there still are a few ugly appearances in the game mode, like the ents.GetAll loops or multiple player.GetAll loops, even if only one is actually needed. (I don't trust myself with that, because I got a different coding style and would mess up, I am actually really tryharding this coding style upon me..)

Bo98:
> When calling ParentToHUD before MakePopup mouse input is disabled.

Is there a reason for `ParentToHUD` in the first place?

FPtje:
Nonsense. Performance will be dominated by bottlenecks, even in Gmod's simplistic Lua, where the engine does a lot of the hard work. Open up any profiling tool and look at the times taken by the top functions to observe that the times are ordered exponentially. 

Even if you're right in that what you call "traditional" optimising is less common, it wouldn't excuse micro optimisations with no noticeable impact with the risk of nil errors. When talking about shit like meta over global `IsValid`, we're talking about **at least** six points behind the comma. We're talking about the kind of shit that needs to be run in loops of tens of thousands to even be able to *measure* the difference. Fix that for thousands of calls per frame and you're still three points behind the comma.

If you want to improve the performance of your server, attack the bottlenecks. You can spend weeks or months turning every `IsValid` to its meta counterpart, but in the end you will find no measurable difference. Attack bottlenecks, however, and you may get rid of lag spikes completely, like I did with FPP a few months ago. I made quite a few operations like 10x faster there.

And in all the fucking years FPP did shit this way I got plenty of complaints that FPP was slow, and plenty of fools coming around to micro optimise the shit out of it. I had known about the specific performance issues for years, but with this immense focus on micro optimisation barely anyone seems to have even noticed the root of the real problem.

mcNuggets1:
@Bo98 The menu hides when you go to the escape menu..

@FPtje The most common bottleneck in DarkRP is the DarkRPPlayerCanHearPlayers hook which is getting some optimisations also, so it is not only micro optimisations.

Kefta:
I think you missed my point - the performance difference between a bottleneck and micro-optimisation will still be incomparable:  that's not what I'm claiming. My point is that in an environment where the fastest methods are basically handed to you through Lua's syntactic sugar, it is much more difficult to create a bottleneck than it would be in a programming language where data structures, sorting, and complex algorithms are all user-implemented or handled. Due to that, the proportion of optimisation possibilities on an unweighted premise will see much less bottlenecks and more constant-time cuts.

In simpler terms, it's much harder to create an inefficient algorithm or implementation in Lua than it would be in another application, which makes optimisation work come down to very close timings and in-depth JIT behaviour. You can see a prime example here: https://facepunch.com/showthread.php?p=52718207#post52718207 (I understand that the hook library would be a bit of an exception since it is one of the most commonly called functions, but the philosophy still applies, though maybe not with the same degree of fine-tuned instruction reduction)

FPtje:
@mcNuggets1 That's good, that function is quadratic, care should be taken there, even though the big optimising there is being done by running the calculations less often than the hook.

@Kefta It's *not* more difficult to create a bottleneck. Bottlenecks will in Lua simply be there until GMod runs without Lua, because then they'll be left in C++. Inefficient ways of storing things in tables, things being calculated more often than necessary. Maybe even the inability to access Source's space partitioning system forcing people to write significantly less efficient algorithms.

Although in GMod I reckon the biggest practical bottlenecks are from inexperienced scripters who don't understand the characteristics of tables, calculate shit unnecessarily often and that kind of shit. DarkRP pure is fast on any server with any amount of players. Add some shitty addons and shit will hit the fan rather quickly. Focusing effort on optimising DarkRP in that context seems like a futile exercise, except of course in the case of macro optimisations or providing APIs that make it more efficient for addons to interact with DarkRP's mechanics.

The link you posted btw lists the top optimisation being the move to a linked list. That shit ain't micro optimisation yo. That shit is macro, big time. Besides, he's put enormous focus into benchmarking and *measuring*. That's precisely the approach I advocate. Benchmark, find bottlenecks, attack those, repeat until the bottleneck is acceptably fast. Of course eventually they did end up doing weird low level shit with the jit, but at least that shit was informed by the benchmarks and the profiler. 

Hooks are the _perfect_ target for optimisation, since they're pretty much the most called functions. Quadratically in amount of players if you take the `PlayerCanHearPlayersVoice` example. Any constant time improvement there automatically scales quadratically, probably more. That's where a constant time optimisation becomes a macro-optimisation.

Constant time optimisations aren't necessarily bad or useless, micro-optimisations are. Especially when they come at the risk of nil errors.

mcNuggets1:
I am going to test all DarkRP functions to check if everything is working properly.

Kefta:
Check my reviews @mcNuggets1

mcNuggets1:
Where exactly can I view them?

Kefta:
Either up on the comment-line when I started them or here: https://github.com/FPtje/DarkRP/pull/2710/files/c347484d5d9ff4075f58c553128506a6f65b3941

Don't know if my visibility is the same as yours.

mcNuggets1:
I am sorry, but I can't find your review.
I searched with CTRL + F for your name, but it didn't find anything. (Yes, the whole page was loaded.)

mcNuggets1:
I now tested around a few minutes and I couldnt find anything.
Going through the code another time to make sure nothing is broken.

Kefta:
Oh it was unpublished, my bad.

mcNuggets1:
Oh shit, haha.

Kefta:
@mcNuggets1 Still not correct, it needs to be **GetPrintName**

mcNuggets1:
Sorry again.
I worked alot with weapons and I was used to automatically write PrintName.

mcNuggets1:
I think, that's it.
I did as good as I can for the moment.

entities/weapons/door_ram/shared.lua
```diff
@@ -85,14 +85,18 @@ local function ramDoor(ply, trace, ent)
...
+        local teamDoors = RPExtraTeamDoors[keysDoorGroup]
+        if teamDoors then
+			allowed = false
+			for _, v in ipairs(player.GetAll()) do
+				if table.HasValue(teamDoors, v:Team()) and canRam(v) then
+					allowed = true
+					break
+				end
+			end
+		end
```
Bo98: Spacing appears to be tabs here (lines 92-99).

entities/weapons/arrest_stick/shared.lua
```diff
@@ -53,6 +53,7 @@ function SWEP:Deploy()
     return BaseClass.Deploy(self)
 end
 
+local distance = SWEP.stickRange * SWEP.stickRange
```
Kefta: This will cache an incorrect value if arrest_stick is inherited and the stickRange is changed. If you want to optimise it, do
```
local range = self.stickRange
if not IsValid(ent) or (self:GetOwner():EyePos():DistToSqr(ent:GetPos()) > range * range) or not ent:IsPlayer() then
```

mcNuggets1: Then the optimisation is less significant, but okay :)

entities/weapons/door_ram/shared.lua
mcNuggets1: Should be corrected.

mcNuggets1:
Gotta sleep for now.
If I missed something performance-ish, just leave a review or pm me.

mcNuggets1:
I guess, this NOW are my final changes.

Kefta:
@mcNuggets1 @FPtje I changed Name->Nick in my #2711 so the conflict can be resolved by removing the changes here

FPtje:
A rebase and squash will probably be most appropriate for this PR. I don't mind solving a couple of conflicts. It'll make 98aa9e1 redundant.

FPtje:
I'm just looking through the code again and found that you remove some clientside functions like this:

```lua
function ENT:Think()
end
```
e.g. in `spawned_money`. The thing is, `spawned_money` inherits `base_gmodentity`, which has [this](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua#L21) `Think` implementation, which keeps checking whether the LocalPlayer is looking at it. For optimisation sake it's probably better to keep it. `ENT:Initialize` and `ENT:Draw` don't have this problem and you were right to see that their empty definitions are redundant.

I see you've added `IsValid` checks in FAdmin's MOTD, ragdoll and jail and in `sv_money.lua`, `sv_sleep.lua` after `ents.Create`. That code will never be reached with [this](https://github.com/FPtje/DarkRP/blob/master/gamemode/modules/workarounds/sh_workarounds.lua#L115) workaround. I maintain the attitude that in cases where `ents.Create` fails, you should **NEVER** go "*oh well, never mind then*", because that hides a huge problem. Even if FAdmin were to live separately again at some point (that is, without that workaround), it would be better to have it error than fail silently.

Also, you've made a bunch of changes to FPP. Those will be gone next time I subtree merge FPP, unless effort is made to backport those changes to FPP's [repository](https://github.com/FPtje/Falcos-Prop-protection). Please don't change FPP code in the DarkRP repository.

FPtje:
What do you mean conflicting files, Github. I squashed and merged it. Go away.

Aka7tsu4:
that was the best update ever made on darkrp !
thanks mcnuggets.

mcNuggets1:
Not really.
Thanks to Falco DarkRP even exists nowaday.
Thank him.

PR #2711, opened 2017-11-14
Fix owner/recipient targeting in cheque

Kefta:
Targeting now uses entity comparisons instead of name, which can be incorrect if two people have the same name or someone has a name equal to DarkRP.getPhrase("unknown")'s return. I also changed Name -> Nick and added the colours as an entity config.


FPtje:
Goodness me, it checked by name!? That's terrible. @Shigbeard how [could](https://github.com/Kefta/DarkRP/commit/6f6b736a4ac436e6fea5e51b195c4deacbbedd69) you? How could *I* have accepted this?

PR #2713, opened 2017-11-18
Only damage entities with stunstick, when they are illegal

mcNuggets1:


Kefta:
I think it should still hurt the hit player

mcNuggets1:
it should still do damage.
the damage for a player is being handled a few lines above.

FPtje:
All entities that the player can damage should be able to get damaged. There will probably be some printers that break with this PR.

mcNuggets1:
Maybe reduce the damage to actually 10 and not entirely break anything as you can just instabreak everything, no matter the health.

PR #2716, opened 2017-11-28
Remove redundant boolean checking.

Randomfish132:


roboderpy:
All == false checks are here for a reason (explicit false check)

Randomfish132:
Yeah, after looking a bit closer I can only see a couple that don't have any side effects, so i'm just going to close this.

strideynet:
Let's not just search and replace an entire codebase...

PR #2722, opened 2018-01-05
Fixed ents.Create override

Kefta:
It was defined shared instead of server-side, and used incorrect unpack logic (unpack will stop to the first nil in the table instead of the real arg count)


Kefta:
Need to make some fixes

PR #2723, opened 2018-01-06
Workarounds fixes

Kefta:
- Add PLAYER.AccountID override
- PLAYER.SteamID64 override now returns a string fallback
- os.date override now passes all args
- Fixed ents.Create override unpack length
- Improved cam.* instance counter
- Merged conflicting InitPostEntity hooks
- Various type verifications to direct errors to the correct location


FPtje:
Parentheses around top level conditions though, tsk

PR #2734, opened 2018-02-19
Fix rendering issues with lab_base

mcNuggets1:


FPtje:
Did you test it?

mcNuggets1:
I don't think, this can do any harm, but yeah, I am gonna test it right now, just to ensure.

mcNuggets1:
After my computer died because I got a shitty HDD, I could test it and it wasn't erroring or doing anything negative.

PR #2735, opened 2018-03-02
new thing to disable

joeyjumper94:
it's now possible to disable just the scoreboard via disabled defaults without removing FAdmin.
check the proposed change to darkrp modification


FPtje:
See https://github.com/FPtje/darkrpmodification/pull/111

PR #2736, opened 2018-03-02
Fix setaccess offline player

chelog:
How the world didn't tell you for so long?
Does that mean nobody (except my friend) uses FAdmin? It's very important feature and this fix is extremely simple. Hope this helps someone.

### DIFF
--- a/gamemode/modules/fadmin/fadmin/access/sv_init.lua
+++ b/gamemode/modules/fadmin/fadmin/access/sv_init.lua
@@ -343,7 +343,7 @@ function FAdmin.Access.SetAccess(ply, cmd, args)
         return false
     end
 
-    if not targets and (string.find(args[1], "STEAM_") and #args == 5 or args[1] == "BOT") then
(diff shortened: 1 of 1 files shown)


FPtje:
Could it have something to do worth older and newer steam ids? What's your steam id? If it's in a different format than, say, mine (STEAM_0:0:8944068) you might indeed need six args instead of five.

FPtje:
Oh wait, the sixth argument is always used in that case for the group name.

FPtje:
And no, no one did tell me.

It's pretty funny, many of the real bugs that get reported nowadays are ones that have gone unnoticed for years and years. In this case, I reckon people might have hit it, but just put quotes around the steam id making it work again.

chelog:
No, it would work for online players only. I tried both ways and did not succeed on setting offline player's rank. It just says "Player not found".

MrBrax:
i used to do it via raw mysql queries, i just thought it wasn't implemented :ok_hand:

Issue #2737, opened 2018-03-15
Adding SWEP to the Pocket serializes invalid self.Weapon


yupi2:
[//]: # (DO NOT REMOVE THIS TEMPLATE.)
[//]: # (GITHUB ISSUES ARE **NOT** FOR HELP)
[//]: # (GO TO THE FORUMS INSTEAD: http://forum.darkrp.com/ )

### Description of the bug
[//]: # (Describe the issue as accurately as possible)
`addPocketItem` on weapons serializes the `self.Weapon` value on SWEPs (which is the same as `self`). When the weapon is removed by `addPocketItem` (`ent:Remove()`) it leaves the serialized `self.Weapon` as an invalid entity / NULL. I was going to work around this by overriding `FindMetaTable("Weapon").__index` to check if the key is `Weapon` and then returning `self` before the `GetTable` call ( around here https://github.com/Facepunch/garrysmod/blob/5f8d8bd6c05508a7ec57153865bd2a2652861d73/garrysmod/lua/includes/extensions/weapon.lua#L26 ) but it just seems better to remove the `Weapon` key in the serialized table if the entity is a weapon.

My proposed fix is to add `if ent:IsWeapon() then serialized.Weapon = nil end` before returning the serialized table right about here https://github.com/FPtje/DarkRP/blob/dc12eebbbdb85fc6c5c9f46d8a34b853645e2cb3/entities/weapons/pocket/sv_init.lua#L149

### How to make the bug happen
[//]: # (Knowing how to make a bug happen allows the developers to figure out)
[//]: # (what the cause of the problem is and whether a certain fix solves it)
Spawn an AK-47 with the toolgun, pocket it, unpocket it, and then attempt to shoot.

### Lua errors
[//]: # (Note that errors on server startup are more important than other ones)
[//]: # (because they can cause other errors. Always look in the startup log of the server!)
An example of trying to shoot the https://github.com/FPtje/DarkRP/tree/master/entities/weapons/weapon_ak472 who's base is `weapon_base` which uses `self.Weapon` in places.
```
[ERROR] gamemodes/base/entities/weapons/weapon_base/shared.lua:206: Tried to use a NULL entity!
  1. Clip1 - [C]:-1
   2. CanPrimaryAttack - gamemodes/base/entities/weapons/weapon_base/shared.lua:206
    3. unknown - gamemodes/darkrp/entities/weapons/weapon_cs_base2/shared.lua:188
```

[...]

PR #2738, opened 2018-03-15
Remove self.Weapon serialzation for SWEPs #2737

yupi2:


FPtje:
Solving it when deserializing is better.

PR #2753, opened 2018-05-06
Added "everyone" option to adminweapons config

Kefta:
Would be better to work those options into CAPI so server owners can select specific groups, but this at least adds an option to remain consistent with adminsents and adminvehicles.


Cocobor:
Thank you very much! Everything works!

PR #2756, opened 2018-05-22
Fixed changeTeam not giving a descriptive message.

Codingale:
It used to simply say "TEAM NAME or TEAM NAME" or even, if you only had one job in the table just "TEAM NAME".

Now it's "You need to be TEAM 1 or TEAM 2 first in order to be able to become WANTED NAME", I also increased the time as 4 seconds might not be enough if the table is big. Unsure how it'd behave with tables with many entires or long job names.


FPtje:
Look it's in there twice now. Github didn't realise I merged it, so I clicked the merge button to get it to shut up. Now it's in there twice. Damnit

PR #2759, opened 2018-05-29
Fix gmod update compat + minor optimisations

mcNuggets1:
The gmod update changed the behaviour of GetObserverTarget to always return NULL when invalid, like it is supposed to.

Therefore weapon dropping can't work.

I fixed it, additionally some minor optimisations.

### DIFF
--- a/entities/entities/fadmin_motd/cl_init.lua
+++ b/entities/entities/fadmin_motd/cl_init.lua
@@ -59,16 +59,17 @@ function ENT:Draw()
     self:DrawModel()
 
     local pos = self:GetPos()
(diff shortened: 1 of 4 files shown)


Kefta:
Thanks, missed this when I sweeped DarkRP for the NULL changes

FPtje:
Surely the de's would find out during development, or at least the moment someone uses it on DarkRP?

Kefta:
Guess no one tested that feature, I missed it myself

Bo98:
Well if chatboxes override Player.IsTyping, they will be able to get the new method to be compatible at least. I don't think there's any such guarantees before, but again I haven't tried.

mcNuggets1:
Looks good to me <3

PR #2762, opened 2018-06-20
Fix advert notification text

Gyaviste:


FPtje:
DarkRP's language system is bad, but that would make it worse. With "advert" we have the excuse that it's the command, "advertise" is an English word that would be impossible to translate in another language.

Gyaviste:
Yes, it IS a command
```
GAMEMODE.Config.chatCommandPrefix .. "advertise"
```
in
```
DarkRP.notify(ply, 1, 4, message or DarkRP.getPhrase("unable", GAMEMODE.Config.chatCommandPrefix .. "advertise", ""))
```

Bo98:
@FPtje This is correcting the untranslated command within the message _from_ /advertise to /advert.

FPtje:
Huh, you're right.

The repo's been deleted though.

PR #2764, opened 2018-06-24
Redesigned chat indicator

Bo98:
Improvements:
* It is now completely clientside - no server processing and no networking.
* It now follows the player smoothly and not lag.
* The rotation speed is now FPS independent.
* Playermodel scaling is now observed.
* Easier to disable.

Caveats:
* Custom chatboxes must support `Player:IsTyping()` to work with the indicator.
   - Previously the chat indicator _may_ have worked, but was not guaranteed due to undefined hook ordering.
   - Because custom chatboxes will have to put in the effort of supporting `Player:IsTyping()` themselves, the indicator _may_ not work out-of-the-box for as many chatboxes.
   - However, upon implementing this function the chatboxes can now be guaranteed that the indicator will work, unlike before.
   - Well-coded chatboxes should already be implementing this method. GMod itself uses the method for player animations.
* Chat indication while ragdolled through most admin mods (not death ragdoll) is still unsupported. Previously it would leave an indicator in the place where you got ragdolled. Now it doesn't display an indicator at all which I believe is a little better.


Bo98:
Changes since my comment https://github.com/FPtje/DarkRP/commit/f18165d4cadbb55f168ed5a887133ac4d564a70a#commitcomment-29226633:
* Now works again while dead.
* Fixed variation in rotation speed while swimming.

PR #2778, opened 2018-07-19
Update Keys

JarosLucky:
Delete double check `!IsValid(ent) or (IsValid(ent)...`

### DIFF
--- a/entities/weapons/keys/shared.lua
+++ b/entities/weapons/keys/shared.lua
@@ -137,7 +137,7 @@ end
 
 function SWEP:Reload()
     local trace = self:GetOwner():GetEyeTrace()
-    if not IsValid(trace.Entity) or (IsValid(trace.Entity) and ((not trace.Entity:isDoor() and not trace.Entity:IsVehicle()) or self.Owner:EyePos():DistToSqr(trace.HitPos) > 40000)) then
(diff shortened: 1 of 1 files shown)


entities/weapons/keys/shared.lua
```diff
@@ -137,7 +137,7 @@ end
 
 function SWEP:Reload()
     local trace = self:GetOwner():GetEyeTrace()
-    if not IsValid(trace.Entity) or (IsValid(trace.Entity) and ((not trace.Entity:isDoor() and not trace.Entity:IsVehicle()) or self.Owner:EyePos():DistToSqr(trace.HitPos) > 40000)) then
+    if not IsValid(trace.Entity) or (((not trace.Entity:isDoor() and not trace.Entity:IsVehicle()) or self.Owner:EyePos():DistToSqr(trace.HitPos) > 40000)) then
```
mcNuggets1: But the check is in bracktes, while the IsValid check is not.
The IsValid checks then runs after all the other checks

Kefta: @mcNuggets1 That's fine, the if-statement is short circuited if the entity is invalid - the second set of statements will only be reached with a valid entity, making the second IsValid useless.

mcNuggets1: me idiot. need to get more rest nowadays, haha.

Issue #2780, opened 2018-07-22
Shutdown FAdmin Jail error


mcNuggets1:
[//]: # (DO NOT REMOVE THIS TEMPLATE.)
[//]: # (GITHUB ISSUES ARE **NOT** FOR HELP)
[//]: # (GO TO THE FORUMS INSTEAD: http://forum.darkrp.com/ )

### Description of the bug
[//]: # (Describe the issue as accurately as possible)
I already reported this error like 4 times, the last time, it says it got fixed, but unfortunately, it's still not fixed.
Still after closing the server, there is a slight chance this error appears:

L 07/22/2018 - 10:24:39: Server successfully shut down
L 07/22/2018 - 10:24:41: Lua Error: 
[ERROR] gamemodes/darkrp/entities/entities/fadmin_jail/init.lua:26: Tried to use a NULL entity!
  1. SetPos - [C]:-1
   2. unknown - gamemodes/darkrp/entities/entities/fadmin_jail/init.lua:26

L 07/22/2018 - 10:24:41: Lua Error: 
[ERROR] gamemodes/darkrp/entities/entities/fadmin_jail/init.lua:26: Tried to use a NULL entity!
  1. SetPos - [C]:-1
   2. unknown - gamemodes/darkrp/entities/entities/fadmin_jail/init.lua:26

L 07/22/2018 - 10:24:42: Log file closed.

Another note:
My server locked for like 1 hour maybe from this error.
It didn't respond, had to restart it.
Was locked for 1 hour due to this.

### How to make the bug happen
[//]: # (Knowing how to make a bug happen allows the developers to figure out)
[//]: # (what the cause of the problem is and whether a certain fix solves it)
Shut down the server
?

### Lua errors
[//]: # (Note that errors on server startup are more important than other ones)
[//]: # (because they can cause other errors. Always look in the startup log of the server!)
[ERROR] gamemodes/darkrp/entities/entities/fadmin_jail/init.lua:26: Tried to use a NULL entity!
  1. SetPos - [C]:-1
   2. unknown - gamemodes/darkrp/entities/entities/fadmin_jail/init.lua:26

### Why the developer of DarkRP is responsible for this issue
[//]: # (By posting on Github, you ask the developers of DarkRP to solve the problem)
[//]: # (It may seem obvious to you, but you have to make clear why they are the right people to look at it)
FAdmin

Kefta:
[...]

PR #2781, opened 2018-07-22
Add new jail validity check

Kefta:
Fixes https://github.com/FPtje/DarkRP/issues/2780. Don't lock up servers if they can't create a new jail.


mcNuggets1:
Shouldn't this be fixed in GMod itself or is it just the way the engine works?

Kefta:
I don't know what it's caused by yet, it doesn't happen on my Windows dedi or listen server

mcNuggets1:
I am on linux, it happens there on some rare occasions.

FPtje:
There are a ton of these in DarkRP already. That's just the reality you have to live with when you code Lua in GMod

PR #2783, opened 2018-07-23
freeze AFK players salary

ReturnEnd:
While it's shown to AFK players that their salary is frozen, it is not actually; only the darkrpvar is set to 0 and the function Player:payDay() isn't using that variable.

### DIFF
--- a/gamemode/modules/afk/sv_afk.lua
+++ b/gamemode/modules/afk/sv_afk.lua
@@ -99,6 +99,7 @@ local function BlockAFKTeamChange(ply, t, force)
 end
 hook.Add("playerCanChangeTeam", "AFKCanChangeTeam", BlockAFKTeamChange)
 
+-- Freeze AFK player's salary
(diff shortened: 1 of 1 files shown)


gamemode/modules/afk/sv_afk.lua
```diff
@@ -99,6 +99,13 @@ local function BlockAFKTeamChange(ply, t, force)
 end
 hook.Add("playerCanChangeTeam", "AFKCanChangeTeam", BlockAFKTeamChange)
 
+-- Freeze player's salary
+hook.Add("playerGetSalary", "AFKGetSalary", function(ply, amount)
+    if ply:getDarkRPVar("AFK") then
+        return true, "", 0 -- false, DarkRP.getPhrase("salary_frozen"), 0
```
FPtje: Could you get rid of this commented code?

ReturnEnd: Sure, I was not sure about how you would like it to be implemented, if you wanted the player to be noticed that he's not receiving his salary upon payDay()

PR #2790, opened 2018-08-11
Dist check in keys in keys code isn't good

ExtReMLapin:
When using the keys on a door (func_door) and if it's origin point of the entity or far from the entity itself, , you won't be able to use the keys on it, fix it by doing a trace.

Example, if a door is 1km wide, you will not be able to use the door anywhere near it but only at the place near the origin point

Video showing the door being far from it's origin point : https://www.youtube.com/watch?v=bp7OlidbMq0&feature=youtu.be


ExtReMLapin:
It's also a micro optimisation since NearestPoint and GetPos are not called anymore

Bo98:
And keys menu

ExtReMLapin:
done

PR #2792, opened 2018-09-06
use player to emit sound

mcNuggets1:
The flashlight sound can mute/remove the sound played on weapons for some reason (presumably because of the one sound object per entity restriction). It's not doing it on the player itself, so we call sounds from the player from now on.

btw. in a lot of places, it already uses the player to emit sounds but in some functions not.

### DIFF
--- a/entities/weapons/lockpick/shared.lua
+++ b/entities/weapons/lockpick/shared.lua
@@ -159,7 +159,7 @@ function SWEP:Think()
     if CurTime() >= self:GetNextSoundTime() then
         self:SetNextSoundTime(CurTime() + 1)
         local snd = {1,3,4}
-        self:EmitSound("weapons/357/357_reload" .. tostring(snd[math.Round(util.SharedRandom("DarkRP_LockpickSnd" .. CurTime(), 1, #snd))]) .. ".wav", 50, 100)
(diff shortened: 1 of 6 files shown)


FPtje:
Surely the flashlight sound isn't emitted from whatever weapon the player is carrying? What if the player carries no weapon, would there be no flashlight sound?

By the way, is the flashlight sound muting weapon sounds a problem at all? Are people being sneaky in some way and hiding weapon sounds with their flashlights? Why would they do that? Is it mainly with the weapon checker?

Kefta:
Regardless of the flashlight issue, sounds should be emitted on the player and not the weapon. Weapon sounds often get muted by random world sounds or just don't play. I think it's due to them being child entities but I haven't looked too much into it.

FPtje:
@Kefta do you mean to say that there is a bug that we need to work around? To me it makes perfect sense to emit weapon sounds from weapons. Especially with players being much larger than the weapons. In an ideal game engine and with an ideal sound system, you'd be able to stand next to a player, close your eyes and be able to tell where their weapon is just by listening to its noises. 

If it's a buggy game issue, though, fuck it. In that case it should just be documented: don't emit sounds from child entities because the game is buggy.

FPtje:
Fuck it

PR #2795, opened 2018-09-18
Changed direct datatable usage to methods

Kefta:
Using self.dt to set datatable variables instead of the provided NetworkVar functions is undocumented behaviour. It also unnecessarily forces replacement entities mimicking the API of stock DarkRP entities to use NetworkVars for networking implementation.

### DIFF
--- a/gamemode/modules/base/sh_createitems.lua
+++ b/gamemode/modules/base/sh_createitems.lua
@@ -297,8 +297,6 @@ local function addEntityCommands(tblEnt)
     local function defaultSpawn(ply, tr, tblE)
         local ent = ents.Create(tblE.ent)
         if not ent:IsValid() then error("Entity '" .. tblE.ent .. "' does not exist or is not valid.") end
-        ent.dt = ent.dt or {}
-        ent.dt.owning_ent = ply
         if ent.Setowning_ent then ent:Setowning_ent(ply) end
         ent:SetPos(tr.HitPos)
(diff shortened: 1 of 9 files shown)


gamemode/modules/base/sh_createitems.lua
```diff
@@ -297,8 +297,6 @@ local function addEntityCommands(tblEnt)
     local function defaultSpawn(ply, tr, tblE)
         local ent = ents.Create(tblE.ent)
         if not ent:IsValid() then error("Entity '" .. tblE.ent .. "' does not exist or is not valid.") end
-        ent.dt = ent.dt or {}
-        ent.dt.owning_ent = ply
```
FPtje: This code is suspicious. Not only am I calling the `Setowning_ent` function, I also set `dt` if it hasn't been already, *and* the dt var itself. I guess I should have documented why I did this shit, now I'm being paranoid about it working around some obscure bug where ent.dt is nil.

Bo98: It was compatibility with the old [DTVar method](https://wiki.garrysmod.com/page/Entity/DTVar), which is now an internal function to NetworkVar.

Even [GMod itself](https://github.com/Facepunch/garrysmod/commit/3e138636eb1b0ad6ed785dedf350020755cff5f1#r16402605) has broken compatibility for it in the past, so I'm not sure how widespread its usage is anymore.

PR #2804, opened 2018-10-13
Added shipment category support for when table syntax isn't used

Kekalainen:
Currently shipment categories can only be set by using the table syntax to create shipments. This change fixes the issue while remaining compatible with any old shipment creation function calls.


FPtje:
I didn't add it because that syntax is deprecated, not having category support is a reason for people to switch. Besides, the glualint website automatically converts jobs to table syntax when you click pretty print (hidden feature!)

I'd say keep it deprecated and incomplete so people use the new syntax.

Issue #2769, opened 2018-07-06
Error log


matiasg90:
[ERROR] gamemodes/darkrp/gamemode/modules/chatindicator/cl_init.lua:2: Tried to use a NULL entity!
  1. IsTyping - [C]:-1
   2. drawIndicator - gamemodes/darkrp/gamemode/modules/chatindicator/cl_init.lua:2
    3. unknown - gamemodes/darkrp/gamemode/modules/chatindicator/cl_init.lua:37

roboderpy:
This error is impossible, but if player got :Remove()'d, this is possible.

Kefta:
This is possible if the ragdoll is still around after the target player left.

PR #2806, opened 2018-10-16
Minor optimisations and code clean up

mcNuggets1:
I tried to optimise DarkRP using FProfiler a little bit.

I removed nearly all unnecessary IsValid checks on frequently called functions and ensured nothing would error.
I also micro optimised a lot of stuff and fixed a few MySQL related lua errors.

I have in mind to optimise FPP as it is currently the most unoptimised addon I privately use.


mcNuggets1:
Don't merge yet.

mcNuggets1:
I'm done with changing stuff.

I tested these changes on my server, my listen server and I didn't encounter any bugs, I went through the whole code 3 times and couldn't see anymore possible mistakes.

Bo98:
How much has all this actually helped?

mcNuggets1:
It's better than before. Atleast for me.
But this isn't a overhaul of DarkRP's optimisation.

FPtje:
The problem with removing IsValid checks is that it'll work just fine when you run it locally, but somewhere, at some point, some weird shit will happen that will make the weirdest shit go invalid.

Most of the reported errors in DarkRP are null ones. As such I don't want any IsValid checks removed. Besides, I don't think removing them will help at all.

mcNuggets1:
I used Notepad++ to track down every usage of the functions I removed the IsValid checks.
There is barely any chance, it errors anywhere, I can promise you that.

I even fixed a few NULL errors related to MySQL.

FPtje:
> Also in functions frequently used they have an impact, even if it is minimal.
Nonsense. This cannot be called an impact:
hl2_2018-10-17_21-48-14

Also, have a search for `null` in the issues. You'll find ones that are [said to be impossible](https://github.com/FPtje/DarkRP/issues/2769) and many other ones that turn out to be really tricky. Seriously, I hate null errors with a passion. Your removal of `IsValid` checks creates a serious risk of null error chasing, despite your Notepad++ checks.

gamemode/modules/hungermod/sv_hungermod.lua
```diff
@@ -20,7 +20,7 @@ local function HMAFKHook(ply, afk)
     if afk then
         ply.preAFKHunger = ply:getDarkRPVar("Energy")
     else
-        ply:setDarkRPVar("Energy", ply.preAFKHunger or 100)
+        ply:setSelfDarkRPVar("Energy", ply.preAFKHunger or 100)
```
FPtje: Why this change? All other places use `setSelfDarkRPVar` for Energy

FPtje:
Also, you're changing stuff in FPP and fspectate. Know that they have their own repositories that the changes should be backported to. I can do it easily enough though.

gamemode/modules/hungermod/sv_hungermod.lua
mcNuggets1: I know, I replaced it with setSelfDarkRPVar

mcNuggets1:
On functions called every frame, the frame impact is still minimal, I never said it would be really significant.
But the fact, you don't get errors when using functions you shouldnt use with null entities is misleading, also 

I know for sure (for 99%), there won't be any null errors occuring.
If some occur, I promise you, I will immediatly fix them via a pull request.

I am developing in GMod for too long now and I get when null errors appear now.
I fixed a few in this pull request already.

gamemode/modules/hungermod/sv_hungermod.lua
FPtje: oh, whoops

Bo98:
I'm bored:

* money_printer: IsValid check seems to be from when there was timer code using the function directly. `timer.Simple(30, ent.CreateMoneyBag, ent)`
* spawned_shipment: the same
* DarkRP.storeMoney, DarkRP.retrieveSalary and DarkRP.payPlayer: Added nearly 10 years ago under the commit "Really minor update :)". Predates the first Google Code issue. It could be argued that API functions like these shouldn't silence errors.
* Player:restorePlayerData: added in 2010 when the purpose of it was to load the RP name. https://code.google.com/archive/p/darkrp/issues/345. GMod 12 beta? I'd argue the cause of the errors are elsewhere.
* Player:setDarkRPVar and Player:teamUnBan: https://code.google.com/archive/p/darkrp/issues/343. Same reporter as above. Seemed to have errors basically everywhere.
* Player:payDay: was because it had similar timer code to what I mentioned earlier.
* Player:isCP: 22499eb

mcNuggets1:
* moneyprinter: The money printer has a built in IsValid check nowadays
* shipment: Same goes for shipment
* money funcs: There was no way these could error in any way.
* setDarkRPVar: I also couldn't find any line a NULL player was used together with this function.
* payDay: only gets called once with a already existing IsValid check.
* isCP: Also couldn't find anything related to isCP causing a NULL error.

Silently fail to run a function should never be the case in any way.
It should always report, when there is something wrong going on.

FPtje:
Thanks for the contribution

PR #2809, opened 2018-10-25
Added player argument to addLaw and removeLaw hook calls

WilliamVenner:


FPtje:
Looks good! Could you update the hook stub?

FPtje:
Whoa, that's quick!

PR #2813, opened 2018-10-31
Tax system: hook taxableEntities to decide which entities are taxable

FPtje:
Alternative implementation for #2808

The behaviour @mcNuggets1 desires would be implemented as follows:

```lua
hook.Add("taxableEntities", "no vehicles", function(ply, taxables)
    for k, _ in pairs(taxables) do
        if k:IsVehicle() then
            taxables[k] = nil
        end
    end

    return taxables
end)
```

Admittedly the fixpoint algorithm is way overkill. It allows one hook listener to modify the output of other hook listeners. So if listener A removes vehicles from the list, and listener B adds all entities spawned by the player to the list (for some reason), the end result will be all entities spawned by the player except for vehicles. Even if listener B runs after listener A. That's pretty cool, huh?

@mcNuggets1 @Bo98 Other than having overengineered this, what do you think?

### DIFF
--- a/gamemode/modules/doorsystem/sv_doors.lua
+++ b/gamemode/modules/doorsystem/sv_doors.lua
@@ -76,7 +76,7 @@ function meta:keysOwn(ply)
 
     ply.OwnedNumz = ply.OwnedNumz or 0
     if ply.OwnedNumz == 0 and GAMEMODE.Config.propertytax then
-        timer.Create(ply:SteamID64() .. "propertytax", 270, 0, function() ply.doPropertyTax(ply) end)
+        timer.Create(ply:SteamID64() .. "propertytax", 270, 0, function() ply:doPropertyTax() end)
     end
 
     ply.OwnedNumz = ply.OwnedNumz + 1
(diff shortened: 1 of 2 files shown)


mcNuggets1:
tbh I dont see any difference to my system instead of this being a whole table.

FPtje:
With my hook people can add entities. Though I'm not sure that'll work with the selling part.

Also, for adding things the hooks need to know whether they're in the first iteration. I just figured that.

mcNuggets1:
If you are going to add new stuff being taxed, people should be allowed to buy more than just doors and vehicles.

Otherwise taxes wouldn't make sense and how is a prop unowned for example?

Bo98:
I like the flexibility. Could make things like "prop tax" easier.

> It allows one hook listener to modify the output of other hook listeners.

I like that. The design of `hook.Call` sucks a lot in some cases.

"Battles" between hooks is an interesting problem - it's good that you considered that with the fix point algorithm.

Indeed you will need to add an indication whether it is the first iteration, and clearly document about the iterations - it may catch people off guard with it being called potentially up to 5 times.

The return value in your example isn't necessary but I suppose the ability to return a value would allow you to completely replace the list for whatever reason that may be.

> In any way taxes wouldn't make sense on non owned entities, because all you can loose is the ownage of these specific entities.

I suppose then there should be some way to handle the "unown" event. We're steering away from the doorsystem a bit here, but it would make sense to be able to do `ent:Remove()` or something the developer chooses on entities added to the tax system.

Are we moving from hook land to entity registration API land? Perhaps not if we think it should be possible to override the behaviour of default ownable entities - would that even make sense on a per entity basis?

Some food for thought.

mcNuggets1:
I personally think, this system is very developer unfriendly.

Also the use case is very limited, being able to say which entities are getting taxed is a good thing, but adding new entities doesn't make sense at all.
There are too many cases where this can become a big problem.

So basically it's the same as my attempt, but with using a table.

Bo98:
One pretty major difference between the two solutions is the tax calculation.

mcNuggets1:
Yeah, right.

With my solution, it uses the standard taxes, but you are still able to override them using the hook which is actually used for taxes calculation, which sounds much better to me.

I dislike the idea having 2 hooks calculating taxes.

FPtje:
Closed in favour of #2808

mcNuggets1:
You did a good job with this.

FPtje:
Thanks, and you did a great job too. Its great being able to have discussions about this sort of thing and then come to a good solution.

PR #2823, opened 2018-12-04
Fixed the failure to demote a player from his weapons license.

FlorianLeChat:
The function now checks an existing variable "win" and no longer a variable "choice" that made the vote always negative.


FPtje:
Another amazing find! The fix you've applied isn't *quite* it. The callback isn't called with two arguments, but just one. I've merged it anyway and quickly fixed that in another commit.

Great work! That bug must have been in there for years!

FlorianLeChat:
Thank you, I'm happy to help a little to improve this gamemode! 👍

FPtje:
Turns out I was wrong, the function *does* get two arguments. I missed the fact that it's a meta call. Thanks @Bo98

PR #2825, opened 2018-12-08
fix typo

Lcast15:
pretty sure players aren't selling other players


Kefta:
That's illegal in most countries

FPtje:
In DarkRP it's only illegal if the mayor says it is.

PR #2831, opened 2019-01-24
Make the NPC Kill Pay system extensible

Lexicality:
A user asked me to make my [NPC Spawn Platforms](https://steamcommunity.com/sharedfiles/filedetails/?id=107821465) allow them to configure the amount of money players get when they kill a spawned NPC.
I've added a generic variable for this (Lexicality/Lexical-Tools@fac1d107615c39b4cd9ef76f0144e20396f7f583), but currently DarkRP doesn't allow you to configure that value per-npc.
This PR changes that.

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -99,6 +99,10 @@ function GM:canEarnNPCKillPay(ply, npc)
 end
 
 function GM:calculateNPCKillPay(ply, npc)
+    -- A NPC spawned by an addon might be worth more money than the default
+    if npc.KillValue then
+        return npc.KillValue
+    end
     return GAMEMODE.Config.npckillpay
 end
 
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -94,6 +94,14 @@ function GM:canSeeLogMessage(ply, message, colour)
     return true
 end
 
+function GM:canEarnNPCKillPay(ply, npc)
+    return GAMEMODE.Config.npckillpay > 0
+end
+
+function GM:calculateNPCKillPay(ply, npc)
+    return GAMEMODE.Config.npckillpay
+end
+
 --[[---------------------------------------------------------
  Gamemode functions
  ---------------------------------------------------------]]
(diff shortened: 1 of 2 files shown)


gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -94,6 +94,18 @@ function GM:canSeeLogMessage(ply, message, colour)
     return true
 end
 
+function GM:canEarnNPCMoney(ply, npc)
+    return GAMEMODE.Config.npckillpay > 0
```
FPtje: Shouldn't this take KillValue into account?

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -94,6 +94,18 @@ function GM:canSeeLogMessage(ply, message, colour)
...
+end
+
+function GM:calculateNPCMoney(ply, npc)
+    -- A NPC spawned by an addon might be worth more money than the default
+    if npc.KillValue then
+        return npc.KillValue
```
FPtje: I'm not too sure about this var being in DarkRP by default. It sounds generic enough to mean something else entirely in another addon. 
It's also obscure enough for people not to know it exists until they read the code. With the hook support you've added this functionality can go into your addon. 

What do you think?

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -263,9 +275,10 @@ function GM:OnNPCKilled(victim, ent, weapon)
     end
 
     -- If we know by now who killed the NPC, pay them.
-    if IsValid(ent) and GAMEMODE.Config.npckillpay > 0 then
-        ent:addMoney(GAMEMODE.Config.npckillpay)
-        DarkRP.notify(ent, 0, 4, DarkRP.getPhrase("npc_killpay", DarkRP.formatMoney(GAMEMODE.Config.npckillpay)))
+    if IsValid(ent) and hook.Call("canEarnNPCMoney", GAMEMODE, ent, victim) then
+        local amount = hook.Call("calculateNPCMoney", GAMEMODE, ent, victim)
```
FPtje: Could you make hook stubs for these two hooks? That way the wiki can be updated.

gamemode/modules/base/sv_gamemode_functions.lua
Lexicality: I figured that someone wanting all NPCs to be worth $0 except certain ones would be so obscure that anyone that did it could just return true in a hook

gamemode/modules/base/sv_gamemode_functions.lua
Lexicality: My bad

gamemode/modules/base/sv_gamemode_functions.lua
Lexicality: (I'm happy to rename the variable or drop the commit if you prefer though)

gamemode/modules/base/sv_gamemode_functions.lua
Bo98: From my point of view, I think this would be more descriptive if named calculateNPCKillAward or calculateNPCKillPay or similar. To me, "NPC money" is quite vague - it could be how much money the NPC is holding, how much one costs to spawn, ...

Maybe it's just me though.

PR #2835, opened 2019-02-12
Explosion effect changes

mcNuggets1:
* Added watersurfaceexplosion, when under water.
* Sprayed a decal on explosion.
* Made the screen shake a little.
* Also reduced explosion damage, because it previously was extremely high.

### DIFF
--- a/entities/entities/lab_base/init.lua
+++ b/entities/entities/lab_base/init.lua
@@ -34,12 +34,15 @@ end
 function ENT:Destruct()
     local vPoint = self:GetPos()
 
-    util.BlastDamage(self, self, vPoint, 200, 200)
+    util.BlastDamage(self, self, vPoint, 100, 100)
+    util.ScreenShake(vPoint, 512, 255, 1.5, 200)
+
     local effectdata = EffectData()
     effectdata:SetStart(vPoint)
     effectdata:SetOrigin(vPoint)
     effectdata:SetScale(1)
-    util.Effect("Explosion", effectdata)
+    util.Effect(self:WaterLevel() > 1 and "WaterSurfaceExplosion" or "Explosion", effectdata)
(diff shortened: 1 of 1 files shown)


entities/entities/lab_base/init.lua
```diff
@@ -34,12 +34,15 @@ end
 function ENT:Destruct()
     local vPoint = self:GetPos()
 
-    util.BlastDamage(self, self, vPoint, 200, 200)
+    util.BlastDamage(self, self, vPoint, 100, 100)
```
FPtje: Why reduce the radius?

AMD-NICK: > Why reduce the radius?

> * because it previously was extremely high 🙃

mcNuggets1: Oh sorry, didn't see the comment.
The radius also was very high.
100 should be alright, especially because it deals 100 damage.

FPtje: That's a nasty way to push an opinion. This pr is about adding effects. Side loading a change in radius is rude.

mcNuggets1: It's a like 5 line code change and I mentioned damage being adjusted.
I wasn't trying to silently push a change, this isn't my first pull request, nor my first issue.

FPtje: Oh, sorry. I missed that.

FPtje:
hl2_2019-02-16_11-20-10
200 really isn't that big. Besides, there's damage dropoff IIRC. In the center you'll get 200 damage, but near the edge you don't get nearly as much.

mcNuggets1:
It's a microwave, not a bomb.

FPtje:
One can argue that gunlabs and drug labs are pretty much bombs.

PR #2836, opened 2019-02-19
Fix MOTD message

FlorianLeChat:
The old link (https://raw.github.com/FPtje/DarkRPMotd/master/motd.txt) [no longer seems to work](https://imgur.com/a/ig4ul7t).
I don't know if this is due to a temporary problem or if the link has changed permanently.


Jophes:
[raw.github.com was deprecated in 2014](https://developer.github.com/changes/2014-04-25-user-content-security/), it moved all sub-domains to githubusercontent.com

Bo98:
It was a temporary outage which has now been resolved. As the blog post mentions, it is supposed to redirect.

PR #2846, opened 2019-02-25
Updated weapon base + weapons

mcNuggets1:
- Removed bone manipulation trash which didn't even work properly.
* Updated the weapon base to support predicted ironsights // code is from TTT like the original ironsights code, but edited to work with DarkRP like it was previous

I tested the whole code in local multiplayer with every possible weapon which could be affected.

### DIFF
--- a/entities/weapons/lockpick/shared.lua
+++ b/entities/weapons/lockpick/shared.lua
@@ -39,14 +39,6 @@ SWEP.Secondary.DefaultClip = -1     -- Default number of bullets in a clip
 SWEP.Secondary.Automatic = false        -- Automatic/Semi Auto
 SWEP.Secondary.Ammo = ""
 
---[[-------------------------------------------------------
-Name: SWEP:Initialize()
-Desc: Called when the weapon is first loaded
----------------------------------------------------------]]
-function SWEP:Initialize()
-    self:SetHoldType("normal")
-end
-
 function SWEP:SetupDataTables()
     self:NetworkVar("Bool", 0, "IsLockpicking")
     self:NetworkVar("Float", 0, "LockpickStartTime")
@@ -56,10 +48,10 @@ function SWEP:SetupDataTables()
     self:NetworkVar("Entity", 0, "LockpickEnt")
 end
 
---[[-------------------------------------------------------
-Name: SWEP:PrimaryAttack()
-Desc: +attack1 has been pressed
----------------------------------------------------------]]
+function SWEP:Initialize()
+    self:SetHoldType("normal")
+end
+
 function SWEP:PrimaryAttack()
     self:SetNextPrimaryFire(CurTime() + 2)
     if self:GetIsLockpicking() then return end
--- a/entities/weapons/ls_sniper/shared.lua
+++ b/entities/weapons/ls_sniper/shared.lua
@@ -46,6 +46,22 @@ function SWEP:SetupDataTables()
     self:NetworkVar("Int", 2, "ScopeLevel")
 end
 
+function SWEP:Deploy()
+    self:GetOwner():SetFOV(0, 0)
+
+    self:SetScopeLevel(0)
+
+    return BaseClass.Deploy(self)
+end
+
+function SWEP:Holster()
+    self:GetOwner():SetFOV(0, 0)
+
+    self:SetScopeLevel(0)
+
+    return BaseClass.Holster(self)
+end
+
 local zoomFOV = {0, 0, 25, 5}
 function SWEP:SecondaryAttack()
     if not self.IronSightsPos then return end
@@ -64,13 +80,4 @@ function SWEP:Reload()
     self:SetScopeLevel(0)
 
     return BaseClass.Reload(self)
-end
-
-function SWEP:Holster()
-    self:GetOwner():SetFOV(0, 0)
-
-    self:SetScopeLevel(0)
-    self:SetIronsights(false)
-
-    return BaseClass.Holster(self)
 end
--- a/entities/weapons/ls_sniper/shared.lua
+++ b/entities/weapons/ls_sniper/shared.lua
@@ -39,9 +39,6 @@ SWEP.Primary.Automatic = false
 SWEP.Primary.Ammo = "smg1"
 SWEP.IronSightsPos = Vector(0, 0, 0) -- this is just to make it disappear so it doesn't show up whilst scoped
 
---[[---------------------------------------------------------------------------
-SetupDataTables
----------------------------------------------------------------------------]]
 function SWEP:SetupDataTables()
     BaseClass.SetupDataTables(self)
     -- Int 0 = BurstBulletNum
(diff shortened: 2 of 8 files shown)


FPtje:
@Bo98 What do you think? I'm asking since you introduced the changes requested to be removed here in 9cdcfde71d74f6db07faed91de9724097e3f41c3

mcNuggets1:
These are broken and only used in the M4 weapon which make it look weird.
Using the default vm looks way cleaner and nicer and doesn't stutter and rob FPS.

mcNuggets1:
I made some final chances to the code after I revisited the whole code again and tested everything in a local server.

mcNuggets1:
This should be the last update.
These are just additional bugfixes for bugs being in the base for centuries.

FPtje:
Could you please in the future have more descriptive commit titles than "updated weapon base #n"? Literally every commit title can start with "updated", it doesn't give any info on what the update actually changes.

The descriptions under the titles are actually good. You can use those as titles, but you may have to split up commits.

mcNuggets1:
The commit names cannot exceed a certain character length, so I just put everything in the description of the commit.
I also wasnt sure about how to call the commits.

Bo98:
Hi, I'll look at this soon.

This why the bone changes were implemented in the first place:



Does this still happen?

mcNuggets1:
They should be the same, however the bone changes were not a good solution, especially because the hand was inside the weapon.

Bo98:
Then something has changed with the model over the years.

Bo98:
Just tried it out. I don't see how "doesn't work". If you're talking about the palm of the hand clipping with the weapon model then that is actually also the case _without_ the bone manipulations. I must have not checked the bottom of the gun before - probably because you can't see it from most angles.

If I were to critique it now, the fingers wrap around way too much. Maybe this is what you're trying to say. I'm just guessing.

If you want to remove it because it's overkill, imperfect or "trash code" then I'm fine with that but I personally think saying it "didn't even work properly" is a bit of stretch.

I would normally recommend filing a bug report at garrysmod-issues but I don't see the point really. It's probably not going to be fixed (there's 679 other unresolved issues there and looking at the still-open list of issues I've reported, 85% are from there with just 1 fixed in the last 3 years) and it's not really designed for what DarkRP uses it for. You can see the no-draw textures while aiming down the sights.

The model is trash and should be replaced ideally but I'm not sure if there's any alternatives without rolling our own.

mcNuggets1:
I just didn't see any sense in having the code there for such a minor issue, especially because the main reason was to make it look less weird.
However the edited version has the hand sticking half inside the weapon.

I just don't see any sense with integrating this into the DarkRP gun base by default.

Bo98:
Forgot to attach this but you got the point anyway:

gm_flatgrass0001

mcNuggets1:
I meant this by the way :P 
https://i.imgur.com/EgFWhSr.png

Bo98:
What the fuck, the bones don't even adjust the right hand?

That doesn't happen when I try it.

Bo98:
Fuck this model, honestly.

Kefta:
All the CS c_models seemed pretty rushed considering most of them have a full skeleton defined for a viewmodel. Just go back to v_models honestly.

mcNuggets1:
Even if no one uses the weapons from the base besides the battering ram, it would be a downgrade to default to the original models in my opinion, especially because there is no c_hands support.

FPtje:
The PR changes the datatable, causing depending weapons to break. I'm really no fan of breaking API changes.
I'm not a fan of that. Also, the M4 is back to a broken state.

mcNuggets1:
Which weapons are depending on the DarkRP weapon base?
I don't know any, but I tried not to change too much.
Also I made all DarkRP weapons compatible with the base.

The M4 isn't better with the bone manipulations, there is still a hand clipping through the model.
We can default to the v_model if you want.

FPtje:
> Which weapons are depending on the DarkRP weapon base?
I don't know, I _cannot_ know and that's the point. There are probably a bunch of servers that have some simple SWEPs based on DarkRP ones.

The M4 definitely looks better before the PR than after.

mcNuggets1:
I can almost guarantee, there are no weapons based on this gun base and if, the chance is low, my changes is gonna break their guns.

But if that happens, I volunteer to fix all their guns by myself, because I know there won't be any people coming around using this gun base with a weapon being demolished with this pull request.

I can revert the M4 change if you'd like, even when I feel more comfortable with leaving it as it is, because this way we don't need bonemanipulations tools inside the gun base and it's still kinda ugly with a hand wrongly placed.

FPtje:
Your overconfidence is unjust. You cannot give such guarantees. Don't pretend that you can. Also, you cannot volunteer to fix all those guns, because for that to happen, you would have to know which guns are affected anyway. Since you're trying to guarantee that they don't exist, I don't have much faith that this will happen.

mcNuggets1:
I assume there are none to 2-3 servers using DarkRP weapons at all, and these servers probably aren't set up completely.

I am able to fix guns, I made my own gun base and converted all 100 m9k guns to it.
Also I did like 100 ttt guns and 20 sandbox guns.

I have the capatibilities to fix a few guns, trust me on that.
Not thousands of course and if people come up with major problems, I am more than happy to fix these incompatibilities in the base itself, even though it would be a regress.

I pushed a lot of pull requests and issues to your addons, so I won't vanish.

Bo98:
The DTVar API is terrible. Indexes should have been removed years ago but we're still stuck with the new NWVar system still in infinite beta under NW2Var. The datatable system could have moved to NW2Var making the index redundant.

Maybe we could make our own system but there's probably reasons why we still have the undocumented NW2Var alias. I've not kept up-to-date with GMod development recently so someone will be able to explain this better than me.

I can look and see what I can do about your issue with the hand clipping if that is desired here. But I can't guarantee anything before Tueday I'm afraid.

mcNuggets1:
It would be decent to fix the hand clipping issue.
I could then add your workaround in the base again.

FPtje:
The nw2vars work fine, but they're limited to player PVS. Addons don't expect that, so switching to nw2vars causes bugs in those addons.

mcNuggets1:
They have some crash issues and bugs though.

Bo98:
Right, the text "will replace the original one at some point in the future" is still littered all over the wiki so I just assumed that was still the plan.

FPtje:
It probably is the intention, but I don't think there is a long term plan for gmod in the first place.

Kefta:
@fptje DTVars also work in the player's PVS but just have different update timing than NW2Vars - you can safely use them as no one expects their var to be networked in x seconds as opposed to y seconds.

Bo98:
I'll look into tweaking the bones a bit today to make everyone happy.

Not sure if the networked vars are still a concern but it seems not so I'll leave touching that for now.

FPtje:
The changed DTVars are acceptable, we'll see what happens.

Bo98:
Oh I see it was a case of add one and remove one. I initially thought it was just adding one.

PR #2849, opened 2019-03-01
Some updates for animations

Wgis:
I think what animations need update and optimization... (without using umsg)

### DIFF
--- a/gamemode/modules/animations/sh_animations.lua
+++ b/gamemode/modules/animations/sh_animations.lua
@@ -27,18 +27,6 @@ function DarkRP.removePlayerGesture(anim)
 end
 
 hook.Add("CalcMainActivity", "darkrp_animations", function(ply, velocity) -- Using hook.Add and not GM:CalcMainActivity to prevent animation problems
-    -- Dropping weapons/money!
-    if ply.anim_DroppingItem then
-        ply:AnimRestartGesture(GESTURE_SLOT_ATTACK_AND_RELOAD, ACT_GMOD_GESTURE_ITEM_DROP, true)
-        ply.anim_DroppingItem = nil
-    end
-
-    -- Giving items!
-    if ply.anim_GivingItem then
-        ply:AnimRestartGesture(GESTURE_SLOT_ATTACK_AND_RELOAD, ACT_GMOD_GESTURE_ITEM_GIVE, true)
-        ply.anim_GivingItem = nil
-    end
-
     if CLIENT then
         if ply.SaidHi then
             ply:AnimRestartGesture(GESTURE_SLOT_ATTACK_AND_RELOAD, ACT_SIGNAL_GROUP, true)
@@ -116,22 +104,6 @@ if SERVER then
     concommand.Add("_DarkRP_DoAnimation", CustomAnim)
     return
 end
-
-local function DropItem(um)
-    local ply = um:ReadEntity()
-    if not IsValid(ply) then return end
-
-    ply.anim_DroppingItem = true
-end
-usermessage.Hook("anim_dropitem", DropItem)
-
-local function GiveItem(um)
-    local ply = um:ReadEntity()
-    if not IsValid(ply) then return end
-
-    ply.anim_GivingItem = true
-end
-usermessage.Hook("anim_giveitem", GiveItem)
 
 local function ThrowPoop(um)
     local ply = um:ReadEntity()
--- a/gamemode/modules/money/sv_money.lua
+++ b/gamemode/modules/money/sv_money.lua
@@ -84,13 +84,7 @@ local function GiveMoney(ply, args)
         return ""
     end
 
-    local RP = RecipientFilter()
-    RP:AddAllPlayers()
-
-    umsg.Start("anim_giveitem", RP)
-        umsg.Entity(ply)
-    umsg.End()
-    ply.anim_GivingItem = true
+    ply:DoAnimationEvent(ACT_GMOD_GESTURE_ITEM_GIVE)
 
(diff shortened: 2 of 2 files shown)


Kefta:
Leave all the clientside code untouched for compatibility in-case anyone calls the usermessage manually.

Wgis:
I think what compatibilities need separate folder and can be deleted if not need.

Kefta:
There's no place for dead code in DarkRP right now so I believe that it should remain in the same place.

Wgis:
I think: old compatibilities is dead code, and not be in DarkRP

Kefta:
I think you're misunderstanding "compatibility." It's a NEW compatibility because you are making it dead code with this PR. Either way, dead code that is registered with the net, hook, or umsg library has to stay for addons to remain functional.

Wgis:
What you mean "dead code" ?

Kefta:
Its code that is never run by the gamemode by default because of changes that made it obsolete. But it has to remain otherwise addons that manually call the usmg (which is a low chance in this case but very present in others) would break since the umsg no longer exists.

Bo98:
I'm not sure where it started but this trend of addons using internal APIs (if you can even call it an API) is horrible. Even though that's not something we have much control over anymore.

Not really specific to this case necessarily, but...

In the past we've been ok with, for example, converting the DarkRPVars system over to net messages because the benefits of doing so vastly outweighed compatibility with terrible addons. Now this animation case is so simple that it doesn't really compare, but for the future we need to remember to not get to a point where we compromise the gamemode for shitty addons.

The worst case of a message we just can't touch anymore is `_Notify`. Now this is something we can take a bit of responsibility for as, unfortunately, the understanding of what the HUD module does is different between us and server developers. The HUD module contains serverside code (/admintell etc.) among other things like handling of `_Notify` - which isn't really a DarkRP concept as it just passes it to GMod's own system. The vast majority of DarkRP HUDs don't consider this notification system (I've lost count how many `_Notify` console errors I see on DarkRP servers) and if they did they should absolutely not look at DarkRP and look at overriding GMod's system. 

Now what a lot of server admins do when they install a custom HUD is they look at `disabled_defaults.lua`. They see the HUD module and they disable it. This is what causes the problems. `_Notify` is now not handled and their custom HUD probably doesn't handle it either. Commands like `/admintell` also cease to work.

One problem that pops up every few months about someone seeing the default GMod/HL2 HUD is also because of disabling the HUD module - DarkRP's HUD module usually hides it and the addon maybe assumes it still will. 

Even without these two cases, people also forget the HUD module covers the "entity display" text...

Because of all this the HUD module should _almost never_ be disabled as 99% of the time it's not what you really want. I honestly recommend removing it from the default list of modules to disable in DarkRPModification, along with a comment telling people to use HUDShouldDraw (like in the HUDReplacement module example) if their addon is shit enough to not handle it itself. Its real consequences are sorely misunderstood.

Anyway, why is this relevant? My question is: apart from `_Notify`, are there other usermessages that are known to be used in addons? If so, there is perhaps a reason why addons are forced to do so and we should, while probably a bit too late, try to at least offer a proper solution for them.

Kefta:
In this specific case I don't think any addons would break since Player:DoAnimationEvent has always existed shared (why this code was written in the first place is beyond me), but Falco and the GMod devs/community in general have always erred on the safe side since the most minuscule changes have broken numerous scripts, and the concept of a "public API" is almost non-existant in GLua, and a weak concept Lua.

It also doesn't help that both GMod and DarkRP are aged codebases that didn't have too much structure in the way of naming/class/API consistency in the first place, making it difficult for inexperienced devs to locate the specific functions and control structures they need.

Wgis:
im try this code for keys and physgun animation and it doesn't work properly

mcNuggets1:
I don't see how this does improve performance in a significant way.

mcNuggets1:
I meant, if we are going to replace some stuff with the actual gmod functions, we should remove the whole hook and do the logic somewhere else.

I'm not entirely sure but the hook call itself should be pretty expensive, because it's called every frame x player count.

Kefta:
It's a clientside hook. It's called once a frame on each client's end and checks if two variables are equal to true. That's hardly expensive.

FPtje:
I don't mind the API change here. As @Kefta said, there's a viable alternative that should have been used in the first place. Other people shouldn't be using those usermessages, and the ones who are can trivially change their code to also use `ply:DoAnimationEvent`

What I don't quite get is that the calls to `ply:DoAnimationEvent` in this PR give `ACT_` enums as arguments, even though the [wiki](https://wiki.garrysmod.com/page/GM/DoAnimationEvent) says they need `PLAYERANIMEVENT_` enums, optionally with an `ACT_` enum as second argument. It seems to work, though, is the code correct?

FPtje:
Never mind that, I was looking at the hook instead of the [function](https://wiki.garrysmod.com/page/Player/DoAnimationEvent). That also explains why it wasn't used in the first place. I can't wiki :V

Wgis:
I think code must be minimal, understand and perfect work without tons of lines. I wanna see darkrp simple and powerfull. It one of reasons starting trend.

mcNuggets1:
I believe, when DoAnimationEvent is called serverside, only other players can see the animation, or none at all.
But the local player can't, I guess.

Sorry, has been I while I worked with that.
But it should be broken from what I know of.

Wgis:
I test with other player, it works perfect for all of us

mcNuggets1:
Then maybe I was wrong.

FPtje:
Yeah, I think we could. For now this PR is approved.

PR #2858, opened 2019-03-15
Fixed invalid phys obj

mcNuggets1:
If phys obj is invalid, don't wake or disable motion.

### DIFF
--- a/entities/entities/spawned_weapon/init.lua
+++ b/entities/entities/spawned_weapon/init.lua
@@ -13,10 +13,16 @@ function ENT:Initialize()
     if not phys:IsValid() then
         self:SetModel("models/weapons/w_rif_ak47.mdl")
         self:PhysicsInit(SOLID_VPHYSICS)
+
         phys = self:GetPhysicsObject()
+
+        if phys:IsValid() then
+            phys:Wake()
+        end
     else
         phys:Wake()
     end
+
 
     if self:Getamount() == 0 then
         self:Setamount(1)
--- a/entities/entities/spawned_weapon/init.lua
+++ b/entities/entities/spawned_weapon/init.lua
@@ -14,9 +14,9 @@ function ENT:Initialize()
         self:SetModel("models/weapons/w_rif_ak47.mdl")
         self:PhysicsInit(SOLID_VPHYSICS)
         phys = self:GetPhysicsObject()
+    else
+        phys:Wake()
     end
-
-    phys:Wake()
 
     if self:Getamount() == 0 then
         self:Setamount(1)
--- a/entities/entities/spawned_weapon/init.lua
+++ b/entities/entities/spawned_weapon/init.lua
@@ -23,7 +23,6 @@ function ENT:Initialize()
         phys:Wake()
     end
 
-
     if self:Getamount() == 0 then
         self:Setamount(1)
     end
--- a/entities/entities/darkrp_billboard/init.lua
+++ b/entities/entities/darkrp_billboard/init.lua
@@ -4,7 +4,7 @@ AddCSLuaFile("shared.lua")
 include("shared.lua")
 
 function ENT:Initialize()
-    self:SetModel(self.Model or "models/props/cs_assault/Billboard.mdl")
+    self:SetModel(self.Model or "models/props/cs_assault/billboard.mdl")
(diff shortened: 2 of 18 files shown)


entities/entities/spawned_weapon/init.lua
```diff
@@ -14,10 +14,10 @@ function ENT:Initialize()
         self:SetModel("models/weapons/w_rif_ak47.mdl")
         self:PhysicsInit(SOLID_VPHYSICS)
         phys = self:GetPhysicsObject()
```
Kefta: You should add another phys:Wake() here

mcNuggets1: Oh yeah.
Thanks for letting me know.

mcNuggets1:
Should be fine now.

Bo98:
What scenarios does this happen apart from not having CS:S mounted?

mcNuggets1:
Invalid food, money.

I just created this pullrequest, because if css is not mounted, there is no need to spam the console with errors, it already tells every player about css not being mounted.

Bo98:
Ok, I was just making sure we weren't silencing a different problem.

Kefta:
Physics objects can also be failed to created if there's too many in the physenv just like entities.

FPtje:
Is there a function to check whether the limit is reached? I'd rather use that than fail silently. With this people are staring at stuff that floats around with no message in console explaining what's going on.

Like we did with the entity limit. Throw a big error explaining what's going on.

Kefta:
Nope, nor does the engine have one. The public vphysics API is extremely limited, but hitting the limit is extremely rare.

Kefta:
You'll almost always hit the entity limit before the physobj assuming your entities have 1-3 physobjs each.

FPtje:
So the expected cause for this error is something on the side of the user, be it CSS not being installed, or a model being set that doesn't exist or doesn't have a physics object.

For CSS not being mounted I understand that the error is redundant. For the other case, though, I think the error is *very* relevant for the user. In that case removing the error would make it more difficult for the user to find out what's going on. IIRC the error isn't that informative in the first place, perhaps we can throw a nicer error that explains the situation better?

The goal here is make it as easy as possible for server owners/developers to solve their own problem, instead of posting on the DarkRP forums about their stuff floating about or errors they don't understand.

mcNuggets1:
There is an error for CSS already.
I never saw it happen, that the physics entity limit was hit.

Kefta:
"Tried to use NULL physobj" also isn't a helpful error. I would change all PhysicsInit calls to something along the lines of:

```
if (not self:PhysicsInit(SOLID_VPHYSICS)) then
   print(string.format("Failed to create the physics of %s, missing model or hit physics limit", self:GetModel() or "")
end
```
``print`` is a bit silent but throwing a lua would be obnoxious on servers with missing models pretty quickly.

FPtje:
Yeah, that's what I had in mind. Some explanation like that.

About print vs error, a missing model renders the entity entirely unusable. It completely breaks it. It's an error state, error states warrant errors to be thrown. I'd use the `DarkRP.error` mechanism, which pesters any joining admin to fix the error. DarkRP should tell them about the problem before they find themselves staring at some floating entity.

Kefta:
I would use ``DarkRP.errornoHalt`` at the very least to let the entity still continue it's creation even without physics.

FPtje:
Yeah, that's the same nagging mechanism. Not halting does mean there could be other errors occurring during initialisation.

Kefta:
That is exactly what this PR fixes - failing to create physics will not create additional Lua errors.

FPtje:
@mcNuggets1 The commit messages in this PR are descriptive. I can tell what they do from reading them. Thanks.

FPtje:
The PR is merged, also with the following change:

`self:PhysicsInit(SOLID_VPHYSICS)` -> `DarkRP.ValidatedPhysicsInit(self, SOLID_VPHYSICS)`

That one throws an error on failure. I've refrained from overriding `PhysicsInit` entirely, since addons might want their own checks and recover in their own way. Overriding it and always throw an error on failure would diminish the return bool of the function.

mcNuggets1:
I think we still check for isvalid in some entities for the physics object.

PR #1988, opened 2015-06-16
Converted all usermessages to netmessages

Kefta:
*Phew*

All these changes were rote conversion and are untested. I call out to
all server owners to test these changes and report back any
errors/functionality problems.

Falco, I noticed on FPP_ShareSettings, you left a comment saying you had
issues with the usermessage being too large on servers with >60 people.
I would test the changes you wanted to make with net messages since they
have an increased query size.

Also, I noticed some weird functionality on player hulls. You broadcasted the hull change clientside of a single player, then, set the hull size of that one player on every client. Was this intentional?

Lastly, I left the VOTE:getFilter() function there for legacy purposes, but changed it to return a table instead.

PR #2860, opened 2019-03-20
Remove deprecated umsg; Use net library instead

Mattzimann:
I do a PR for this because umsg ist deprecated and will be removed, so at some point you have to switch.
And because I am planning on starting a server and wan't it to be reliable I have changed this for it.

### DIFF
--- a/entities/entities/darkrp_laws/cl_init.lua
+++ b/entities/entities/darkrp_laws/cl_init.lua
@@ -37,26 +37,26 @@ local function addLaw(inLaw)
     hook.Run("addLaw", lawNumber, inLaw)
 end
 
-local function umAddLaw(um)
-    local law = um:ReadString()
+local function umAddLaw()
+    local law = net.ReadString()
     timer.Simple(0, fn.Curry(addLaw, 2)(law))
 end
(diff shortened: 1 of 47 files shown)


FPtje:
You're not the first to suggest this. @Kefta Has made a PR before: https://github.com/FPtje/DarkRP/pull/1988

The problem is that many, many addons rely on these usermessages to work. This change would break them. Usermessages are deprecated and inferior to net messages, but there's no way they're going to be removed any time soon. Until we know beyond reasonable doubt that they are indeed going to be removed, I'd rather not break the tons of addons.

Also note that I have tried to change the usermessages of the notification system, but had to revert it here: https://github.com/FPtje/DarkRP/commit/0ea7d028ea3ce17861012cd2371533bee1598b74 because of the enormous backlash I got from shit breaking.

Sorry, I know they're deprecated, but right now the cure is worse than the disease.

mcNuggets1:
Nobody should directly use user messages, instead use a developer function.

oninoni:
What the hell is a developer function?

mcNuggets1:
https://wiki.darkrp.com/index.php/Category:Functions
Functions for direct use as developer.
That's what I called them in all my addons.

PR #2874, opened 2019-04-11
Check if a customer is alive in order to place a hit.

bobbyjoe30:
As of now, dead people can interact with the hit menu and place hits from the grave. This commit should fix this.


gamemode/modules/hitmenu/sh_init.lua
```diff
@@ -30,6 +30,7 @@ DarkRP.getHitmanTeams = fp{fn.Id, hitmanTeams}
 function DarkRP.hooks:canRequestHit(hitman, customer, target, price)
     if not hitman:isHitman() then return false, DarkRP.getPhrase("player_not_hitman") end
     if customer:GetPos():DistToSqr(hitman:GetPos()) > minHitDistanceSqr then return false, DarkRP.getPhrase("distance_too_big") end
+    if not customer:Alive() then return false, Darkrp.getPhrase("must_be_alive_to_do_x", DarkRP.getPhrase("place_a_hit")) end
```
Bo98: Did you test this? `Darkrp` is not the correct casing.

bobbyjoe30: Whoops, I think I fixed that in my test build, but forgot to fix it here. Pretty glaring mistake haha. Good catch

PR #2876, opened 2019-04-15
Add ability to disable chat indicator

sweepyoface:
Title


Bo98:
Do you need this code? Does the module loader not already cover this?

https://github.com/FPtje/DarkRP/blob/fd95cf68e6a01a1b0522c4a989d4512fc0fb202d/gamemode/cl_init.lua#L17

sweepyoface:
Good point, didn't think to check how the other modules were being done.

https://github.com/FPtje/darkrpmodification/pull/138 should cover it.

PR #2880, opened 2019-04-26
Small updates

Wgis:
Remove 3 cycles and make it one. I think it's better.

### DIFF
--- a/gamemode/cl_init.lua
+++ b/gamemode/cl_init.lua
@@ -16,13 +16,8 @@ local function LoadModules()
     for _, folder in SortedPairs(folders, true) do
         if DarkRP.disabledDefaults["modules"][folder] then continue end
 
-        for _, File in SortedPairs(file.Find(root .. folder .. "/sh_*.lua", "LUA"), true) do
-            if File == "sh_interface.lua" then continue end
-            include(root .. folder .. "/" .. File)
-        end
-
-        for _, File in SortedPairs(file.Find(root .. folder .. "/cl_*.lua", "LUA"), true) do
-            if File == "cl_interface.lua" then continue end
+        for _, File in SortedPairs(file.Find(root .. folder .. "/*.lua", "LUA"), true) do
+            if string.find( File:lower(), "_interface" ) then continue end
             include(root .. folder .. "/" .. File)
         end
     end
--- a/gamemode/init.lua
+++ b/gamemode/init.lua
@@ -80,22 +80,18 @@ for _, v in pairs(files) do
 end
 
 for _, folder in SortedPairs(folders, true) do
-    if folder == "." or folder == ".." or DarkRP.disabledDefaults["modules"][folder] then continue end
-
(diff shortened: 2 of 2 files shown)


FPtje:
Looping over every file is different than just cl_ and sh_ files. It means DarkRP will try to load sv_ files when it find them. This can happen when you have DarkRP installed locally.

Wgis:
> Looping over every file is different than just cl_ and sh_ files. It means DarkRP will try to load sv_ files when it find them. This can happen when you have DarkRP installed locally.

I dont see this, but if it regular, you can add prefix check... yep?)

FPtje:
You can, but then what is the optimisation?

Wgis:
I wanna try make small code

gamemode/init.lua
```diff
@@ -80,22 +80,18 @@ for _, v in pairs(files) do
...
-
-    for _, File in SortedPairs(file.Find(fol .. folder .. "/sv_*.lua", "LUA"), true) do
-        if File == "sv_interface.lua" then continue end
-        include(fol .. folder .. "/" .. File)
-    end
-
-    for _, File in SortedPairs(file.Find(fol .. folder .. "/cl_*.lua", "LUA"), true) do
-        if File == "cl_interface.lua" then continue end
-        AddCSLuaFile(fol .. folder .. "/" .. File)
+    if folder == "." or folder == ".." then continue end
```
Bo98: Not sure why you removed the `DarkRP.disabledDefaults` check here.

gamemode/init.lua
```diff
@@ -80,22 +80,18 @@ for _, v in pairs(files) do
...
+    for _, File in SortedPairs(file.Find(fol .. folder .. "/*.lua", "LUA"), true) do
+        if string.find( File:lower(), "_interface" ) then continue end
```
Bo98: This might break addons naming their files `sh_api_interface.lua` or something. I don't know if any addon has that sort of file or not.

gamemode/cl_init.lua
```diff
@@ -16,13 +16,8 @@ local function LoadModules()
...
         if DarkRP.disabledDefaults["modules"][folder] then continue end
 
-        for _, File in SortedPairs(file.Find(root .. folder .. "/sh_*.lua", "LUA"), true) do
-            if File == "sh_interface.lua" then continue end
-            include(root .. folder .. "/" .. File)
-        end
-
-        for _, File in SortedPairs(file.Find(root .. folder .. "/cl_*.lua", "LUA"), true) do
-            if File == "cl_interface.lua" then continue end
+        for _, File in SortedPairs(file.Find(root .. folder .. "/*.lua", "LUA"), true) do
```
Bo98: Load order has changed here so we are now loading client before shared. Does it matter? Who knows - this is just an observation I wanted to note.

FPtje:
Yeah, I appreciate the contribution, but I think this will do more harm than good

PR #2885, opened 2019-04-30
Fixed ammo maths

mcNuggets1:
* Fixed loosing ammo when collecting clip1 ammo twice. Clip1 ammo is automatically added to spare ammo if the ammo type is working.
* Clip2 ammo is automatically substracted from the player, when dropping a weapon with clip2 ammo in it.
The clip2 is automatically added back, when the weapon is collected again.

This took me a bit.
I tested the whole shit for like 10 minutes.
Now it should work flawlessly.

HL2 weapons don't work with ammo, but they are broke as fuck anyway.
It's because GetPrimaryAmmoType returns 0.
It was broken even before I updated the behaviour.

### DIFF
--- a/entities/entities/spawned_weapon/init.lua
+++ b/entities/entities/spawned_weapon/init.lua
@@ -52,26 +52,45 @@ function ENT:Use(activator, caller)
         return
     end
 
-    local ammoType = weapon:GetPrimaryAmmoType()
     local CanPickup = hook.Call("PlayerCanPickupWeapon", GAMEMODE, activator, weapon)
     local ShouldntContinue = hook.Call("PlayerPickupDarkRPWeapon", nil, activator, self, weapon)
     if not CanPickup or ShouldntContinue then return end
 
-    local newAmmo = activator:GetAmmoCount(ammoType) -- Store ammo count before weapon pickup
-
+    -- Store ammo count before weapon pickup
+    local primaryAmmoType = weapon:GetPrimaryAmmoType()
+    local secondaryAmmoType = weapon:GetSecondaryAmmoType()
     weapon:Remove()
 
-    activator:Give(class)
+    weapon = activator:Give(class)
 
-    weapon = activator:GetWeapon(class)
-    newAmmo = newAmmo + (self.ammoadd or 0) -- Gets rid of any ammo given during weapon pickup
-
-    if self.clip1 then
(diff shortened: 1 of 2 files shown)


entities/entities/spawned_weapon/init.lua
```diff
@@ -52,26 +52,45 @@ function ENT:Use(activator, caller)
...
     weapon:Remove()
 
-    activator:Give(class)
-
-    weapon = activator:GetWeapon(class)
-    newAmmo = newAmmo + (self.ammoadd or 0) -- Gets rid of any ammo given during weapon pickup
+    weapon = activator:Give(class)
+
+    local clip1, clip2 = self.clip1, self.clip2
+    if weapon:IsValid() then
```
FPtje: This will never be true because the weapon is removed 5 lines above this line

entities/entities/spawned_weapon/init.lua
```diff
@@ -52,26 +52,45 @@ function ENT:Use(activator, caller)
...
+            clip1 = 0
+        end
+        if clip2 and clip2 ~= -1 and weapon:Clip2() ~= -1 then
+            weapon:SetClip2(self.clip2)
+            clip2 = 0
+        end
+    else
+        weapon = activator:GetWeapon(class)
+        if clip2 and clip2 > 0 and weapon:Clip2() ~= -1 then
+            weapon:SetClip2(weapon:Clip2() + clip2)
```
FPtje: Shouldn't this set secondary ammo instead? Your original issue was that clips shouldn't be affected. Clip2 should ideally have the same behaviour as Clip1

entities/entities/spawned_weapon/init.lua
mcNuggets1: weapon = activator:Give(class)

entities/entities/spawned_weapon/init.lua
mcNuggets1: Clip2 is such a broken mess.
Clip2 by default cannot be reloaded, thus just adding ammo won't work.

mcNuggets1: Some weapons don't use Clip2, but the Ammo Count of weapons.
I added possible spare secondary ammo.

mcNuggets1:
The issue should be revoled.
The first one should be invalid.

PR #2889, opened 2019-05-03
spawned_shipment use engine behaviour for ammo

mcNuggets1:
Currently, spawning 2 weapons from a spawned shipment will make you loose ammo, because clip1 isn't stored. (Only the first weapon is given a full clip1)

This pull request corrects this behaviour, making it match the engine and fixes loosing ammo due to shipments completely.

### DIFF
--- a/entities/entities/spawned_shipment/init.lua
+++ b/entities/entities/spawned_shipment/init.lua
@@ -130,10 +130,16 @@ function ENT:SpawnItem()
     local class = CustomShipments[contents].entity
     local model = CustomShipments[contents].model
 
+    local defaultClip
+    local wep_tbl = weapons.Get(class)
+    if wep_tbl and wep_tbl.Primary and wep_tbl.Primary.DefaultClip then
+        defaultClip = wep_tbl.Primary.DefaultClip
+    end
+
     weapon:SetWeaponClass(class)
     weapon:SetModel(model)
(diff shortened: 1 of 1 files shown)


entities/entities/spawned_shipment/init.lua
```diff
@@ -130,10 +130,16 @@ function ENT:SpawnItem()
     local class = CustomShipments[contents].entity
     local model = CustomShipments[contents].model
 
+    local defaultClip
+    local wep_tbl = weapons.Get(class)
```
Bo98: With your history of micro-optimisations, I expected this to be `weapons.GetStored`.

Kefta: weapons.GetStored doesn't inherit settings from superclasses. Really, the weapons/scripted_ents libraries should have some kind of GetInherited function that caches the generated table.

FPtje: Better use CPU there to micro optimise memory usage

Bo98: This is the type of thing that should be documented on the wiki but is not. The wiki recommends using `weapons.GetStored` where possible.

mcNuggets1:
For clarification, this has nothing to so with microoptimisations for me, it's just tidier and easier to understand.
Calling weapons.Get 4 times in 2 lines just looks like bad untidy code.
I don't believe the performance impact is noticable however.

I kept using weapons.Get because it was used before and I didn't want to potentially break code. (Which is in Gmod extremely easy)

PR #2893, opened 2019-05-06
Fixed duplication of ammo

mcNuggets1:
* The Give function gave players additional ammo, while they shouldn't receive any, fixed.
* DefaultClip was used to determine clip1 fill, now ClipSize is used, that's how it should have been.
* sleep module, weapon checker were making player loose ammo, fixed.

### DIFF
--- a/entities/entities/spawned_shipment/init.lua
+++ b/entities/entities/spawned_shipment/init.lua
@@ -130,16 +130,17 @@ function ENT:SpawnItem()
     local class = CustomShipments[contents].entity
     local model = CustomShipments[contents].model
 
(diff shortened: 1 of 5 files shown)


Bo98:
Question: what happens if DefaultClip is _less than_ ClipSize? (Not tried it, just asking)

mcNuggets1:
I didn't try either.

But a clip of 2 bullets for example with a default clip of 4, will give you 2 clip for the clip and 2 for reserve.
I think having 1 for default clip is just giving you 1 in clip.

It would only make sense and I'm 99% sure, it's handled that way.

PR #2898, opened 2019-05-11
Update cl_menu.lua

Wgis:
Del weird update items lag


FPtje:
You're just changing behaviour to close the frame when you spawn something. This has nothing to do with lag.

PR #2899, opened 2019-05-11
Convert #table to table.IsEmpty

Kapppa:
I was going to extend my other pull req however, didn't know if you wanted it done. This should be most of them. I looked for them using #([A-z])+ regex check.

### DIFF
--- a/gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
+++ b/gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
@@ -26,11 +26,11 @@ function FAdmin.TargetsToString(targets)
         return FAdmin.PlayerName(targets)
     end
 
-    if table.IsEmpty(targets) then
+    local targetCount = #targets
+    if targetCount == 0 then
         return "no one"
     end
 
-    local targetCount = #targets
     if targetCount == player.GetCount() and targetCount ~= 1 then
         return "everyone"
     end
--- a/gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
+++ b/gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
@@ -26,11 +26,11 @@ function FAdmin.TargetsToString(targets)
         return FAdmin.PlayerName(targets)
     end
 
-    local targetCount = #targets
-    if targetCount == 0 then
+    if table.IsEmpty(targets) then
         return "no one"
     end
 
+    local targetCount = #targets
     if targetCount == player.GetCount() and targetCount ~= 1 then
         return "everyone"
     end
--- a/gamemode/modules/fpp/pp/server/ownability.lua
+++ b/gamemode/modules/fpp/pp/server/ownability.lua
@@ -268,9 +268,9 @@ local function netWriteEntData(ply, ent)
 end
 
 function FPP.plySendTouchData(ply, ents)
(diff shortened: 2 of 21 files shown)


Bo98:
I'd argue some of these aren't "optimisations" at all, but it otherwise looks ok.

Kapppa:
I was thinking this after a friend pointed it out at it will be using two global calls. However I did some testing and got 
```
#table == 0
9.7523843578529e-07
table.IsEmpty
2.4380969989579e-06
#table == 0
3.9009547435853e-05
table.IsEmpty
2.6819061531569e-06
```
Just ran the same code twice. Compared an array of 50 length 25 times.

Bo98:
There's also the situations where you do a `table.IsEmpty` check and then go ahead and calculate the length anyway if it's not empty (which might be a more common scenario?). But this is all really minor and shouldn't matter.

gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
```diff
@@ -26,11 +26,11 @@ function FAdmin.TargetsToString(targets)
         return FAdmin.PlayerName(targets)
     end
 
-    local targetCount = #targets
-    if targetCount == 0 then
+    if table.IsEmpty(targets) then
```
Kefta: Dont use IsEmpty here since you use the length below.

gamemode/modules/fpp/pp/server/ownability.lua
```diff
@@ -268,9 +268,9 @@ local function netWriteEntData(ply, ent)
 end
 
 function FPP.plySendTouchData(ply, ents)
+    if table.IsEmpty(ents) then return end
+    
     local count = #ents
```
Kefta: Same here.

Bo98:
Yeah those were the two situations I meant

FPtje:
Don't change FPP code in the DarkRP repository. FPP has its own repository.

gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
Kapppa: I was debating not doing this, as if the table is empty is will be faster but trades of performance if they array is filled. As you have complained I will remove

FPtje:
Merged, but FPP and MySQLite have been reverted to their original versions.

Bo98:
This isn't really a problem specific to this pull request, nor really is it something that needs action taken but it's worth mentioning.

Recently, I was made aware of this that was sent on the Discord server: 

This is really a problem with GMod. Most GMod updates no longer require you to update your server. This includes the April 30th one, despite claims otherwise. You could have a server from a few updates back and people can join fine.

When this is eventually merged to Steam Workshop, anyone who restarts their server (which they may do automatically on a daily schedule) will notice everything has broken until they update their server. Unfortunately, there's not exactly anything that notifies them of that.

This is not a DarkRP-specific problem. Any addon the server could have could start using new features that will cause breakages on servers on older versions. I'm just mentioning it here as if you see someone mention the error, or if anyone with the issue stumbles upon this pull request, it's probably because the server needs updating.

Kefta:
Sounds like DarkRP should do a minimum _VERSION check and notify the server owner with a message telling them their server is outdated.

PR #2900, opened 2019-05-19
Created A Config Value

BLEEPBLOOPforLife:
Created a config value called GM.Config.entitySpawnDelay and used it to be able to change the purchasing delay of an entity.


FPtje:
Wouldn't it be better to be able to set that per entity?

BLEEPBLOOPforLife:
You're probably right. I'll edit it later.

PR #2901, opened 2019-05-19
Add a delay option

BLEEPBLOOPforLife:
Adds a delay option in custom entity creation with tblEnt.delay.


FPtje:
Aan entry should be added to modules/base/sh_checkitems.lua

BLEEPBLOOPforLife:
Is there a way to edit pull request commits because I can't seem to find it.

PR #2902, opened 2019-05-19
Add a delay option

BLEEPBLOOPforLife:
Adds a delay option in custom entity creation with tblEnt.delay and adds validation for the delay option.

### DIFF
--- a/gamemode/modules/base/sh_checkitems.lua
+++ b/gamemode/modules/base/sh_checkitems.lua
@@ -487,6 +487,12 @@ DarkRP.validateEntity = fn.FAnd{buyableSchema, tc.checkTable{
                 "The allowTools must be either true or false."
             )
         ),
+
+    delay =
+        tc.addHint(
+            tc.optional(isnumber),
+            "The delay must be a number."
+        ),
 }}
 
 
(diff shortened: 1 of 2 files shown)


BLEEPBLOOPforLife:
So, is this being added?

FPtje:
I usually use weekends to work on DarkRP. I'll look at this later today.

BLEEPBLOOPforLife:
Got it.

FPtje:
Merged. Some feedback: 

Use more descriptive names than "Updated <file.lua>". Also, it should have been a single commit. The idea of a commit is that it describes a unit of change over the entire repository. Ideally, DarkRP should work properly at every commit in history. In this case the change is split into two commits, creating a point in the commit history where a commit is half implemented. In this case that won't lead to errors, but in more sophisticated changes it might.

Regardless, I'm happy with the contribution. The delay of two seconds isn't ideal for every entity's chat command, and you're right in that it should be possible to set it per entity.

Kefta:
@FPtje As repository owner, you can squash all commits on merge.

FPtje:
I know

BLEEPBLOOPforLife:
Alright. Will do next time.

Issue #1852, opened 2015-03-26
Logging


advancedrazor:
I would like a option for disable just some logging features but not all, as some are useful tho.

FPtje:
I don't want a config hell.

advancedrazor:
Just because you dont want it?

Many want it, and I love DarkRP also.
This feature is normally main for every gamemode that has its own admin system.

FPtje:
The DarkRP logging system is basic and provides basic functionality. Being
able to set every logging option in the config file makes it bloated. It'll
make the config file even larger and most people wouldn't even use it.

There are specialised logging mods that can be super fancy if you want to.
Use those.

advancedrazor:
Its not basic, its perfectly for DarkRP.
Additionally you could let the user setup the "config" in the fadmin interface.
I didnt ever read something about FAdmin in the config file also.

FPtje:
Are you talking about FAdmin logging or DarkRP logging?

advancedrazor:
FAdmin normally logs things like Hit accepted or spawned prop

FPtje:
I know what my own admin mod does, I was asking whether you were talking
about FAdmin logging or DarkRP logging.

Neosinos:
How about not making a config option? Just console commands you can
auto-load in your server.cfg

Am Donnerstag, 26. März 2015 schrieb Razor Glaze :

FPtje:
That's even worse

advancedrazor:
I just dont like that it logs double.
DarkRP + ULX.
It pretty spamms my console.
And DarkRP doesnt log tool uses etc. so I would like to disable some things, but no Config for that.
Even Hitaccepts etc. is logged, which are quite important for servers, to look after RDMs etc.

Neosinos:
Then disable ULX Logging / Console Output / Echo.

Am Donnerstag, 26. März 2015 schrieb Razor Glaze :

FPtje:
FAdmin would mostly log the same things as ulx. DarkRP logs the hits.

In conclusion I still don't know whether you're talking about FAdmin
logging or DarkRP logging.

advancedrazor:
Then Im talking about FAdmin.

FPtje:
You can remove FAdmin's logging hooks in a darkrpmod module.

advancedrazor:
And why not a config for this?
would make things easier to setup etc.

FPtje:
[...]

PR #2903, opened 2019-05-25
GAMEMODE.Config.jobBecomeNotification

BLEEPBLOOPforLife:
Creates GAMEMODE.Config.jobBecomeNotification, which is a config value that allows the server owner to choose whether to display the notification notifying everyone about a job switch.


FPtje:
I'm sorry, but I don't agree with this change. There shouldn't be config options for disabling single notifications, as it sets a bad precedent. The reason for adding an option for this specific notification can justify options for the dozens of other notifications that live in DarkRP.

Granted, there may be a valid use case for not wanting to see specific notifications, but I reckon a better approach would be to have some generic means of deciding which notifications you don't want to appear.

There was a similar discussion about adding options for logging here, the arguments are similar: https://github.com/FPtje/DarkRP/issues/1852

PR #2905, opened 2019-05-25
Adds the onNotify and onNotifyAll hook.

BLEEPBLOOPforLife:
The onNotify hook is a serverside hook that runs when DarkRP.notify is run with the parameters: ply, msgtype, len, and msg.

The onNotifyAll hook is a serverside hook that runs when DarkRP.notifyAll is run with the parameters: msgtype, len, and msg.

### DIFF
--- a/gamemode/modules/base/sv_util.lua
+++ b/gamemode/modules/base/sv_util.lua
@@ -14,6 +14,8 @@ function DarkRP.notify(ply, msgtype, len, msg)
         rcp:AddPlayer(v)
     end
 
+    hook.Run("onNotify", ply, msgtype, len, msg)
+
     umsg.Start("_Notify", rcp)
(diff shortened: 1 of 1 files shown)


Kefta:
I think it should be one hook onNotify and when notifyAll is called, send in player.GetAll. I also think you should be able to block the notification by returning true.

BLEEPBLOOPforLife:
Whoops, you're right. I was going to make it so you could block the notification, but I forgot.

FPtje:
Why choose the pull request? You can change stuff by pushing to the same branch. Then rebase to make the commits nicer. Please see the GitHub for Windows program if you're not too familiar with git yet.

BLEEPBLOOPforLife:
I'll install that and try it out.

PR #2906, opened 2019-05-25
Adds the onNotify hook.

BLEEPBLOOPforLife:
The onNotify hook is a serverside hook that runs when DarkRP.notify is run with the parameters: ply (table), msgtype, len, and msg.


FPtje:
Note: I've edited the commit with the following changes:

- Use `player.GetAll()` instead of `player.GetHumans()`. I think this is less surprising for developers
- The first argument is *always* a table now, so the type is consistent
- Return `true` to suppress message, as opposed to false. This is more consistent with the other hooks
- Added documentation.

PR #2912, opened 2019-06-02
something to encourage servers to stop using CAC

joeyjumper94:
turns out people use CAC even though work on it stopped years ago,
https://forum.facepunch.com/gmoddev/bupuf/CAC-false-positives-on-64-bit-clients
https://forum.facepunch.com/f/gmoddev/bssdp/cake-Anti-Cheat-1-Line-Bypass

### DIFF
--- a/gamemode/modules/workarounds/sh_workarounds.lua
+++ b/gamemode/modules/workarounds/sh_workarounds.lua
@@ -352,3 +352,46 @@ if not DarkRP.disabledDefaults["workarounds"]["Deprecated console commands"] the
         concommand.Add(cmd.command, fp{msgDeprecated, cmd})
     end
 end
+
+--[[---------------------------------------------------------------------------
+so i created this cause CAC is still used by servers, is out
+of date, can be bypassed with one line of code, and actually
+kicks innocent people when they are using the x86-64 gmod branch
+---------------------------------------------------------------------------]]
+hook.Add("InitPostEntity","fuckCAC",function()
+    timer.Simple(1,function()
+        if DarkRP and DarkRP.disabledDefaults.workarounds["CAC"] then return end--option to disable it
+        if CAC then
+            --remove CAC's hooks
+            hook.Remove("CheckPassword","CAC.CheckPassword")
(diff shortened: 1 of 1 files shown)


mcNuggets1:
We don't need another workaround.
We already have like thousands of them, which are most likely unnecessary  now.

Bo98:
Reminds me of a different shitty AC I came across years ago. Got permabanned from one server for having modified _menu_ code due to a pull request I had open. I did manage to convince the author to remove the check entirely so at least they did that.

Meanwhile (not sure if this is the case now though) there were many servers using them to block clientside scripts when they have clientside scripts enabled...

Proper ACs across games aim for a virtually 0% false positive rate. Sadly, the same can’t be said in general for these various ones for GMod.

CAC at least was one of the better ones but there are grave dangers to using unmaintained ACs and it’s unfortunate that it’s still available to download.

joeyjumper94:
thank you B098, you understand why this needs to be done,

Kefta:
I still don't think DarkRP should be the addon police. Most of the workarounds already in the gamemode are entirely obsolete - HexAC is dead, APAnti doesn't even use that hook anymore, etc.

FPtje:
DarkRP being addon police in the past has had very positive effects in the past. I still stand by the decisions I've made in the past. These workarounds have stopped innocent people being fucked over, improved player experience and played key parts in discussions in the coding community. The one about DarkRP being addon police being one of them, but also about ban lists, shitty anti cheats, and the lot.

I've contacted cake about it. I think it's a good idea to break it in DarkRP if cake refuses to fix it.
We're not there yet, though. I've pinged him, and I'm still waiting for a response. For all I know he'll fix it and we're good.

joeyjumper94:
any news?

FPtje:
Yeah, looks like cake has no means to update the addon, so to prevent it from kicking innocent people I agree with this PR. I would like to change the wording, though. I'll look into that tomorrow.

mcNuggets1:
Why does it add a new InitPostEntity hook, when there is already one?
Why is it only removed, if the CAC check is enabled?

Two issues.

PR #2916, opened 2019-06-16
Replacing the default page link in the F1 menu

FlorianLeChat:
The old page can no longer be displayed with Awesomium.


Bo98:
Slightly cautious about making it the DarkRP forum, as people may think it is the forum for that particular server.

FlorianLeChat:
I thought the players were making that difference, but you're absolutely right.

mcNuggets1:
We could go back to the DarkRP wiki.

PR #2920, opened 2019-06-21
Updated spawning of entities to prevent unstucking

mcNuggets1:
This pull request fixes a lot of situations, in which entities are spawned in a wall or weapons get lost because dropping in the floor.

It uses the same functions as sandbox to ensure everything spawns properly on solid surface.

If an entity has an spawn function written onto it, it will not try to unstuck it, just like in the sandbox gamemode.

### DIFF
--- a/entities/weapons/weapon_cs_base2/sv_commands.lua
+++ b/entities/weapons/weapon_cs_base2/sv_commands.lua
@@ -16,10 +16,24 @@ function meta:dropDRPWeapon(weapon)
     self:DropWeapon(weapon) -- Drop it so the model isn't the viewmodel
 
     local ent = ents.Create("spawned_weapon")
+
     local model = (weapon:GetModel() == "models/weapons/v_physcannon.mdl" and "models/weapons/w_physics.mdl") or weapon:GetModel()
     model = util.IsValidModel(model) and model or "models/weapons/w_rif_ak47.mdl"
 
-    ent:SetPos(self:GetShootPos() + self:GetAimVector() * 30)
+    local trace = {}
+    trace.start = self:GetShootPos()
+    trace.endpos = self:GetShootPos() + self:GetAimVector() * 30
+    trace.filter = self
+
+    local tr = util.TraceLine(trace)
+
+    local ang = ply:EyeAngles()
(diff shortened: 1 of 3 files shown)


mcNuggets1:
Deeply sorry for the extra 2 pull requests.
But I am used to working with tabs instead of spaces and also it's quite a lot code, even if it is pretty repetitive.

FPtje:
Your commits are added to this PR. You can squash them and force push to update your PR.

mcNuggets1:
I'm slow in mind.
I don't understand, what you exactly mean by that.

Do you mean, deleting the whole PR, making another one and only doing 1 commit, because that's the only thing I am capable of.
I am still not used to Github, especially because I don't use it apart from pull request for DarkRP.

FPtje:
See explanation of rebasing, squashing and force pushing here: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase

mcNuggets1:
I'm using Github Desktop, I never worked with Git.
So I guess I have to install Git, use the command line interface to rebase the pull request and then everything should be fine?
I'm confused, because you sent me a huge text to read and I only understood like 30% of it.

FPtje:
You can rewrite commits, that's basically what rebasing is. You can join several commits to become one commit. That's called squashing. You can also edit commits in hindsight and change a bunch of things about them. The biggest caveat there is that editing commits after they've been pushed to Github has consequences. You're basically changing the "history" of commits, meaning that some commits that were on github are deleted in favour of the ones overwriting them. Git won't do this willy nilly, and will throw an error if you try to push normally. To ignore the error, one must force push. There's also force push with lease, but that's not important right now.

Rewriting git history and force pushing is problematic if you have multiple people working on the same branch. With pull requests this is not the case. It's just you working on your branch that you request to be merged into DarkRP's master branch.

You might have to use the git command line tool to do these things, but maybe Github Desktop supports it as well. I don't know. It looks like support was added here: https://github.blog/2019-06-05-github-desktop-expands-to-support-stashing-and-rebasing/

mcNuggets1:
After 1 hour of trying, I'm just too stupid to get it to work.
May I just create another pull request?

PR #2924, opened 2019-06-26
Fixed cool down on tip jar not being set

deadmonstor:
Must have been accidentally removed when you were making it


FPtje:
Hmm, one of the fun parts of the tipjar is spamming the donate button. I might want to remove the delay altogether or set the delay really low.

PR #2926, opened 2019-06-30
Fix pattern for bind fspectate

realpack:
```
| It creates an error.
|
|     | Spaces are not needed.
|     |   |
V     V   V
+([a-z A-Z 0-9]+)


FPtje:
This isn't the repo for FSpectate. The right repo is here: https://github.com/FPtje/FSpectate

PR #2929, opened 2019-07-02
Recursive function to add all files

RoyalScythe:
Adds all files within folders of folders - rather than adding the same code in each module


FPtje:
I'd rather not change this code if not strictly necessary. What's the motivation for this change?

FlorianLeChat:
This change allows files to be included in sub-folders ?
Like for example  "darkrp_modules/mainfolder/subfolder1/myscript.lua"

FPtje:
I can see what it does. What I can't see is why. The module loading works fine as it is. DarkRP's modules don't need recursive loading since there are no modules defined in subfolders. This change can only break things.

PR #2941, opened 2019-08-15
Add postLoadCustomDarkRPItems

Kefta:
I am writing an entity for DarkRP that can only be registered after all teams have been loaded. This is not guaranteed by a loadCustomDarkRPItems hook due to unpredictable hook ordering. Though a timer would work, I also want the entity to be registered by the time map entities are loaded so I don't have to do extra hacks to recreate them after InitPostEntity.

GAMEMODE -> self when passing to hook.Call to support custom GAMEMODE tables (``GAMEMODE.DarkRPFinishedLoading(MY_CUSTOM_GM_TABLE)`` or ``local GM = setmetatable({}, {__index = GAMEMODE}) GM:DarkRPFinishedLoading()``). This should really be done across the entire gamemode.

### DIFF
--- a/gamemode/libraries/modificationloader.lua
+++ b/gamemode/libraries/modificationloader.lua
@@ -147,5 +147,6 @@ function GM:DarkRPFinishedLoading()
     loadLanguages()
     loadModules()
     loadCustomDarkRPItems()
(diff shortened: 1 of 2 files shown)


FPtje:
Make a stub

Kefta:
Done

PR #2947, opened 2019-09-12
Update sv_data.lua

ImStayFrosty:
By using INTEGER it sets a default wallet hard cap of $2,147,483,647  by changing to a BIGINT this hard cap is much higher at $9,223,372,036,854,775,807 which seems unrealistic. Without the change of INTEGER to BIGINT a players money might be reset due to an sql error of "too high range" since the value of wallet is bigger than the max size of INT.

### DIFF
--- a/gamemode/modules/base/sv_data.lua
+++ b/gamemode/modules/base/sv_data.lua
@@ -54,7 +54,7 @@ function DarkRP.initDatabase()
                 uid BIGINT NOT NULL PRIMARY KEY,
                 rpname VARCHAR(45),
                 salary INTEGER NOT NULL DEFAULT 45,
-                wallet INTEGER NOT NULL
+                wallet BIGINT NOT NULL
             );
         ]])
 
@@ -191,7 +191,7 @@ function migrateDB(callback)
                     uid BIGINT NOT NULL PRIMARY KEY,
                     rpname VARCHAR(45),
                     salary INTEGER NOT NULL DEFAULT 45,
-                    wallet INTEGER NOT NULL
+                    wallet BIGINT NOT NULL
                 );
             ]])
 


FPtje:
I don't think you can just do that without a migration

ImStayFrosty:
`        -- Fix money cap
        MySQLite.query([[
            ALTER TABLE darkrp_player MODIFY wallet BIGINT;
        ]])`
I just added this in the same file for my server and it worked like a charm to migrate

mcNuggets1:
Or just don't migrate it?
It won't be much different at all.

FPtje:
Bad idea, that'll get you different behaviour depending on when the server was made. Also, why would you make an improvement if it doesn't apply to literally any of the current running servers?

mcNuggets1:
To reduce unnecessary code.
People who want it, can migrate it themselves.

But I guess your point does make sense for a addon used by so many peoples.

ImStayFrosty:
Just figured I'd try and help out considering it's a hard cap on literally every DarkRP server that hasn't figured out the issue themselves. Any player that makes it over ~2.1 bil gets their money reset. Maybe make an announcement or something bringing attention to the issue?

ImStayFrosty:
Or just fix it for any new server that is downloading from this repository

Bo98:
In addition to this change, it would probably be a good idea to cap it on the Lua side in `ply:addMoney` or something, just in case someone does get that high.

Bo98:
That said, there’s perhaps also restrictions on the DarkRPVar side - I haven’t checked how many bits are allocated for money.

ImStayFrosty:
Just thought I would bring the issue to light since I couldn't find anything else relating to a fix or known problem

Bo98:
Yeah, money resetting is never a good thing.

ImStayFrosty:
Especially for a server owner who might not want to edit the darkrp core files considering there is files everywhere condemning editing them

Bo98:
The only other thing I'll add, is due to Lua using a `double` internally for numbers, there is only really integer precision up to 2^53. If you do `print(2^53, 2^53 + 1)` you notice the two values are the same. This can be limited on the Lua side in `ply:addMoney`/`DarkRP.storeMoney` etc.

The money DarkRPVar uses net.Read/WriteDouble, so there isn't any additional restrictions there.

FPtje:
> To reduce unnecessary code.

This is a pretty big issue, code that fixes it cannot in any way, shape or form be regarded as unnecessary.

FPtje:
This one?
> Or just fix it for any new server that is downloading from this repository

I'd rather have a migration. We already have a system in place, and it would fix the issue for everyone. It would cause the server to take a bit longer to start the first time after the update, but after that it's fixed for everyone. Whenever someone asks about this problem, we'll know that all they have to do is update DarkRP.

ImStayFrosty:
“-- Fix money cap MySQLite.query([[ ALTER TABLE darkrp_player MODIFY wallet BIGINT; ]])” why wouldn’t this work? Not to be a pain I’m just curious so I would know for the future

FPtje:
That's part of the migration I had in mind, along with the mechanism we already have to perform migrations only once and only when necessary.

Bo98:
It’s also worth noting that I think it’s only necessary on MySQL. SQLite only has one (adaptable) integer type doesn’t it?

FPtje:
@Bo98 You're right: https://www.sqlite.org/datatype3.html

BIGINT will work in SQLite, but it translates to INTEGER. Meaning a migration on SQLite would probably not change anything, rendering it inert.

FPtje:
Holy shit the migration mechanism is extreme poop. I gotta rewrite that shit to be at least palatable. There currently is no easy way to add a migration.

FPtje:
So basically, the migration system was shit. It always ran some queries that only had to be run once. Some migrations used the DB version mechanism, others just didn't. You can run the next migration only when the previous one has been committed, but that was messy too. Adding a migration would force me to nest another anonymous function somewhere in that mess.

Now I could have gone full abstraction and stored migrations in a table with a version number and migration function, but this is fine. To prevent deep nesting there's now a recursive `migrate` function. Some code duplication every migration, but that's fine. 

Testing this was pretty fun, for both SQLite and MySQL I had to drop the database, checkout to a version of DarkRP before all migrations (I chose 6e43d9f4), have it recreate the database, go back to latest and see if all the migrations run without error.

Goes to show how useful git is in actual development and testing changes. Also, it turns out  the version of DarkRP at June 3rd 2016 still actually runs just fine in the current version of GMod.

PR #2949, opened 2019-09-17
Wanted improvements

MilkGames:
This prevents players from jumping off buildings to get unwanted and also allows GAMEMODE.Config.wantedtime to be set to 0 (will not create a timer, wanted will only go away when killed)

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -460,7 +460,7 @@ function GM:PlayerDeath(ply, weapon, killer)
         end
     end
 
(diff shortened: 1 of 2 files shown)


mcNuggets1:
When you die, you should not be wanted anymore.

Kefta:
I think it should be made into a config option, it's a preference if a server wants to prioritise NLR over using death to escape a want.

FPtje:
DarkRP is not built for NLR. I don't mind changes that fuck it over. Removing wanted on death is something that can always be scripted externally, the other way around is more difficult.

FPtje:
The wantedtimer 0 is a good idea, the isValid doesn't seem like a too popular option. I've selectively merged.

PR #2956, opened 2019-10-25
Config option to allow remaining wanted after death

vliberio:
This adds an option that skips the code setting the DarkRPVar, wanted, to nil on death by players other then themselves. When false (default), it continues the current behavior of removing the wanted status of the perp in most cases. 

This has been discussed in #2949, but it seems there might have been a misunderstanding. Players are indeed losing their wanted status on death if it wasn't from suicide, etc. Falco summarizes the issue with this in his comment:
> DarkRP is not built for NLR. I don't mind changes that fuck it over. Removing wanted on death is something that can always be scripted externally, the other way around is more difficult.

_Originally posted by @FPtje in https://github.com/FPtje/DarkRP/pull/2949#issuecomment-535073645_

On top of that, since it doesn't call the playerUnwanted hook or use the unwanted player function, there is no clean way of blocking, handling, logging, or otherwise dealing with this. 

This won't modify the current behavior unless you set it to true, but it will permit server owners to keep people wanted after death so that even through the fiery red virtual hell of your HL2-style demise, you cannot escape the highly trained professionals of {insert name}RP and their unyielding pursuit of justice.

We appreciate your time and consideration.

### DIFF
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -159,7 +159,7 @@ GM.Config.unlockdoorsonstart            = false
 GM.Config.voiceradius                   = true
 -- tax - Whether players pay taxes on their wallets.
 GM.Config.wallettax                     = false
--- wantedrespawn
+-- wantedrespawn - Whether players remain wanted on respawn.
 GM.Config.wantedrespawn                 = false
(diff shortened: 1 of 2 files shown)


FPtje:
LGTM

PR #2959, opened 2019-11-01
More types check changes and minor cleanup

FlorianLeChat:
More changes following the last changes of this one https://github.com/FPtje/DarkRP/commit/8ef48734f6dab07aa8d1f3cf4bcb2e76568c5e10

### DIFF
--- a/gamemode/modules/fadmin/fadmin/cl_interface/cl_scoreboardPlayer.lua
+++ b/gamemode/modules/fadmin/fadmin/cl_interface/cl_scoreboardPlayer.lua
@@ -129,16 +129,16 @@ function FAdmin.ScoreBoard.Player.Show(ply)
     FAdmin.ScoreBoard.Player.Controls.ButtonPanel:Clear()
     FAdmin.ScoreBoard.Player.Controls.ButtonPanel:DockMargin(5, 5, 5, 5)
 
-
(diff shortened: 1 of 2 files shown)


gamemode/modules/fadmin/fadmin/cl_interface/cl_scoreboardPlayer.lua
```diff
@@ -129,16 +129,16 @@ function FAdmin.ScoreBoard.Player.Show(ply)
     FAdmin.ScoreBoard.Player.Controls.ButtonPanel:Clear()
     FAdmin.ScoreBoard.Player.Controls.ButtonPanel:DockMargin(5, 5, 5, 5)
 
-
     for _, v in ipairs(FAdmin.ScoreBoard.Player.ActionButtons) do
-        if v.Visible == true or (isfunction(v.Visible) and v.Visible(FAdmin.ScoreBoard.Player.Player) == true) then
```
FPtje: This changes behaviour. When Visible is a function that returns false, the condition would still be true after this change. That's a bug.

FlorianLeChat: Sorry, I didn't realize that.

PR #2963, opened 2019-11-21
Don't use hands with viewmodel which don't support them

roboderpy:
I got a truck of issue reports about floating hands in PPM/2 and now you can assume i went completely mad from that.


FPtje:
What is PPM/2?

FPtje:
Seriously, what is that?

Kefta:
I mean, UseHands should still be false regardless. v_stunbaton is a v_model.

roboderpy:
Some brony stuff

PR #2970, opened 2019-12-27
Update init.lua

pilot22:
Avoid following error:
[ERROR] gamemodes/darkrp/entities/entities/spawned_weapon/init.lua:13: Tried to use a NULL physics object!
1. Wake - [C]:-1
2. unknown - gamemodes/darkrp/entities/entities/spawned_weapon/init.lua:13
3. Spawn - [C]:-1
4. dropDRPWeapon - gamemodes/darkrp/entities/weapons/weapon_cs_base2/sv_commands.lua:43
5. unknown - gamemodes/darkrp/entities/weapons/weapon_cs_base2/sv_commands.lua:69

### DIFF
--- a/entities/entities/spawned_weapon/init.lua
+++ b/entities/entities/spawned_weapon/init.lua
@@ -10,7 +10,9 @@ function ENT:Initialize()
     self:SetSolid(SOLID_VPHYSICS)
     self:SetUseType(SIMPLE_USE)
     local phys = self:GetPhysicsObject()
-    phys:Wake()
+    if IsValid(phys) then
+        phys:Wake()
+    end
 
     if self:Getamount() == 0 then
         self:Setamount(1)


FPtje:
Why is the physobj invalid?

pilot22:
I don't know, but he is xD

FPtje:
Then this hides the problem instead of solving it.

Kefta:
No, this is a correct PR. You should always check physics object validity before using it.

Kefta:
It has a null physics object either because:
1. The limit was hit (unlikely)
2. The server has a custom model that doesn't have a physics model

FPtje:
The physics object of a spawned_weapon must exist. There's a serious problem when it doesn't. That problem should not be ignored.

We've been through this before, the end goal is not to get rid of all errors, the end goal is to help people solve the actual problems they have. Using a non-existent model or one without a physobj in a weapon causes problems. The resulting error is not the root of the problem, the bad model is.

Kefta:
Then add an error when the physobj isn't valid:  NULL physobj error is not descriptive.

FPtje:
Surely that error is already thrown by `DarkRP.ValidatedPhysicsInit`?
How could the PhysObj be invalid after those checks? It would mean `ent:PhysicsInit(solidType)` returns true, indicating having succeeded, while `ent:GetPhysicsObject()` returns `NULL`

Kefta:
I forgot you added that function, you're correct then. I've just tested a number of models with and without physics and I cannot get ``PhysicsInit`` to return true while ``GetPhysicsObject`` returns ``NULL``. It would be nice to know what model the OP is getting that error with.

PR #2973, opened 2020-01-03
Micro-optimizations

FlorianLeChat:
Hi Falco, I know you don't like micro-optimizations but I've been using them for several days on my server without any problem and I think we would gain a bit in performance.

You can also revert [**this**](https://github.com/FPtje/DarkRP/commit/93a4afb7afe953aea88cb4c77d1e2ef9e1e36309) commit because `PLAYER:GetWeapons()` is now fully sequential : https://github.com/Facepunch/garrysmod-issues/issues/3052#issuecomment-520445703 , so we can use `ipairs` instead.

Have a nice day!

### DIFF
--- a/gamemode/modules/fadmin/fadmin/playeractions/god/sv_init.lua
+++ b/gamemode/modules/fadmin/fadmin/playeractions/god/sv_init.lua
@@ -59,10 +59,8 @@ FAdmin.StartHooks["God"] = function()
     FAdmin.Access.AddPrivilege("God", 2)
 end
 
-hook.Add("PlayerSpawn", "FAdmin_God", function()
-    for _, ply in ipairs(player.GetAll()) do
-        if ply:FAdmin_GetGlobal("FAdmin_godded") then
-            ply:GodEnable()
-        end
+hook.Add("PlayerSpawn", "FAdmin_God", function(ply)
+    if ply:FAdmin_GetGlobal("FAdmin_godded") then
+        ply:GodEnable()
     end
 end)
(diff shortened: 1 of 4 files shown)


FPtje:
```lua
function DarkRP.removeFoodItem(i)
    local food = FoodItems[i]
    FoodItems[i] = nil
    hook.Run("onFoodItemRemoved", i, food)
end
```

Any use of the above function in combination with your micro optimisation breaks things. Like I've said earlier, micro-optimisations are a risk. The performance gain is negligible, and the chance of breaking things again proves to be bigger than everyone seems to estimate.

Kefta:
The removal of the player loop in PlayerSpawn should still be done though. Micro optimizations aren't a risk if you fully understand the code.

FPtje:
People seem to be more confident than they should be about how well they understand the code.

FPtje:
That's not to be mean at @FlorianLeChat by the way, Lua code bases can get incredibly subtle like that. You can see it in the amount of mistakes I've made in the last 100 commits, simply because I'm not as deep into it anymore.

FlorianLeChat:
Oh, that's not a problem. I clearly don't have the same knowledge as you or Kefta. When I created this pull request, I was sure I did a good thing with a few _little_ changes, but I realize that it's more complicated than that.

I revert all the changes except for the "PlayerSpawn" hook.

Thanks anyway!

FPtje:
I'm not sure you're right on me having more knowledge, given the mistakes ice been corrected on.

Sorry about my last comment, it came out harsher than I intended it to be.

PR #2978, opened 2020-01-05
Update cl_util.lua

Srlion:
Fixed `total` being wrong in `charWrap` function, it's supposed to include the character in the new line.
~~Optimized `DarkRP.textWrap` and `charWrap` functions by using loops and string concatenation instead of creating closures which can be heavy if used inside hooks like "HudPaint".~~

### DIFF
--- a/gamemode/modules/base/cl_util.lua
+++ b/gamemode/modules/base/cl_util.lua
@@ -23,20 +23,22 @@ Wrap strings to not become wider than the given amount of pixels
 ---------------------------------------------------------------------------]]
 local function charWrap(text, pxWidth)
     local total = 0
+    local newText = ""
 
-    text = text:gsub(".", function(char)
+    for i = 1, #text do
+        local char = text:sub(i, i)
         total = total + surface.GetTextSize(char)
 
-        -- Wrap around when the max width is reached
         if total >= pxWidth then
-            total = 0
-            return "\n" .. char
+            newText = newText .. ("\n" .. char)
+            -- total needs to include the character size
+            total = surface.GetTextSize(char)
+        else
+            newText = newText .. char
         end
+    end
 
-        return char
-    end)
-
-    return text, total
+    return newText, total
 end
 
 function DarkRP.textWrap(text, font, pxWidth)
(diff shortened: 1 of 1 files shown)


FPtje:
Are you sure string concatenation in a loop is faster? That's a quadratic operation.

Also, the textWrap function is a very subtle one, as it has been through quite a few rather subtle bug fixes over the years. I feel like big changes such as this one should be tested against those earlier bugs.

Srlion:
Well, string concatenation can be inefficient too, I'm going to push a better way using tables.

FPtje:
The important part is to profile, don't guess which is faster, measure.

Srlion:
Yours is 4% faster lol, anyway gonna leave the fix for `textWrap`
Here is an image for what it fixes:
a

It's because when you insert the character in next line, you don't include it's width.

Srlion:
Another fix:
a
You pass a width to `charWrap` that is the left width in current line, but you never reset the width to the full width when getting into a new line.

gamemode/modules/base/cl_util.lua
```diff
@@ -21,15 +21,16 @@ usermessage.Hook("blackScreen", blackScreen)
...
     local total = 0
 
     text = text:gsub(".", function(char)
         total = total + surface.GetTextSize(char)
 
         -- Wrap around when the max width is reached
         if total >= pxWidth then
-            total = 0
+            total = surface.GetTextSize(char) -- total needs to include the character width because it's inserted in a new line
+            pxWidth = fullWidth
```
FPtje: This looks off, the `pxWidth` argument of this function is different than the `pxWidth` variable in `DarkRP.textWrap`, which is given as the third argument here. That's confusing. Also, the `pxWidth` argument seems to be used only in the first iteration of this function, only to be replaced by `fullWidth` at the second character. That seems to imply the calculation `pxWidth - (total - wordlen)` given as argument below is disregarded almost always.

Srlion: `fullWidth` does not get replaced at the second character, it gets replaced when we get into a new line. Both of them could be disregarded.

FPtje:
Just tested it with the following code:
```
lua_run_cl text = "a aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

lua_run_cl hook.Add("HUDPaint", "a", function() draw.RoundedBox(0, 0, 0, 200, 200, Color(100, 100, 100)) draw.DrawText(DarkRP.textWrap(text, "DermaDefault", 200), "DermaDefault", 0, 0) end)
```
Before 89578b2f315fe62ce9dcee7076f04b541c0030ab:
hl2_2020-01-05_16-55-26

After 89578b2f315fe62ce9dcee7076f04b541c0030ab:
hl2_2020-01-05_16-56-35

Looks like your change is a clear improvement. I also cannot reproduce the issue mentioned in fd0ac1af1c76f4c49a5c30dda99b42df6645617c

Also, I cannot find any code to reproduce 28f9c62c7f8ca2224af9c31a7bb6b61542996f31, so I'll assume that remains fixed.

Srlion:
#2978 could be easily reproduced:
a

FPtje:
Alright, merged. Great work. You've improved a function that I would be *very* careful to even touch.
I've changed the commit messages to describe the changes, and I've thrown a commit on top to change the variable/parameter names of both functions so it's easier to understand what's going on.

Good changes, good work, awesome.

Issue #2082, opened 2015-08-19
Converting DarkRP


WGHydrogen:
How about converting the DarkRP  usermessages to net message library?

FPtje:
No, that breaks backwards compatibility.

PR #2983, opened 2020-01-15
Convert umsg to net

FoksVHox:
DarkRP used umsg instead of the net library

### DIFF
--- a/entities/entities/darkrp_laws/cl_init.lua
+++ b/entities/entities/darkrp_laws/cl_init.lua
@@ -37,21 +37,21 @@ local function addLaw(inLaw)
     hook.Run("addLaw", lawNumber, inLaw)
 end
 
-local function umAddLaw(um)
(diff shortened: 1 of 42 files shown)


FPtje:
This has been discussed before. We're not doing this because it breaks tons of addons.

FoksVHox:
Examples of addons? And this is pull will secure the future for DarkRP, because who knows when Facepunch is going to remove usermessages.

FPtje:
Every addon that changes how notifications look.

Facepunch isn't going to remove usermessages any time soon, probably never.

See:
https://github.com/FPtje/DarkRP/pull/2860
https://github.com/FPtje/DarkRP/issues/2082
https://github.com/FPtje/DarkRP/pull/1988

TomDotBat:
People just aren't getting the "usermessage", so many have tried but we'll never forget them ;)
kmn

PR #2992, opened 2020-02-25
fixes for the cam library block and more

joeyjumper94:
the way the block that contains the cam library detours was setup prevented the rest of the file from being read by the client,
also some of the blocks could also prevent the rest of the file from being read.

### DIFF
--- a/gamemode/modules/workarounds/sh_workarounds.lua
+++ b/gamemode/modules/workarounds/sh_workarounds.lua
@@ -87,7 +87,7 @@ timer.Simple(3, function()
 end)
 
 if game.SinglePlayer() or GetConVar("sv_lan"):GetBool() then
-    if not DarkRP.disabledDefaults["workarounds"]["nil SteamID64 and AccountID local server fix"] then return
+    if not DarkRP.disabledDefaults["workarounds"]["nil SteamID64 and AccountID local server fix"] then
         local plyMeta = FindMetaTable("Player")
 
         if SERVER then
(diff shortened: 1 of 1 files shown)


Kefta:
The additional cam.* overrides are unnecessary as cam.Start* calls cam.Start internally. The styling is also messed up. Best to leave the cam.* overrides how they were.

joeyjumper94:
i find the lack of protection on cam.End3D2D and cam.End disturbing

FPtje:
Those never crashed the client

joeyjumper94:
read these
https://wiki.facepunch.com/gmod/cam.End
https://wiki.facepunch.com/gmod/cam.End2D
https://wiki.facepunch.com/gmod/cam.End3D
https://wiki.facepunch.com/gmod/cam.End3D2D
on all pages we have this
> This will crash the game if there is no context to end.

gamemode/modules/workarounds/sh_workarounds.lua
```diff
@@ -87,89 +87,93 @@ timer.Simple(3, function()
...
-
-        return cam_Start3D2D(...)
-    end
 
-    return
+if CLIENT and not DarkRP.disabledDefaults["workarounds"]["Cam function descriptive errors"] then
+	local cams3D, cams2D = 0, 0
+	local cam_Start = cam.Start
+	cam.Start=function(tbl, ...)
+		-- https://github.com/Facepunch/garrysmod-issues/issues/3361
```
FPtje: This issue is pretty much marked as resolved. Perhaps we can remove some code here?

gamemode/modules/workarounds/sh_workarounds.lua
```diff
@@ -87,89 +87,93 @@ timer.Simple(3, function()
...
+		else
+			error("bad argument #1 to '%s' (bad key 'type' - 2D or 3D expected, got %s)",debug.getinfo(1,"n").name,tbl.type, 2)
+		end
+		-- Could pcall this but it'd be impossible to
+		-- tell if a render instance was created or not.
+		-- Assume creation/deletion
+		return cam_Start(tbl,...)
+	end
+	local cam_Start3D,cam_End3D = cam.Start3D,cam.End3D
+	cam.Start2D=function(...)
```
FPtje: I think I agree with @Kefta on the overrides of other functions not being necessary. They fall back to `cam.Start`, so you don't need to override them all.

gamemode/modules/workarounds/sh_workarounds.lua
```diff
@@ -87,89 +87,93 @@ timer.Simple(3, function()
...
+if CLIENT and not DarkRP.disabledDefaults["workarounds"]["Cam function descriptive errors"] then
+	local cams3D, cams2D = 0, 0
+	local cam_Start = cam.Start
+	cam.Start=function(tbl, ...)
+		-- https://github.com/Facepunch/garrysmod-issues/issues/3361
+		if (not istable(tbl)) then
+		   argError(tbl, 1, "table")
+		end
+		if tbl.type=="3D" then
+			cams3D=cams3D+1
```
FPtje: The style is inconsistent here.

Kefta:
There was really no need to change the old override code. You can just revert it to the old function body.

joeyjumper94:
so a few more tweaks and it should be working now.

PR #2993, opened 2020-02-27
Save the source of usergroups

### LINKED ISSUE #2991: FAdmin remembers usergroups added to CAMI by other mods.
[//]: # (GO TO THE FORUMS INSTEAD: https://forum.darkrp.com/ )

### Description of the bug
FAdmin registers usergroups that were registered by other CAMI mods, but claims FAdmin as "src".

I'm working on a CAMI-compliant admin mod, and I recently found that if I delete the user groups off-line (i.e. I don't call UnregisterUsergroup()), FAdmin will helpfully remember and re-add them for me, except it will pass "FAdmin" as the "src" parameter instead of whatever "src" I originally added the group with. This is a problem because I depend on "src" to decide whether my admin mod should allow users to change permissions or delete a particular usergroup. The result is "orphaned" usergroups that my mod does not think belong to it, but FAdmin remembered and re-added. I intend to prevent users of my mod from deleting usergroups that were not added through my mod.

Add usergroups with CAMI.RegisterUsergroup() in a mod after FAdmin has been started (maybe before too, haven't tested), shut down the server, remove usergroups from being added by the mod next start up, then start the server again. Hooking CAMI.OnUsergroupRegistered before FAdmin starts will show that the usergroups are still added, and "FAdmin" is passed as the "src" parameter.

### Lua errors

None

FAdmin should only save/load usergroups that have been made through FAdmin. By saving and loading usergroups from other admin mods, you require other admin mods to call UnregisterUsergroup(), which requires other admin mods to keep track of changes that have been made offline and apply them the next time the server is started.

Which, if that's what expected of CAMI mods that's fine, but maybe this issue should be on the CAMI tracker to clarify what's expected of compliant mods. The way the it's worded now, I expect each admin mod to remember it's own custom usergroups, and add them during startup.

Searching through issues I see this may run contrary to #963 , but I'm not concerned if FAdmin wants to **save** the usergroups, I'm only concerned that FAdmin re-**adds** them.

bbblitz:
Create a new table `FADMIN_GROUPS_SRC`.
Save the "source" of usergroups added by other admin mods over CAMI, if they are strings.
During server start, use the original source instead of "FAdmin" when re-adding usergroups.

Fixes #2991

### DIFF
--- a/gamemode/modules/fadmin/fadmin/access/sh_shared.lua
+++ b/gamemode/modules/fadmin/fadmin/access/sh_shared.lua
@@ -13,7 +13,7 @@ function FAdmin.Access.AddGroup(name, admin_access --[[0 = not admin, 1 = admin,
     --Make sure things that come from CAMI come with a CAMIsrc
     assert((fromCAMI and CAMIsrc ~= nil) or ((not fromCAMI) and CAMIsrc == nil))
     --If the CAMIsrc is a string, save it, otherwise save an empty string
-    if type(CAMIsrc) ~= "string" then
+    if not isstring(CAMIsrc) then
         CAMIsrc = ""
     end
 
--- a/gamemode/modules/fadmin/fadmin/access/sh_shared.lua
+++ b/gamemode/modules/fadmin/fadmin/access/sh_shared.lua
@@ -7,8 +7,15 @@ FAdmin.Access.ADMIN[0] = "user"
 FAdmin.Access.Groups = FAdmin.Access.Groups or {}
 FAdmin.Access.Privileges = FAdmin.Access.Privileges or {}
 
-function FAdmin.Access.AddGroup(name, admin_access --[[0 = not admin, 1 = admin, 2 = superadmin]], privs, immunity, fromCAMI)
+function FAdmin.Access.AddGroup(name, admin_access --[[0 = not admin, 1 = admin, 2 = superadmin]], privs, immunity, fromCAMI, CAMIsrc)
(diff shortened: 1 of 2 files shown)


gamemode/modules/fadmin/fadmin/access/sh_shared.lua
```diff
@@ -7,9 +7,16 @@ FAdmin.Access.ADMIN[0] = "user"
...
 FAdmin.Access.Privileges = FAdmin.Access.Privileges or {}
 
-function FAdmin.Access.AddGroup(name, admin_access --[[0 = not admin, 1 = admin, 2 = superadmin]], privs, immunity, fromCAMI)
+function FAdmin.Access.AddGroup(name, admin_access --[[0 = not admin, 1 = admin, 2 = superadmin]], privs, immunity, fromCAMI, CAMIsrc)
     FAdmin.Access.Groups[name] = FAdmin.Access.Groups[name] or {ADMIN = admin_access, PRIVS = privs or {}, immunity = immunity}
 
+    --Make sure things that come from CAMI come with a CAMIsrc
+    assert((fromCAMI and CAMIsrc ~= nil) or ((not fromCAMI) and CAMIsrc == nil))
+    --If the CAMIsrc is a string, save it, otherwise save an empty string
+    if type(CAMIsrc) ~= "string" then
```
FPtje: You can use the `isstring` function here.

bbblitz: Done!

PR #2997, opened 2020-03-08
Minor optimisations to DarkRP code

mcNuggets1:
* Cached localplayer in a more useful way for cl_hud.
* Cached entity in GetEyeTrace call instead of the GetEyeTrace result table.
* Simple table indexing for doors in cl_doors.lua and for HUDCanDraw in cl_hud.lua
* Removal of IsValid checks in cl_doors, won't work on non entities anyways and shouldn't hide the incorrect usage of a function.
* Next Update change: Used GetDriver() instead of going through player.GetAll() to determine vehicle driver. (This only works after the next gmod update is pushed, the next GMod-update will release 18th of March)

### DIFF
--- a/gamemode/modules/base/sh_gamemode_functions.lua
+++ b/gamemode/modules/base/sh_gamemode_functions.lua
@@ -8,7 +8,7 @@ end
 function GM:StartCommand(ply, usrcmd)
     -- Used in arrest_stick and unarrest_stick but addons can use it too!
     local wep = ply:GetActiveWeapon()
-    if IsValid(wep) and isfunction(wep.startDarkRPCommand) then
+    if wep:IsValid() and isfunction(wep.startDarkRPCommand) then
         wep:startDarkRPCommand(usrcmd)
     end
 end
--- a/gamemode/modules/doorsystem/sh_doors.lua
+++ b/gamemode/modules/doorsystem/sh_doors.lua
@@ -1,28 +1,36 @@
 local meta = FindMetaTable("Entity")
 local plyMeta = FindMetaTable("Player")
 
+local ownableDoors = {
+    ["func_door"] = true,
+    ["func_door_rotating"] = true,
+    ["prop_door_rotating"] = true
+}
+local unOwnableDoors = {
+    ["func_door"] = true,
+    ["func_door_rotating"] = true,
+    ["prop_door_rotating"] = true,
+    ["func_movelinear"] = true,
+    ["prop_dynamic"] = true
+}
 function meta:isKeysOwnable()
-    if not IsValid(self) then return false end
     local class = self:GetClass()
 
-    if ((class == "func_door" or class == "func_door_rotating" or class == "prop_door_rotating") or
+    if (ownableDoors[class] or
             (GAMEMODE.Config.allowvehicleowning and self:IsVehicle() and (not IsValid(self:GetParent()) or not self:GetParent():IsVehicle()))) then
-            return true
-        end
+        return true
+    end
+
     return false
 end
 
 function meta:isDoor()
-    if not IsValid(self) then return false end
     local class = self:GetClass()
 
-    if class == "func_door" or
-        class == "func_door_rotating" or
-        class == "prop_door_rotating" or
-        class == "func_movelinear" or
-        class == "prop_dynamic" then
+    if unOwnableDoors[class] then
         return true
     end
+
     return false
 end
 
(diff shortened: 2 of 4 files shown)


gamemode/modules/base/sh_gamemode_functions.lua
```diff
@@ -8,7 +8,7 @@ end
 function GM:StartCommand(ply, usrcmd)
     -- Used in arrest_stick and unarrest_stick but addons can use it too!
     local wep = ply:GetActiveWeapon()
-    if IsValid(wep) and isfunction(wep.startDarkRPCommand) then
+    if wep:IsValid() and isfunction(wep.startDarkRPCommand) then
```
FPtje: Even though `ply:GetActiveWeapon()` returns `NULL` when the player has no weapons, I prefer `IsValid` in principle, because that allows me to ignore the difference between `nil` and `NULL`. 

Usage of `ent:IsValid()` raises the question *"Oh, but can this thing be `nil`?"*, and I have to check the wiki and/or the implementation of the function to see if this is the case. I argue that readability here is more important than the unnoticeable performance benefit.

gamemode/modules/doorsystem/sh_doors.lua
```diff
@@ -1,28 +1,36 @@
 local meta = FindMetaTable("Entity")
 local plyMeta = FindMetaTable("Player")
 
+local doorsOwnable = {
```
FPtje: Perhaps a minor technical detail. This code is easier to read, but might not necessarily be faster. Table lookups with string involve some hashing function. Such hashing function might be ever slightly more expensive for small tables than a linear if statement checking all entries.

Regardless, this change is good, as it improves readability, and is pretty much as fast as the other implementation. Good work 👍

gamemode/modules/base/sh_gamemode_functions.lua
mcNuggets1: We already changed it in a lot f places.
I just want it to be changed in hooks being called very often.
The IsValid check in general here isn't needed I think.

gamemode/modules/doorsystem/sh_doors.lua
mcNuggets1: If I may ask, why are there unownable doors which cannot really be interacted with?

FPtje: The answer to that question lies many years back. I believe there were maps with those entity types which for some reason should be treated as doors. I don't know which maps anymore, and tbh I don't know what, if anything, will break if this is changed.

gamemode/modules/doorsystem/sh_doors.lua
```diff
@@ -1,28 +1,36 @@
...
+    ["func_door"] = true,
+    ["func_door_rotating"] = true,
+    ["prop_door_rotating"] = true
+}
+local doorsUnownable = {
+    ["func_door"] = true,
+    ["func_door_rotating"] = true,
+    ["prop_door_rotating"] = true,
+    ["func_movelinear"] = true,
+    ["func_movelinear"] = true
```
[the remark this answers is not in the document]
mcNuggets1: Oh shit, good catch.
What a idiot I am sometimes.

gamemode/modules/base/sh_gamemode_functions.lua
Kefta: The GMod API has slowly been moving toward NULL consistency - there are only a few small edge cases left (NPC:GetEnemy and GetTarget specifically) that are waiting to be resolved. Differentiating between nil and NULL is now an almost completely safe paradigm that is well-defined and yields constant time optimisation.

gamemode/modules/doorsystem/sh_doors.lua
```diff
@@ -1,28 +1,36 @@
 local meta = FindMetaTable("Entity")
 local plyMeta = FindMetaTable("Player")
 
+local doorsOwnable = {
+    ["func_door"] = true,
+    ["func_door_rotating"] = true,
+    ["prop_door_rotating"] = true
+}
+local doorsUnownable = {
```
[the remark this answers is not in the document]
mcNuggets1: ownableDoors should do, even if I don't really think, a local table in a file somewhere is gonna confuse someone, when the actual use of it, is 10 lines below.

FlorianLeChat:
[The update is out!](https://gmod.facepunch.com/blog/march-2020-update)

PR #3001, opened 2020-03-24
Add isvalid checks to more loops in doorsystem

mcNuggets1:
Errors otherwise in some occasions.

### DIFF
--- a/gamemode/modules/doorsystem/sv_doors.lua
+++ b/gamemode/modules/doorsystem/sv_doors.lua
@@ -107,7 +107,7 @@ end
 
 function pmeta:keysUnOwnAll()
     for entIndex, ent in pairs(self.Ownedz or {}) do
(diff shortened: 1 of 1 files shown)


Bo98:
I was about to ask why this was only needed now, but I answered my own question: it's because the behaviour of `isKeysOwnable` was changed recently.

mcNuggets1:
Most of the loops already had isvalid checks, but some of them were missing.

FPtje:
Fixes #3004

FPtje:
Great and all, but removing the isvalid check from `isKeysOwnable` caused errors. I'm sure other addons use it as well, and like DarkRP's own code, may depend on that check being there.

It should not have been removed, and I'm re-adding it. Invalid entities are not ownable.

mcNuggets1:
Maybe remove the isvalid checks from the remaining loops then?

PR #3003, opened 2020-03-26
Update sv_doors.lua

BluetonicBeats:
Stoping vehicle doors from being sold on job-change or unownalldoors as it will cost the player to rebuy the doors or he/her could lose the ownership of his/her car.


FPtje:
This function is not called when the job is changed. Also, there's other code in that function that takes vehicles into account. That doesn't make a whole lot of sense.

Besides, you're setting `ply.Ownedz[entIndex] = nil` for every vehicle a player owns when they run that command, even though you don't want them to unown the vehicle. That cannot be good.

PR #3014, opened 2020-05-07
Update cl_init.lua

MoskalykA:


FPtje:
This can't be right. What are you doing?

TomDotBat:
oPtImIsAtIoN BrO WhAt Do YoU ThInK?

MoskalykA:
I'm thinking of optimization.

FPtje:
Not drawing things makes things faster in the same way not running a gamemode does. Sure, faster, but you kinda lose something along the line.

PR #3017, opened 2020-05-12
Remove PreDrawViewModel IsValid check

Kefta:
To merge when the next update comes out. PreDrawViewModel will now only be called with valid viewmodels: https://github.com/Facepunch/garrysmod-issues/issues/3024


FPtje:
When will that be released? I don't keep track of gmod updates anymore.

Kefta:
Still unknown, they are only announced two weeks in advance. I will let you know when this happens.

PR #3033, opened 2020-07-23
Fix hardcoded check in door info HUD

FlorianLeChat:
By ignoring CAMI permissions, the DarkRP doesn't display an information on the door HUD.
Currently, only superadmins can see the text [keys_disallow_ownership](https://github.com/FPtje/DarkRP/blob/4fd2c3c315427e79bb7624702cfaefe9ad26ac7e/gamemode/modules/language/sh_english.lua#L150). Now, by checking CAMI's `DarkRP_ChangeDoorSettings` permission, other user groups will be able to see the text.

### DIFF
--- a/gamemode/modules/doorsystem/cl_doors.lua
+++ b/gamemode/modules/doorsystem/cl_doors.lua
@@ -12,7 +12,7 @@ function meta:drawOwnableInfo()
     if doorDrawing == true then return end
 
     local blocked = self:getKeysNonOwnable()
-    local superadmin = ply:IsSuperAdmin()
+    local access = CAMI.PlayerHasAccess(ply, "DarkRP_ChangeDoorSettings")
     local doorTeams = self:getKeysDoorTeams()
     local doorGroup = self:getKeysDoorGroup()
     local playerOwned = self:isKeysOwned() or table.GetFirstValue(self:getKeysCoOwners() or {}) ~= nil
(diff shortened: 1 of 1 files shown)


FPtje:
It's not really safe to assume that `CAMI.PlayerHasAccess` returns immediately.

FlorianLeChat:
I have added the superadmin check again in case CAMI doesn't return a response immediately. But during my tests, CAMI always returned an immediate answer.

FPtje:
That's because it depends on the implementation of the admin mod. That behavior will never change if you don't change admin mod.

gamemode/modules/doorsystem/cl_doors.lua
```diff
@@ -12,7 +12,7 @@ function meta:drawOwnableInfo()
     if doorDrawing == true then return end
 
     local blocked = self:getKeysNonOwnable()
-    local superadmin = ply:IsSuperAdmin()
+    local access = CAMI.PlayerHasAccess(ply, "DarkRP_ChangeDoorSettings") or ply:IsSuperAdmin()
```
Heyter, line 15: ply:IsSuperAdmin() or CAMI.PlayerHasAccess(ply, "DarkRP_ChangeDoorSettings")

Kefta: That makes no sense.

FPtje:
I mean the point is that you should use the callback. Also, some privilege calculation functions from some admin mods can be expensive. Calling this in `drawOwnableInfo` might form a serious bottleneck, depending on the admin mod.

FlorianLeChat:
I spent literally two hours looking for a good way to do what I wanted to do, but by searching a little bit in the gamemode files, I found a function in the FAdmin that is similar to the one in CAMI and that gives an immediate result.

FlorianLeChat:
I think using the `DarkRPVarChanged` hook is a good compromise to update privileges. Since I put a check if the player of the hook is LocalPlayer to avoid performance issues, the update is not so "fast".

If you have a better idea, I'll take it. 😢

FPtje:
Fuck it, timers it is.

PR #3036, opened 2020-08-06
Added darkRPPreGiveWeapon, darkRPPostGiveWeapon

DBotThePony:
This will allow other addons to modify weapon given to players.

Example usage: Other addon spawn `spawned_weapon`, giving it it's special meta, which then will apply to newly given weapon to player.

### DIFF
--- a/entities/entities/spawned_weapon/init.lua
+++ b/entities/entities/spawned_weapon/init.lua
@@ -66,6 +66,9 @@ function ENT:Use(activator, caller)
 
     weapon = activator:Give(class, true)
 
+    hook.Call("darkRPPreGiveWeapon", nil, activator, self, weapon)
+    if not IsValid(weapon) then return end
+
     local clip1, clip2 = self.clip1, self.clip2
     if weapon:IsValid() then
         if clip1 and clip1 ~= -1 and weapon:Clip1() ~= -1 then
@@ -94,6 +97,8 @@ function ENT:Use(activator, caller)
         local secAmmo = activator:GetAmmoCount(secondaryAmmoType) + (clip2 or 0)
         activator:SetAmmo(secAmmo, secondaryAmmoType)
     end
+
+    hook.Call("darkRPPostGiveWeapon", nil, activator, self, weapon)
 
     self:DecreaseAmount()
 end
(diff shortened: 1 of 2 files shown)


DBotThePony:
Also take a look at https://github.com/Facepunch/garrysmod-issues/issues/4589 and decide whenever do you still need such logic in spawned_weapon.

entities/entities/spawned_weapon/init.lua
```diff
@@ -66,6 +66,9 @@ function ENT:Use(activator, caller)
 
     weapon = activator:Give(class, true)
 
+    hook.Call("darkRPPreGiveWeapon", nil, activator, self, weapon)
```
FPtje, line 69: It's not really "Pre" GiveWeapon if you call it right after the :Give call

FPtje:
Also the naming of "GiveWeapon" give the very strong impression that it's global to any weapon given to a player, rather than being very specific to `spawned_weapon`. Won't it be better to create an overridable `ENT:PreGiveWeapon` (which is actually called *before* the give), and `ENT:PostGiveWeapon` that addon authors can override?

Can't addon authors just override `ENT:Use` to get the functionality of at least `ENT:PostGiveWeapon`? They can also define their own clip/ammo behavior if the override `ENT:Use` without calling the base class.

I'm not entirely convinced of the point here.

DBotThePony:
I added these hooks to not to do something like this
Code_ANX5YPYpBj

And i call it `pre` because it is executed before *DarkRP*'s logic at giving weapon, not actually giving. You see, there is already `PlayerPickupDarkRPWeapon` for that.

FPtje:
> i call it pre because it is executed before DarkRP's logic

That is just not going to fly. Also I really don't see the need for two hooks. Here's an alternative solution: you override `ENT:GivePlayerAmmo(ply, weapon, playerHadWeapon)`. Either when spawning the `spawned_weapon` or in your own custom spawned_weapon derivative which you spawn instead.

DBotThePony:
Anyway, this still won't make sense for other mods. I override Give because i need to set specific metadata on that weapon, for example, setting attachments. DarkRP `spawned_weapon`'s implementation is going to lose them anyway when player drop weapon with these attachments for example. You still need to add these global hooks for other mods to hook onto.

But following this addition of hooks you are probably right, and my PR could add more hooks for purposes such as storing metadata of weapon player dropped into `spawned_weapon`. But either way, both my and your solutions are incomplete.

PR #3037, opened 2020-08-09
Optimise talking

mcNuggets1:
Stopped using direct index calls on the player and started using table indexes.
Should save a lot of performance.

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -329,6 +329,7 @@ hook.Add("PlayerDisconnected", "DarkRPCanHearVoice", function(ply)
     DrpCanHear[ply] = nil
     if ply:IsBot() then return end
     for _, v in ipairs(player.GetHumans()) do
+        DrpCanHear[v] = DrpCanHear[v] or {}
         DrpCanHear[v][ply] = nil
     end
     timer.Remove(ply:UserID() .. "DarkRPCanHearPlayersVoice")
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -304,27 +304,28 @@ local vrad = GM.Config.voiceradius
 local dynv = GM.Config.dynamicvoice
 local deadv = GM.Config.deadvoice
 local voiceDistance = GM.Config.voiceDistance * GM.Config.voiceDistance
+local DrpCanHear = {}
 -- proxy function to take load from PlayerCanHearPlayersVoice, which is called a quadratic amount of times per tick,
 -- causing a lagfest when there are many players
 local function calcPlyCanHearPlayerVoice(listener)
-    if not IsValid(listener) then return end
-    listener.DrpCanHear = listener.DrpCanHear or {}
+    DrpCanHear[listener] = DrpCanHear[listener] or {}
     local shootPos = listener:GetShootPos()
     for _, talker in ipairs(player.GetAll()) do
         local talkerShootPos = talker:GetShootPos()
-        listener.DrpCanHear[talker] = not vrad or -- Voiceradius is off, everyone can hear everyone
+        DrpCanHear[listener][talker] = not vrad or -- Voiceradius is off, everyone can hear everyone
             (shootPos:DistToSqr(talkerShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                 (not dynv or IsInRoom(shootPos, talkerShootPos, talker))) -- Dynamic voice is on and players are in the same room
     end
 end
 hook.Add("PlayerInitialSpawn", "DarkRPCanHearVoice", function(ply)
+    calcPlyCanHearPlayerVoice(ply)
     timer.Create(ply:UserID() .. "DarkRPCanHearPlayersVoice", 0.5, 0, fn.Curry(calcPlyCanHearPlayerVoice, 2)(ply))
 end)
+
 hook.Add("PlayerDisconnected", "DarkRPCanHearVoice", function(ply)
-    if not ply.DrpCanHear then return end
     for _, v in ipairs(player.GetAll()) do
-        if not v.DrpCanHear then continue end
-        v.DrpCanHear[ply] = nil
+        if not DrpCanHear[v] then continue end
+        DrpCanHear[v] = nil
     end
     timer.Remove(ply:UserID() .. "DarkRPCanHearPlayersVoice")
 end)
@@ -332,8 +333,7 @@ end)
 function GM:PlayerCanHearPlayersVoice(listener, talker)
     if not deadv and not talker:Alive() then return false end
 
-    local canHear = listener.DrpCanHear and listener.DrpCanHear[talker]
-    return canHear, threed
+    return DrpCanHear[listener][talker], threed
 end
 
 function GM:CanTool(ply, trace, mode)
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -326,6 +326,7 @@ hook.Add("PlayerDisconnected", "DarkRPCanHearVoice", function(ply)
     for _, v in ipairs(player.GetAll()) do
         DrpCanHear[v][ply] = nil
     end
+    DrpCanHear[ply] = nil
     timer.Remove(ply:UserID() .. "DarkRPCanHearPlayersVoice")
 end)
 
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -309,8 +309,9 @@ local DrpCanHear = {}
 -- causing a lagfest when there are many players
 local function calcPlyCanHearPlayerVoice(listener)
     DrpCanHear[listener] = DrpCanHear[listener] or {}
+    if listener:IsBot() then return end
     local shootPos = listener:GetShootPos()
-    for _, talker in ipairs(player.GetAll()) do
+    for _, talker in ipairs(player.GetHumans()) do
         local talkerShootPos = talker:GetShootPos()
         DrpCanHear[listener][talker] = not vrad or -- Voiceradius is off, everyone can hear everyone
             (shootPos:DistToSqr(talkerShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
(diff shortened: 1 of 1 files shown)


FPtje:
Call me paranoid, but I have a suspicion that addons are using that member variable. How much faster is using tbl index versus a  player index anyway?

mcNuggets1:
A ton.
I dont know an addon which uses this variable.
However this increases Performance by a lot.

I can benchmark

mcNuggets1:
] lua_openscript performance.lua
Running script performance.lua...
darkrp versiontook: 1.97
mcnuggets versiontook: 0.0020000000000095

`function check(func, name)
	local check_ply = Entity(1)
	check_ply.DarkRP_Muted = {}
	local timee = os.clock()
	for i=1, 10000000 do
		func(check_ply, check_ply)
	end
	print(name.."took: "..os.clock() -  timee)
end

local function performance(talk, lis)
	if talk.DarkRP_Muted and talk.DarkRP_Muted[talk] then
		return false
	end
end
check(performance, "darkrp version")

local DarkRP_Muted = {}
local function performance2(talk, lis)
	if DarkRP_Muted[talk] then
		return false
	end
end
check(performance2, "mcnuggets version")`

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -304,36 +304,36 @@ local vrad = GM.Config.voiceradius
...
                 (not dynv or IsInRoom(shootPos, talkerShootPos, talker))) -- Dynamic voice is on and players are in the same room
     end
 end
 hook.Add("PlayerInitialSpawn", "DarkRPCanHearVoice", function(ply)
+    calcPlyCanHearPlayerVoice(ply)
     timer.Create(ply:UserID() .. "DarkRPCanHearPlayersVoice", 0.5, 0, fn.Curry(calcPlyCanHearPlayerVoice, 2)(ply))
 end)
+
 hook.Add("PlayerDisconnected", "DarkRPCanHearVoice", function(ply)
-    if not ply.DrpCanHear then return end
```
FPtje, line 324: `DrpCanHear[ply] = nil`?

mcNuggets1: It cannot be nil now.
It simply cannot.

FPtje: That's not what I meant. I meant you should SET `DrpCanHear[ply] = nil`. You're setting the table for all other players, but not for the leaving player.

Say you have `ply1`, `ply2`, `ply3`. At some point the table looks like this:

```lua
{ [ply1] = {[ply2] = true, [ply3] = true}
, [ply2] = {[ply1] = true, [ply3] = true}
, [ply3] = {[ply1] = true, [ply2] = true}
}
```

When `ply1` leaves, you want `DrpCanHear[ply1]` to be nil, besides making `DrpCanHear[ply2][ply1]` and `DrpCanHear[ply3][ply1]` nil. Then you should have a table with two members, each holding a table of one value:

```lua
{ [ply2] = {[ply3] = true}
, [ply3] = {[ply2] = true}
}
```

I'd like you to run some tests to make sure this works correctly.

Also, are you 100% sure that the `PlayerInitialSpawn` is called before `player.GetAll()` can hold the player anywhere? I wouldn't be so sure. GMod has whacky stuff like that.

DBotThePony: PlayerInitialSpawn is called after PlayerAuthed. It should be guaranteed player.GetAll() will contain them.

mcNuggets1:
I currently run this on my servers for testing purposes.

gamemode/modules/base/sv_gamemode_functions.lua
FPtje: > PlayerInitialSpawn is called after PlayerAuthed. It should be guaranteed player.GetAll() will contain them.

`PlayerInitialSpawn` being called *after* things is even more worrying. I worry about this happening:

- Ply1 joins
- Ply2 disconnects
- `PlayerDisconnected` gets called for `ply2`
- `Ply1` is in `player.GetAll()`, but `DrpCanHear[ply1]` is `nil`, so you get a nil error
- `PlayerInitialSpawn` gets called for `ply1`, which sets `DrpCanHear[ply1]`

Don't prove to me `PlayerInitialSpawn` is called AFTER something. Prove to me that is definitely called BEFORE the `PlayerDisconnected` of another player.

mcNuggets1: InitialSpawn is always called before Disconnected, I tested.
With a bot.

FPtje: This is a very timing sensitive thing. Something that may take a year before someone reports the error. How did you test this?

mcNuggets1: Spawning the bot.
Next frame, kick them.
It's literally a frame away.

mcNuggets1:
after running it for a day, it made a big difference performance wise.
32 players with all props and vehicles and 50% CPU on a fully modded server, was never this low in history.

FPtje:
That's the kind of performance improvement I value seeing.

FlorianLeChat:
Small suggestion: you think there is a way to put a check for if `ply` is a bot at the `PlayerInitialSpawn` and `PlayerDisconnected` hook (same thing for the local function `calcPlyCanHearPlayerVoice` to use `player.GetHumans` instead of `player.GetAll`), this would prevent them from creating timers for nothing or making extra calculations.

mcNuggets1:
player.GetHumans is slower and who the fuck adds bots to their live darkrp servers?

Kefta:
I disagree, GetHumans is hardly slower since the loop checks are all in C, and it would yield massive benefits for anyone who has a lot of bots, even in the rare case. I think the tradeoff is fine.

FlorianLeChat:
@mcNuggets1 Some WW1/WW2 servers I know (mostly French) sometimes use bots to simulate waves of enemies 😄

gamemode/modules/base/sv_gamemode_functions.lua
FPtje: I'm still having a bad stomach feeling about this. Years of people reporting the most insane errors made me paranoid. 

Let me as the @Kefta guru to be sure: I don't develop for gmod nearly enough to be sure about subtle foot guns like this. Do you think there's a risk of the described error?

mcNuggets1:
They should use NPCs, not simulated players for these purposes.
I guess, if my god Kefta tells me, it's fine, then I suppose, it's fine.
I still strongly disagree, especially because the timer is running so fucking often.

Kefta:
Though I am a vehement supporter of "micro" optimisations (better known as good habits), this one actually could yield benefits for servers that do dumb things with bots.

mcNuggets1:
The PlayerCanHearPlayersVoice hook still runs for bots and I need to keep it performing well.
So my implementation still creates the table for the bot.

gamemode/modules/base/sv_gamemode_functions.lua
mcNuggets1: Even if there is.
It corrects it within 0.5 seconds.
There is no risk here.
Even if it errors, we can quickly make a change within seconds.

FPtje: Yeah, you're going to have a hard time explaining that to the server owner who doesn't have a clue on how bad the error is or what it even means. The risk is server owners being stressed out about errors that we are anticipating right here. When they see the error we've already lost, regardless of whether the error has technical consequences.

Also, such errors degrade the perceived quality of DarkRP. It shouldn't throw errors for issues not caused by the server owner.

mcNuggets1: I am running it on 3 servers without any issues.
If it errors, I see it first.
An error which does nothing, does no harm, because it's easily fixed.

We just had an error caused by an update and it did no harm at all, because I reported it within a day.

FPtje: No, you're explicitly **not** the first to see it. This error would be extremely rare. It's the kind that gets reported YEARS after being introduced. Thinking that you will be the one to find it, and soon even, is extremely naive.

The check isn't even in a hot code path. It's in the PlayerDisconnected hook.

mcNuggets1: I'm running it on 3 servers, simultaneously, pre launch.
I'm a developer which is frequently checking their console.

If the owner of the server doesn't fuck up PlayerDisconnected, there will be no error.
I've even tested further, with kicking on PlayerAuthed, which then resulted in being still valid in PlayerInitialSpawn and calling PlayerDisconnected properly.

The only edge case scenario which even exists is, if the player errors on PlayerDisconnected, because then, the timer wouldn't discontinue, when the player disconnects.

Additionally PlayerAuthed isn't called for bots, AT ALL, which is essencially.

HERE:
https://pastebin.com/sapQAwAw -- Tested with a real player AND a bot.
Both didn't error.

Code:
https://pastebin.com/r9YNvSUP

mcNuggets1:
So now even the most stupid creature in the world has no ability to fuck this up.
It's impossible. It costed an IsValid check, but these are lightweight anyways.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -304,36 +304,40 @@ local vrad = GM.Config.voiceradius
...
 hook.Add("PlayerInitialSpawn", "DarkRPCanHearVoice", function(ply)
+    calcPlyCanHearPlayerVoice(ply)
+    if ply:IsBot() then return end
     timer.Create(ply:UserID() .. "DarkRPCanHearPlayersVoice", 0.5, 0, fn.Curry(calcPlyCanHearPlayerVoice, 2)(ply))
 end)
+
 hook.Add("PlayerDisconnected", "DarkRPCanHearVoice", function(ply)
-    if not ply.DrpCanHear then return end
-    for _, v in ipairs(player.GetAll()) do
-        if not v.DrpCanHear then continue end
```
FPtje, line 326: It was this check I was talking about, though, not the one in `calcPlyCanHearPlayerVoice`. Sorry if I didn't make that clear earlier.

mcNuggets1: In order to fuck this one up, PlayerInitialSpawn HAS to error and then only the part with removing OTHER players from your table can cause errors.

mcNuggets1: The rest of the thing, physically can't error. It just can't now.

mcNuggets1:
I really believed, you meant the timer error, because it would spam and hinder players from actually finding their own lua error, which caused the discontinuation of functionality of the voice system.

FPtje:
Awesome. Thanks!

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -308,6 +308,7 @@ local DrpCanHear = {}
 -- proxy function to take load from PlayerCanHearPlayersVoice, which is called a quadratic amount of times per tick,
 -- causing a lagfest when there are many players
 local function calcPlyCanHearPlayerVoice(listener)
+    if not IsValid(listener) then return end
     DrpCanHear[listener] = DrpCanHear[listener] or {}
     if listener:IsBot() then return end
```
FoksVHox: Why not merge the to if statements?

mcNuggets1: Because bots need to be added to the list, as I stated above.
They are considered by PlayerCanHearPlayersVoice.

FoksVHox: Oh

FPtje:
Thanks for the contribution! This is really worth something.

PR #3039, opened 2020-08-11
Removed pointless think function

Down-s:


NoSharp:
legitimately no point in this.

TomDotBat:
^

mcNuggets1:
You basically make it worse.
base_gmodentity, the entity this is derived from, runs a Think hook, which, when not overwritten, will calculate stuff.

Changing from base_gmodentity to base_anim would be helpful in performance, but this, is, definitely, not.

Kefta:
Well if this stays as a base_gmodentity, the Think function shouldn't be overwritten, but if it were switched to base_anim, this could be removed. I don't know if any of base_gmodentity's features are necessary here, perhaps @FPtje could provide insight.

FPtje:
Both the Think override and the `base_gmodentity` base come from the [second commit](https://github.com/FPtje/DarkRP/commit/e94a38eac56dc1df40d46b4ba2d073c97e428cdd) in this repo, which uploads all the DarkRP files. Now I don't know what happened between 2008 and 2012, but I can take a wild guess: literally nothing.

I can't provide more insight without looking into the differences between `base_anim` and `base_gmodentity` myself.

Kefta:
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua

base_gmodentity is a base_anim with a few extra functions. Are any of these needed?

FPtje:
I don't know, but even if they aren't for the DarkRP printer itself, they might be used by entities that derive DarkRP's printer. I'd rather not touch it.

PR #3040, opened 2020-08-13
Localise self:GetOwner()

Down-s:
self:GetOwner() is used a damn lot of times in this SWEP, localising it would be a good idea

### DIFF
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -59,6 +59,8 @@ function SWEP:PrimaryAttack()
     if CLIENT then return end
     
     local Owner = self:GetOwner()
+    
+    if not IsValid(Owner) then return end
 
     Owner:LagCompensation(true)
     local trace = util.QuickTrace(Owner:EyePos(), Owner:GetAimVector() * 90, {Owner})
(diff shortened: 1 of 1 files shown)


FlorianLeChat:
I've spotted the same thing in almost every other weapons (`unarrest_stick`, `keys`, `door_ram`...).

Kefta:
You should also check if the owner is valid before using it in any function.

Bo98:
> You should also check if the owner is valid before using it in any function.

In what scenario is a weapon in the process of a priamry attack but does not have a valid owner? I've not seen any owner-related Lua errors reported here.

Kefta:
I thought I remember there being a condition where it was possible to happen right after the owner dropped a weapon, but I can't seem to find it now. The only feasible way it'd be invalid is if it were called manually from another function, such as for Lua implementations of burst fire. In general it's a good idea to always check entity return validity in every SWEP function call to prevent it from erroring when being manually called, but I suppose it doesn't matter here unless I could find that codepath I was originally thinking of.

Bo98:
> it was possible to happen right after the owner dropped a weapon

Possibly something to do with `Holster`? I seem to recall something to do with that but that was many many years ago.

Kefta:
It would occur after Holster, but the owner being invalid in Holster is a whole other can of worms with some duplicitous ``Holster`` call Garry put in for no reason.

entities/weapons/arrest_stick/shared.lua
```diff
@@ -57,43 +57,50 @@ function SWEP:PrimaryAttack()
     BaseClass.PrimaryAttack(self)
 
     if CLIENT then return end
+    
+    local Owner = self:GetOwner()
+
```
FPtje, line 62: Please turn "trim trailing whitespace" on in your editor next time

PR #3041, opened 2020-08-15
More self:GetOwner() localizing

FlorianLeChat:
As done [here](https://github.com/FPtje/DarkRP/commit/a7d7fc31afb221705f3b4a4b7019025c1c21a6bb), other localizations could be made in almost all the weapons included in the DarkRP.
Some areas have been intentionally omitted because I think that `self:GetOwner` is not called enough many times to be localized (<2/3 times) and that they are not present in critical areas of the code.

### DIFF
--- a/entities/weapons/door_ram/shared.lua
+++ b/entities/weapons/door_ram/shared.lua
@@ -201,16 +201,20 @@ Desc: +attack1 has been pressed
 function SWEP:PrimaryAttack()
     if not self:GetIronsights() then return end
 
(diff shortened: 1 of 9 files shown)


Down-s:
Great work 👍

FoksVHox:
All of this could be done in one commit

FlorianLeChat:
Yep sorry, usually I use Github Desktop for big changes but this time I used the web version directly.

PR #3042, opened 2020-08-15
Use ipairs instead of pairs in PLAYER:GetWeapons()

FlorianLeChat:
Since [September 2019 update](https://gmod.facepunch.com/blog/september-2019-update) (**see below**), `PLAYER:GetWeapons` returns always a sequential table, so we can use without any problem `ipairs` instead of `pairs` for better performance.

image

**Note**: This is a revert of [this](https://github.com/FPtje/DarkRP/commit/93a4afb7afe953aea88cb4c77d1e2ef9e1e36309) commit.


FPtje:
Thanks for your contributions!

PR #3043, opened 2020-08-16
Unclamp the Respawn Time

Down-s:
Not sure why this is clamped, would make sense to let server owners freely change the respawn time to whatever they want.

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -449,7 +449,7 @@ function GM:PlayerDeath(ply, weapon, killer)
         DarkRP.notify(ply, 4, 4, DarkRP.getPhrase("dead_in_jail"))
     else
         -- Normal death, respawning.
-        ply.NextSpawnTime = CurTime() + math.Clamp(GAMEMODE.Config.respawntime, 0, 10)
+        ply.NextSpawnTime = CurTime() + GAMEMODE.Config.respawntime
     end
     ply.DeathPos = ply:GetPos()
 


FPtje:
Political reasons. Servers were actually doimg this, and leaving the respawn times low for admins, VIPs and whatnot. This happened a lot actually, sometimes up to 90 seconds. DarkRP is a game where you die a lot. Waiting more than 10 seconds every time is plain bullying. 

People were actually abusing those long spawn times, shooting people and then taunting them as they wait to respawn.

Kefta:
So the only time to get longer respawn times would be to override GM:PlayerDeath or modify core files, instead of having a hook or config option to disable this behaviour? Good shit.

FPtje:
Hey, this was a really big fucking problem. Server owners and admins didn't even notice the clamp for a long time because they were pissed off by their own respawn times in the first place and disabled it for themselves.

Can someone explain to me what purpose is served by effectively banning people from the game every time they die? Dying happens really often. People getting angry, calling admins about losing shit when they die is already a problem. Having them wait ages will piss them off even more, and rightly so. When any kind of conflict is going on, you're spending more time waiting to respawn than actually playing the damn game.

The hipocrisy of admins and VIPs not having these respawn times is what pissed me off the most. They know full damn well how God awful their choice is. Instead of reverting their shitty decision they went ahead and wrote code to exempt themselves. That's messed up.

Kefta:
The "effectively banning people" thing with high respawn times is very annoying, yet was an integral part of PERP that sucked but people collectively dealt with it - it made death MEAN something in serious RP, especially in well-moderated servers where any foul play was swiftly dealt with. Sure it's a cheap way to reach that effect, but it achieved its purpose of making players think harder before mindlessly running around and dying like in the current state of mass non-serious RP. 10 seconds as the upper limit also just seems arbitrary, I think this whole thing should be removed so people can run their servers how they want to. If they abuse high respawn time, people won't play there and their current players will just suffer. You're just adding extra steps for servers to achieve what they want, which in the marketplace of ideas, the gamemode should act as a shell, not its own statement on how servers should be run.

The ``/advert`` thing with billboards was funny, but a lot of servers just overrided the command to return to how it was because their servers were already built around it:  you just added an annoying deterrence that people begrudgingly overcame, not changed as you might have expected. Even in the long run, it hasn't changed much: all the major DarkRP servers (based on player count/popularity metric) still use ``/advert`` like it originally was implemented, not the billboard version. I also know at least one server that has respawn times above 10 seconds, though I believe they may be using a custom DarkRP build.

The high respawn times made sense in PERP because there was a First Responder class that would get calls when someone died and they had time to go and revive them, creating a purpose and economy for that class by going to revive random people.  If there were no First Responders, the time would be much shorter. Limiting the respawn time in DarkRP makes it harder for someone to achieve that same environment, which if you really want to deter serious RP from being implemented with DarkRP, then I guess you achieved what you wished for with this.

FPtje:
> it achieved its purpose of making players think harder before mindlessly running around and dying

Bullshit. You don't "think harder" before dying. You get shot by someone else, and that's precisely why this logic does not, and will not ever work. The result of mindlessly messing about most directly is *other people* dying. With DarkRP people just kill way too often for this to be of any real use, ever. Like I said, you spend more time waiting than actually playing the game.

> 10 seconds as the upper limit also just seems arbitrary

It's the limit of what I was willing to deal with before ramming my keyboard through my monitor. That's still arbitrary, but a fair one in my book.

> If they abuse high respawn time, people won't play there and their current players will just suffer.

Sounds like something that would indeed happen, but the fact that this clamp is there proves that it most definitely did not. I remember this being a problem MANY servers had. Server owners copied the idea from other server owners, and after a while there just weren't many servers where you wouldn't get fucked over like this.

> In the marketplace of ideas, the gamemode should act as a shell, not its own statement on how servers should be run.

I have always made myself *very* clear in DarkRP. Making statements all over the place. See the workarounds file. Many arguments were had about such statements, and I very, **VERY** strongly stand by them. These statements have improved gameplay. They have prevented people from being fucked over by other people's incompetence. I will not let some bullshit about "free marketplace of ideas" shut me up. All my statements can be overridden without editing core files, if you so wish to go against them:

> If there were no First Responders, the time would be much shorter. Limiting the respawn time in DarkRP makes it harder for someone to achieve that same environment

Shortening the time would already require a custom hook. Put the other respawn time in that same hook and DarkRP will form no restriction.

> if you really want to deter serious RP from being implemented with DarkRP, then I guess you achieved what you wished for with this.

Serious RP is contradictory to the core concept of DarkRP in my opinion. But even if it wasn't, extremely high respawn times (without mechanics like the mentioned first responders) isn't serious RP, it's fucking people over. I can imagine something like the first responders being a valid game mechanic, but that is custom addon territory, and in that territory one is not much bothered by this clamp.

FPtje:
The core standpoint I have is this: sometimes you have to protect players of the game from incompetent or even malicious server owners and other developers. Whether that's Hex's malicious anticheat, the global ban lists people made, prop kill detection scripts that allowed normal players to get other players to get autokicked, or indeed respawn times that only apply to the plebs.

These statements have started great and passionate discussions in the past, just like it has now. Ultimately they have changed the community, the ideas people had about above subjects and the player experience for the better. Developers have been convinced that ban lists are a bad idea. Neosinos was shown that no matter what he tried, I could get random people kicked with his prop kill detection script. And with this, for years, respawn times have been acceptable on the vast majority of servers.

These interventions have had a positive effect on the game, even if all they did was start a discussion.

Kefta:
There's really no point in arguing this, the way you design the gamemode is like a pro-choice/pro-life argument (in antics and soft-comparison, not in the actual political belief behind it) - no matter the argument presented, you have convictions and a strong moral basis presented that will never change your mind. It's respectful from politically-engaged developers, but many server owners and players alike interpret DarkRP as a shell, not a basis for how you believe the gamemode should be used, so it comes into conflict regularly. Especially when the top 10 or more DarkRP servers run custom builds to get around your statements so they can make the gamemodes in their own heavily politicised vision. It just seems futile imo, but you do you.

Kefta:
Also, to respond specifically, the only point I have to refute is
>Bullshit. You don't "think harder" before dying. You get shot by someone else, and that's precisely why this logic does not, and will not ever work. The result of mindlessly messing about most directly is other people dying. With DarkRP people just kill way too often for this to be of any real use, ever. Like I said, you spend more time waiting than actually playing the game.

PERP wasn't mindless. There was no mindless shooting. If it occurred in the rare chance, admins would step in (on big servers, usually very mature and experienced ones) and would fix the issue at hand with strict punishments. The way it was handled is an argument for another PR, but it still enforced that the death time was worth it. Incident casualties in gun fights were seen as bystanders much as in real life, and those who shot faced heavy consequences from whatever clan or gang the person was a part of. It almost felt like real life, it felt political, which DarkRP has never aimed to imitate in execution.

ThatLing:
> if you really want to deter serious RP from being implemented with DarkRP

image

Kefta:
Yeah, people don't use it like that though. If they're all wrong, then so be it.

PR #3046, opened 2020-08-29
Change random sound method

Down-s:


Kefta:
Who wrote that in the first place lmfao

FlorianLeChat:
Going back in history, it looks like it was dated before the creation of this github.

PR #3047, opened 2020-08-30
Replace math.huge

Down-s:
Would make more sense to use the Jail Timer instead of math.huge


FPtje:
Not necessarily. If something resets the arrest timer, they'll respawn before they're unarrested. That `math.huge` means indefinitely, until some other process explicitly allows respawning.

PR #3048, opened 2020-08-30
Optimise FPP hud.lua

Down-s:
Localised some draw/surface functions for more optimisation

### DIFF
--- a/gamemode/modules/fpp/pp/client/hud.lua
+++ b/gamemode/modules/fpp/pp/client/hud.lua
@@ -29,6 +29,13 @@ hook.Add("GravGunPunt", "FPP_CL_GravGunPunt", function(ply, ent)
     end
 end)
 
+local surface_SetFont = surface.SetFont
+local surface_GetTextSize = surface.GetTextSize
+local surface_SetDrawColor = surface.SetDrawColor
+
+local draw_SimpleText = draw.SimpleText
+local draw_DrawText = draw.DrawText
+local draw_RoundedBox = draw.RoundedBox
 
 local HUDNote_c = 0
 local HUDNote_i = 1
@@ -41,8 +48,8 @@ function FPP.AddNotify( str, type )
     tab.recv    = SysTime()
     tab.velx    = 0
     tab.vely    = -5
-    surface.SetFont( "TabLarge" )
-    local w = surface.GetTextSize( str )
+    surface_SetFont( "TabLarge" )
+    local w = surface_GetTextSize( str )
 
     tab.x       = ScrW() / 2 + w * 0.5 + (ScrW() / 20)
     tab.y       = ScrH()
@@ -59,8 +66,10 @@ function FPP.AddNotify( str, type )
     HUDNote_c = HUDNote_c + 1
     HUDNote_i = HUDNote_i + 1
 
-    if not IsValid(LocalPlayer()) then return end -- I honestly got this error
-    LocalPlayer():EmitSound("npc/turret_floor/click1.wav", 10, 100)
+    local ply = LocalPlayer()
+
+    if not IsValid(ply) then return end -- I honestly got this error
+    ply:EmitSound("npc/turret_floor/click1.wav", 10, 100)
 end
 
 usermessage.Hook("FPP_Notify", function(u) FPP.AddNotify(u:ReadString(), u:ReadBool()) end)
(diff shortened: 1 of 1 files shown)


Kefta:
Do you have any benchmarks for comparison to show this makes a significant different?

Down-s:
Makes roughly a 5fps difference for me, better than nothing

FPtje:
Could be potentially a big performance increase for some people if some addon overrides these functions with some badly optimised change 🤷 

Then again, it could also break things for addons that override these functions.
Then again, they probably shouldn't in the first place.

Kefta:
Function overrides shouldn't really be a consideration here - if someone wants to detour core functions, they should be doing it as early as possible (before this file is loaded).

gamemode/modules/fpp/pp/client/hud.lua
```diff
@@ -83,17 +92,17 @@ local function DrawNotice(k, v, i)
...
 
-    surface.SetDrawColor( 255, 255, 255, v.a )
+    surface_SetDrawColor( 255, 255, 255, v.a )
 
-    draw.SimpleText(v.text, "TabLarge", x + 1, y + 1, Color(0, 0, 0, v.a * 0.8), TEXT_ALIGN_RIGHT)
-    draw.SimpleText(v.text, "TabLarge", x - 1, y - 1, Color(0, 0, 0, v.a * 0.5), TEXT_ALIGN_RIGHT)
-    draw.SimpleText(v.text, "TabLarge", x + 1, y - 1, Color(0, 0, 0, v.a * 0.6), TEXT_ALIGN_RIGHT)
-    draw.SimpleText(v.text, "TabLarge", x - 1, y + 1, Color(0, 0, 0, v.a * 0.6), TEXT_ALIGN_RIGHT)
-    draw.SimpleText(v.text, "TabLarge", x, y, Color(255, 255, 255, v.a), TEXT_ALIGN_RIGHT)
+    draw_SimpleText(v.text, "TabLarge", x + 1, y + 1, Color(0, 0, 0, v.a * 0.8), TEXT_ALIGN_RIGHT)
```
Kefta, line 101: The biggest time save in this function would be to inline all these draw.SimpleText calls. It caches internal data with the right alignment that you could only calc once for all these. Also, surface.SetDrawColor(num, num, num, num) is way more efficient than surface.SetDrawColor(Color(...)).

You can see the implementation here https://github.com/Facepunch/garrysmod/blob/34d2ccc460688750cb81c0ca031c75635aacfbaf/garrysmod/lua/includes/modules/draw.lua#L56

gamemode/modules/fpp/pp/client/hud.lua
```diff
@@ -188,12 +197,12 @@ local function HUDPaint()
     originalOwner = originalOwner ~= "" and (" (previous owner: %s)"):format(originalOwner) or ""
     reason = reason .. originalOwner
 
-    surface.SetFont("Default")
-    local w,h = surface.GetTextSize(reason)
+    surface_SetFont("Default")
+    local w,h = surface_GetTextSize(reason)
     local col = FPP.canTouchEnt(LAEnt, touchType) and Color(0, 255, 0, 255) or Color(255, 0, 0, 255)
```
Kefta, line 202: You can cache these two const colours outside of this function as upvalues.

gamemode/modules/fpp/pp/client/hud.lua
```diff
@@ -188,12 +197,12 @@ local function HUDPaint()
...
 
-    draw.RoundedBox(4, 0, ScrH() / 2 - h - 2, w + 10, 20, Color(0, 0, 0, 110))
-    draw.DrawText(reason, "Default", 5, ScrH() / 2 - h, col, 0)
-    surface.SetDrawColor(255, 255, 255, 255)
+    draw_RoundedBox(4, 0, ScrH() / 2 - h - 2, w + 10, 20, Color(0, 0, 0, 110))
```
Kefta, line 204: You can localise the duplicitous ScrH call and cache the const colour.

FPtje:
Oh hold on, this is a PR against the DarkRP repository. Please make it for the FPP repo.

gamemode/modules/fpp/pp/client/hud.lua
```diff
@@ -59,8 +66,10 @@ function FPP.AddNotify( str, type )
     HUDNote_c = HUDNote_c + 1
     HUDNote_i = HUDNote_i + 1
 
-    if not IsValid(LocalPlayer()) then return end -- I honestly got this error
-    LocalPlayer():EmitSound("npc/turret_floor/click1.wav", 10, 100)
+    local ply = LocalPlayer()
+
+    if not IsValid(ply) then return end -- I honestly got this error
```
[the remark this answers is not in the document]
FPtje: Why no return? That's a style I prefer.

FPtje:
I've merged this commit manually into the FPP repo, and merged that back in here. I've also included some of @Kefta's feedback. 
Thanks for your contributions! @Down-s and @Kefta!

PR #3049, opened 2020-09-07
police/sv_init: added hook to adjust arrest time

Marcus3x3:
Hook to adjust arrest time.

### DIFF
--- a/gamemode/modules/police/sv_init.lua
+++ b/gamemode/modules/police/sv_init.lua
@@ -340,6 +340,9 @@ function DarkRP.hooks:playerArrested(ply, time, arrester)
 
     if ply:isArrested() then return end -- hasn't been arrested before
 
+    local adjTime = hook.Call("adjustArrest", DarkRP.hooks)
+    time = adjTime or time
+    
     ply:PrintMessage(HUD_PRINTCENTER, DarkRP.getPhrase("youre_arrested", time))
 
     local phrase = DarkRP.getPhrase("hes_arrested", ply:Nick(), time)


FlorianLeChat:
I think you can use `hook.Run` function in this case, like: `local adjustedTime = hook.Run("adjustArrest", ply, time, arrester)`

FPtje:
The arrest time is generally decided by the caller of the arrest function. What do you want to use the hook for?

Marcus3x3:
Arrest function is being called by the arrest baton.

FPtje:
Well yeah, but the classic method there is to make a custom arrest baton or change the arrest time config setting. If you have two sources of that arrest time (parameter and hook) you're gonna get conflicts. The writer of the parameter thinking "hey I call with time = 60 seconds, why is it suddenly 90?"

So the question is, what kind of logic would you put in such a hook, and let's discuss if there are better ways to do that than this hook.

Marcus3x3:
The default arrest baton passes nil to the the arrest function. The goal is to use the default arrest baton instead of adding a custom one (less to maintain if default arrest baton gets updated in future). The logic that will be calculated inside the suggested hook will be along the lines of: more people you kill, higher arrest timer.

Right now there are hooks for canArrest and playerArrested. Neither of these hooks allow for an easy modification of the arrest time parameter. While the easy solution is to create a duplicate arrest baton, it requires more maintenance in the future should an update occur to the default one.

Bubka3:
Could add a hook to the default arrest stick for changing the time instead of only putting in nil for the time in arrest function.

```
local time = hook.Call("setStickArrestTime", DarkRP.hooks, ent) or nil
ent:arrest(time, Owner)
```

FPtje:
Yeah, the problem with a hook like the one in this PR is that it would also affect commands like `darkrp arrest ply time`.

Marcus3x3:
You are right, Bubka3's idea to place it inside the arrest stick is a better location. When I originally thought of this I didn't think about the arrest command (have never used it).

Whether the hook goes here, or in the arrest stick (which is honestly a better place for it), the goal is to allow a easy way to adjust the time of an arrest. If necessary I'll copy the arrest stick just to change two lines (really don't want to have to do it). I'm going to close this PR and open a new one with the hook relocated to the arrest stick.

PR #3050, opened 2020-09-08
setArrestStickTime hook added

Marcus3x3:
Allows a dynamic time to be set when being arrested from the arrest stick (instead of a constant set from the gamemode config).

### DIFF
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -53,8 +53,13 @@ DarkRP.hookStub{
     description = "Sets arrest time during arrest via the arrest stick",
     parameters = {
         {
-            name = "ent",
-            description = "The player being arrested."
+            name = "arrester",
+            description = "The player trying to arrest someone.",
+            type = "Player"
+        },
+        {
+            name = "arrestee",
+            description = "The player being arrested.",
             type = "Player"
         }
     },
@@ -105,7 +110,7 @@ function SWEP:PrimaryAttack()
         return
     end
 
-    local time = hook.Call("setArrestStickTime", DarkRP.hooks, ent)
+    local time = hook.Call("setArrestStickTime", DarkRP.hooks, Owner, ent)
     ent:arrest(time, Owner)
     DarkRP.notify(ent, 0, 20, DarkRP.getPhrase("youre_arrested_by", Owner:Nick()))
 
--- a/entities/weapons/arrest_stick/shared.lua
+++ b/entities/weapons/arrest_stick/shared.lua
@@ -48,6 +48,26 @@ DarkRP.hookStub{
     realm = "Server"
 }
 
+DarkRP.hookStub{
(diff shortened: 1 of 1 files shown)


FlorianLeChat:
Can you pass the weapon's owner in this hook ? (like previous hook.Call in the file : https://github.com/Marcus3x3/DarkRP/blob/474827e4a466e6e4d7a6375312135f82ccef559e/entities/weapons/arrest_stick/shared.lua#L102)

Marcus3x3:
I suppose you can, but I didn't see the need for it. My thought process was that in any use case for this hook the arrest time of the victim wouldn't be altered by any data from the player arresting him. The only thing this hook would do is set a dynamic time to a player's jail sentence; I don't think there is a need to have the owner passed to it.

The end result of the hook (if implemented) would be that it allows me to make the gameplay more enjoyable. The goal is to allow players who get manually wanted (when there isn't proof they did anything) a lighter jail sentence (cause who wants to spend 2 minutes+ because a minge 'said' they did something), while still being able to jail the guy that mass-murdered 15 people an appropriate jail sentence.

Kefta:
The owner should absolutely be passed into the hook.

FlorianLeChat:
> I suppose you can, but I didn't see the need for it. My thought process was that in any use case for this hook the arrest time of the victim wouldn't be altered by any data from the player arresting him.

Don't worry, the hook will always return a single argument which will be the arrest time. If you pass the weapon's owner, that will allow you to make operations on its owner.
**Dumb example**: if the owner is an administrator then the arrest time is multiplied by 2.

Marcus3x3:
Is a good point, if PR is accepted I imagine he'll modify it by passing in the Owner, too.

I didn't include it because there is no use case that I'd personally use that needed it :D Others might find use of having the owner inside the hook, though.

FlorianLeChat:
As you are new to Github, I inform you that as the creator of a pull request, you can always make changes as long as it hasn't been merged. Currently, you can still modify this [file](https://github.com/Marcus3x3/DarkRP/edit/patch-2/entities/weapons/arrest_stick/shared.lua) on your fork of this repository. As soon as you have finished editing the file, the modification will be displayed here. You can find out more [here](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request).

Marcus3x3:
Thanks 👍

Bubka3:
Finally, a way to whip my big super admin dick onto people by quadrupling their arrest time if they were arrested by me. Thank you!

Bubka3:
If it's not obvious the above comment is a joke. I could of already done this by making an owners baton. Also, that's fucking stupid so don't do it.

Kefta:
Cool man

FPtje:
Thanks for the contribution. I've fixed an error and added the arrest stick weapon to the hook. That makes the arrester argument in the hook kind of redundant (since you can get it with `weapon.Owner()`), but it's still easier to use.

PR #3053, opened 2020-09-20
Break from loop when command is found

Kapppa:


FPtje:
That break should have been there in the first place. Thanks!

PR #3058, opened 2020-10-19
Fix error when hitting csent limit

mcNuggets1:


FPtje:
That's not a fix, that's a mask

mcNuggets1:
That is a fix for the lua error, not for the second red error message which tells you already, that the limit is hit.

FPtje:
Oh, that's useful. In that case, could you add a comment?

FPtje:
Thanks

PR #3059, opened 2020-11-04
workshopid added to darkrp.txt

7aninhos:
so we don't have to do resource.AddWorkshop('248302805') as long DarkRP is in the workshop collection


FPtje:
Thanks!

PR #3060, opened 2020-11-11
Localize vector constants

djtb2924:
I know it's micro-optimisation but why not.

Just localizing vector before loops


mcNuggets1:
Maybe we should use Player:GetHull() for this anyways?
Servers wo use the playerscale module would appreciate.

FlorianLeChat:
I'm not sure but `ply:SetHull` (or a kind of collisions update) isn't called automatically after `ply:SetModelScale` since the last update?

mcNuggets1:
Obviously.
ModelScale and Hull is different.
It never updated Hull, as far as I'm concerned.

FPtje:
Thanks!

PR #3061, opened 2020-11-11
Change weapon cloak method

djtb2924:
Actual cloak works like this :
loop for all targets > SetNoDraw(true) > loop trough ents.GetAll for phys_beam > SetNoDraw(true)
\+ do that every tick

My method :
loop for all targets > SetNoDraw(true) > loop trough target:GetChildren() for phys_beam > SetNoDraw(true)
\+ do that every PlayerSwitchWeapon hook

I think that is better


djtb2924:
Ok, I think PlayerSwitchWeapon is called too early, and add a timer or something is really messy.

djtb2924:
The best solution to me :
- Uses PlayerSwitchWeapon
- Hide Physgun Beam clientside => No more latency when switching at low tickrate

djtb2924:
Not working, fuck it

PR #3062, opened 2020-11-18
Use hull instead of wrong player size

mcNuggets1:
Hull correctly returns 16, 16, 72 for the players height and width.


FPtje:
I find it fascinating that after all those years things like this can still be found. Thanks for fixing it!

PR #3064, opened 2020-11-24
Add more control to spawned entities without overriding

### LINKED ISSUE #3063: [Request] Add hooks when a player interacts with default DarkRP entities.
### Description of the request
It would be nice for new modders to add hooks when a player interacts with a DarkRP entity (detouring the `Use` function in entity tables is not easy). It would typically be called after the hook `canDarkRPUse`. The [money entity](https://github.com/FPtje/DarkRP/blob/aa0dc11eb1a9018be17f7fb262180e386d28861e/entities/entities/spawned_money/init.lua#L35) already has that and could be extended to other `spawned_*` entities or all default entities if possible.

I could have made a pull request to make these changes but I wouldn't have my laptop to do and test this for a while.

FlorianLeChat:
**Original request:**
> It would be nice for new modders to add hooks when a player interacts with a DarkRP entity (detouring the `Use` function in entity tables is not easy). It would typically be called after the hook `canDarkRPUse`. The [money entity](https://github.com/FPtje/DarkRP/blob/aa0dc11eb1a9018be17f7fb262180e386d28861e/entities/entities/spawned_money/init.lua#L35) already has that and could be extended to other `spawned_*` entities or all default entities if possible.

**Note**: the [PlayerUse](https://wiki.facepunch.com/gmod/GM:PlayerUse) hook is called before `canDarkRPUse`, so this is not really possible without overriding entities through the darkrp modification.

**Here are the changes**:
* Add `playerPickedUpAmmo` hook when a player picks up spawned ammo.
* Add `playerAteFood` hook when a player eats food.
* Add `playerOpenedShipment` hook when a player opens a spawned shipment.
* Add `playerPickedUpWeapon` hook when a player picks up a spawned weapon.
* Add a `canDarkRPUse` hook check in **food** entity.
* Changes the way of adding energy in **food** entity (reduces unnecessary operations).
* Fix spaces inconsistencies.
* Fix function arguments inconsistencies.

This closes #3063.

### DIFF
--- a/entities/entities/food/init.lua
+++ b/entities/entities/food/init.lua
@@ -34,7 +34,13 @@ function ENT:OnTakeDamage(dmg)
 end
 
 function ENT:Use(activator, caller)
-    caller:setSelfDarkRPVar("Energy", math.Clamp((caller:getDarkRPVar("Energy") or 0) + 100, 0, 100))
+    local canUse, reason = hook.Call("canDarkRPUse", nil, activator, self, caller)
+    if canUse == false then
+        if reason then DarkRP.notify(activator, 1, 4, reason) end
+        return
+    end
+
+    caller:setSelfDarkRPVar("Energy", 100)
     umsg.Start("AteFoodIcon", caller)
     umsg.End()
 
--- a/entities/entities/darkrp_cheque/init.lua
+++ b/entities/entities/darkrp_cheque/init.lua
@@ -21,12 +21,11 @@ function ENT:Initialize()
     hook.Add("PlayerDisconnected", self, self.onPlayerDisconnected)
 end
 
-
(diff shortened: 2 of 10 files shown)


entities/entities/food/init.lua
```diff
@@ -34,7 +34,13 @@ function ENT:OnTakeDamage(dmg)
 end
 
 function ENT:Use(activator, caller)
-    caller:setSelfDarkRPVar("Energy", math.Clamp((caller:getDarkRPVar("Energy") or 0) + 100, 0, 100))
```
FPtje, line 37: This clamp here makes no sense. Thanks for fixing it!

FPtje (approved):
LGTM!

Issue #3065, opened 2020-11-25
Pysgun Twirl bug


KennyHarrer:
[//]: # (DO NOT REMOVE THIS TEMPLATE.)
[//]: # (GITHUB ISSUES ARE **NOT** FOR HELP)
[//]: # (GO TO THE DISCORD INSTEAD: https://darkrp.page.link/discord )

### Description of the bug
[//]: # (Describe the issue as accurately as possible)
physgun_check works while in a chair (finger twirl thing)

### How to make the bug happen
[//]: # (Knowing how to make a bug happen allows the developers to figure out)
[//]: # (what the cause of the problem is and whether a certain fix solves it)
pull out your physgun, sit in a chair. Aim at a player and left click.

### Lua errors
[//]: # (Note that errors on server startup are more important than other ones)
[//]: # (because they can cause other errors. Always look in the startup log of the server!)
n/a

### Why the developer of DarkRP is responsible for this issue
[//]: # (By posting on Github, you ask the developers of DarkRP to solve the problem)
[//]: # (It may seem obvious to you, but you have to make clear why they are the right people to look at it)
because you missed a check

I would make a pull request but I can't be fucked to do it right now.

mcNuggets1:
Dafuq

Get fucked to do it please, we don't understand.

KennyHarrer:
> 
> 
> Dafuq
> 
> Get fucked to do it please, we don't understand.

Done :)

FPtje:
Fixed by #3066

PR #3066, opened 2020-11-26
don't allow physGunCheck while seated

KennyHarrer:
Without this change you would still have a "finger twirl" while seated.
Bug mentioned in #3065

### DIFF
--- a/gamemode/modules/animations/sh_animations.lua
+++ b/gamemode/modules/animations/sh_animations.lua
@@ -29,7 +29,7 @@ end
 local function physGunCheck(ply)
     local hookName = "darkrp_anim_physgun_"..ply:EntIndex()
     hook.Add("Think", hookName, function()
-        if IsValid(ply) and ply:Alive() and ply:GetActiveWeapon():IsValid() and ply:GetActiveWeapon():GetClass() == "weapon_physgun" and ply:KeyDown(IN_ATTACK) then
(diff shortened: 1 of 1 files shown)


KennyHarrer:
whoops added an extra not in there

gamemode/modules/animations/sh_animations.lua
```diff
@@ -29,7 +29,7 @@ end
 local function physGunCheck(ply)
     local hookName = "darkrp_anim_physgun_"..ply:EntIndex()
     hook.Add("Think", hookName, function()
-        if IsValid(ply) and ply:Alive() and ply:GetActiveWeapon():IsValid() and ply:GetActiveWeapon():GetClass() == "weapon_physgun" and ply:KeyDown(IN_ATTACK) then
+        if IsValid(ply) and ply:Alive() and ply:GetActiveWeapon():IsValid() and ply:GetActiveWeapon():GetClass() == "weapon_physgun" and ply:KeyDown(IN_ATTACK) and (not ply:GetAllowWeaponsInVehicle() and ply:InVehicle()) then
```
FPtje, line 32: This will break the animation when the player is not in a vehicle, but `ply:GetAllowWeaponsInVehicle()` return `false`.

The correct condition would be `and (ply:GetAllowWeaponsInVehicle() or not ply:InVehicle())`. That way, the player is allowed to do the animation when they can hold the physgun in a vehicle, or they are not in a vehicle.

FPtje (changes requested):

FPtje:
Merged! Thanks for the contribution!

PR #3067, opened 2020-11-26
Fix rare clientside error

mcNuggets1:
Very rare error.

### DIFF
--- a/gamemode/modules/base/cl_gamemode_functions.lua
+++ b/gamemode/modules/base/cl_gamemode_functions.lua
@@ -3,6 +3,7 @@ local mouseX, mouseY = ScrW() / 2, ScrH() / 2
 function GM:ShowSpare1()
     local jobTable = LocalPlayer():getJobTable()
 
+    -- We need to check for the existance of jobTable here, because in very rare edge cases, the player's team isn't set, when the getJobTable-function is called here.
     if jobTable and jobTable.ShowSpare1 then
         return jobTable.ShowSpare1(LocalPlayer())
     end
@@ -20,6 +21,7 @@ end
 function GM:ShowSpare2()
     local jobTable = LocalPlayer():getJobTable()
 
+    -- We need to check for the existance of jobTable here, because in very rare edge cases, the player's team isn't set, when the getJobTable-function is called here.
     if jobTable and jobTable.ShowSpare2 then
         return jobTable.ShowSpare2(LocalPlayer())
     end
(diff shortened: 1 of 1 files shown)


FPtje:
`getJobTable` throws a descriptive error when the job table does not exist. How would this error be possible?

mcNuggets1:
It's extremely rare and clientside only.
Its upon joining a dedicated Server game, the error message is non descriptive and not possible to track or fix.
This will get rid of this rare error.

FPtje:
What is the error message? Does it complain about jobTable being nil? Who is running those functions before join finished? Some join menu script?

mcNuggets1:
[ERROR] gamemodes/darkrp/gamemode/modules/base/cl_gamemode_functions.lua:23: attempt to index local 'jobTable' (a nil value)
  1. Call - gamemodes/darkrp/gamemode/modules/base/cl_gamemode_functions.lua:23
   2. unknown - gamemodes/darkrp/gamemode/modules/base/cl_gamemode_functions.lua:62

This is the message.

mcNuggets1:
Its while joining, i dont know why, it cant be traced.
Its gmod at its best.

DBotThePony:
it will still throw an error since you still index `jobTable` inside your `if` condition

mcNuggets1:
It checks if jobTable exists tho in the first place, which the error pinpoints, it doesn't.
Happens on newest 64 bit.
If its not existing and not indexiert, its not erroring.

DBotThePony:
I can only assume this happen when player spam F4 on loading screen and short time after, somehow before player receive their team number

FPtje:
I think I understand why:
https://github.com/FPtje/DarkRP/blob/27d33d7aaf26288e96507cb590ad0ec92958b1a1/gamemode/modules/base/sh_createitems.lua#L396

getJobTable doesn't throw an error when the player is not yet initialized. Could you change the condition to use that same initialized variable and add comment why it's there? That's much better than just the jobTable check imo.

mcNuggets1:
I may misunderstand you, but are you asking me to add a check for a servside value in a clientside environment, which didn't receive said value to check if the player is initialized?

FPtje:
If that var does not exist clientside then it will never throw that error clientside. That's a deeper problem to solve.

mcNuggets1:
That's a serverside var for initializing database stuff onto the player.
I looked everywhere for it clientside, it's not there.

This can't error clientside.

FPtje:
Very well. Can you add a comment?

mcNuggets1:
Done.

FPtje:
Thanks!

PR #3068, opened 2020-11-28
Replace some fonts

kirukiru357:
Replace ugly TargetID font to Roboto
Replace Windows-only Tahoma to Roboto
Replace non-free latin-only Coolvetica to Roboto

### DIFF
--- a/gamemode/modules/base/cl_fonts.lua
+++ b/gamemode/modules/base/cl_fonts.lua
@@ -2,17 +2,15 @@
 The fonts that DarkRP uses
 ---------------------------------------------------------------------------]]
 local function loadFonts()
-    local robotoSize = system.IsLinux() and fp{fn.Flip(fn.Add), 1} or fn.Id
-
     surface.CreateFont("Target", {
-        size = robotoSize(20),
+        size = 20,
         weight = 600,
         antialias = true,
         shadow = false,
         font = "Roboto"})
 
     surface.CreateFont("DarkRPHUD1", {
-        size = robotoSize(20),
+        size = 20,
         weight = 600,
         antialias = true,
         shadow = true,
@@ -54,7 +52,7 @@ local function loadFonts()
         font = "Trebuchet MS"})
 
     surface.CreateFont("TabLarge", {
-        size = robotoSize(18),
+        size = 18,
         weight = 700,
         antialias = true,
         shadow = false,
--- a/gamemode/modules/base/cl_fonts.lua
+++ b/gamemode/modules/base/cl_fonts.lua
@@ -2,13 +2,6 @@
 The fonts that DarkRP uses
 ---------------------------------------------------------------------------]]
 local function loadFonts()
-    surface.CreateFont("Target", {
-        size = 20,
-        weight = 600,
-        antialias = true,
-        shadow = false,
-        font = "Roboto"})
-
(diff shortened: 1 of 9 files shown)


Kefta:
You shouldn't change the name of the GM function.

kirukiru357:
> You shouldn't change the name of the GM function.

Sorry, my bad. Fixed.

gamemode/modules/base/cl_fonts.lua
```diff
@@ -2,22 +2,28 @@
 The fonts that DarkRP uses
 ---------------------------------------------------------------------------]]
 local function loadFonts()
-    local tahoma = system.IsLinux() and "DejaVu Sans" or "Tahoma"
-    local tahomaSize = system.IsLinux() and fp{fn.Flip(fn.Add), 1} or fn.Id
+    local robotoSize = system.IsLinux() and fp{fn.Flip(fn.Add), 1} or fn.Id
```
FPtje, line 5: Does the hack that was applied for Tahoma really also apply for Roboto?

FPtje (commented):
The roboto font indeed looks nicer. One question about a hack in cl_fonts.

gamemode/modules/base/cl_fonts.lua
mcNuggets1: No it does not.
Roboto may look different on linux however, I can't know.

gamemode/modules/base/cl_fonts.lua
```diff
@@ -2,22 +2,28 @@
...
+
+    surface.CreateFont("Target", {
```
mcNuggets1, line 7: I like your changes, but calling a font "Target" these days is a dangerous step.
It could override fonts with the same name from other addons.

mcNuggets1 (changes requested):
Looks much cleaner, but I would change the fontname "Target" to something else.
Potential conflicts with different fonts.

kirukiru357:
The main two reasons i suggested it because when non-latin words used in Tab menu it hurts my eyes and default TargetID font is just ugly and small.

gamemode/modules/base/cl_fonts.lua
```diff
@@ -2,22 +2,26 @@
...
+        font = "Roboto"})
 
     surface.CreateFont("DarkRPHUD2", {
         size = 23,
         weight = 400,
         antialias = true,
         shadow = false,
-        font = "Coolvetica"})
+        font = "Roboto"})
+
```
FPtje, line 18: Trailing whitespace

FPtje (changes requested):
I see you've merged master into your branch. I'd prefer it if you rebased instead. Otherwise, when I merge your branch into master, the logs will have a commit called `Merge branch master into master`, which then makes little sense.

FPtje:
One of the consequences might be that other addons that use DarkRP's fonts suddenly switch to being based on Roboto. I don't think that's a bad thing, since it looks nicer. Let's hope there's not too much of a change in size, causing things to break 👀

kirukiru357:
Rebased and removed trailing whitespace.

kirukiru357:
Probably best is to leave it as is, because rebase causes even more nonsense commits.

FPtje:
Thanks!

PR #3069, opened 2020-11-30
Minor optimisations to spawn and death code

mcNuggets1:
* Don't do a ents.FindByClass loop (internal ents.GetAll loop, which are extremely expensive) just to extinguish all viewmodels, especially because we only want to extinguish our own viewmodel.
(Viewmodels cant be ignited by the printer anymore anyways...)

* It is unnecessary to send a usermessage to every player respawning, if the player never had blackscreen enabled upon.

* Use ipairs instead of pairs for ents.FindIn* loop for money_printer.

1 major optimisations and 2 minor optimisations sqashed into a single commit, I hope you like it.

### DIFF
--- a/gamemode/modules/afk/sv_afk.lua
+++ b/gamemode/modules/afk/sv_afk.lua
@@ -19,10 +19,8 @@ local function SetAFK(ply)
     local rpname = ply:getDarkRPVar("rpname")
     ply:setSelfDarkRPVar("AFK", not ply:getDarkRPVar("AFK"))
 
-    if not ply.blackScreen then
-        ply.blackScreen = ply:getDarkRPVar("AFK")
-        SendUserMessage("blackScreen", ply, ply:getDarkRPVar("AFK"))
-    end
+    ply.blackScreen = ply:getDarkRPVar("AFK")
+    SendUserMessage("blackScreen", ply, ply:getDarkRPVar("AFK"))
 
     if ply:getDarkRPVar("AFK") then
         DarkRP.retrieveSalary(ply, function(amount) ply.OldSalary = amount end)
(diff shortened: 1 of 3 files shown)


Kefta:
You need to extinguish all three viewmodels the player has in the case a weapon base utilises the other two.

mcNuggets1:
At this point I ask myself, if this is really necessary.
The money printer doesn't ignite view models anymore.

gamemode/modules/afk/sv_afk.lua
```diff
@@ -19,7 +19,10 @@ local function SetAFK(ply)
     local rpname = ply:getDarkRPVar("rpname")
     ply:setSelfDarkRPVar("AFK", not ply:getDarkRPVar("AFK"))
 
-    SendUserMessage("blackScreen", ply, ply:getDarkRPVar("AFK"))
+    if not ply.blackScreen then
+        ply.blackScreen = ply:getDarkRPVar("AFK")
+        SendUserMessage("blackScreen", ply, ply:getDarkRPVar("AFK"))
```
FPtje, line 24: You can re-use `ply.blackScreen` here:

```suggestion
        SendUserMessage("blackScreen", ply, ply.blackScreen)
```

Also, because of the condition this only sends the usermessage when a player is going AFK. Surely you'd want a message to turn it off when the player comes back from AFK?

Perhaps `ply.blackScreen == nil or ply.blackScreen ~= ply:getDarkRPVar("AFK")`?

FPtje (changes requested):
Thanks for the contribution, one point of feedback

gamemode/modules/afk/sv_afk.lua
mcNuggets1: Good catch, dude.

FPtje:
Thanks!

PR #3078, opened 2021-01-01
a slightly smarter solution for hobo zombie sounds

cresterienvogel:
i believe it will be better this way

### DIFF
--- a/gamemode/modules/hobo/sv_hobo.lua
+++ b/gamemode/modules/hobo/sv_hobo.lua
@@ -1,21 +1,37 @@
-local function MakeZombieSoundsAsHobo(ply)
-    if ply:EntIndex() == 0 then return end
-    if not ply.nospamtime then
-        ply.nospamtime = CurTime() - 2
-    end
+local sounds = {
+	[1] = {
+		range = 14,
+		path = "npc/zombie/zombie_voice_idle"
+	},
+	[2] = {
+		range = 6,
+		path = "npc/zombie/zombie_pain"
+	},
+	[3] = {
+		range = 3,
+		path = "npc/zombie/zombie_alert"
+	}
+}
 
-    local Team = ply:Team()
-    if not RPExtraTeams[Team] or not RPExtraTeams[Team].hobo or CurTime() < (ply.nospamtime + 1.3) or (ply:GetActiveWeapon():IsValid() and ply:GetActiveWeapon():GetClass() ~= "weapon_bugbait") then
-        return
-    end
-    ply.nospamtime = CurTime()
-    local ran = math.random(1,3)
-    if ran == 1 then
-        ply:EmitSound("npc/zombie/zombie_voice_idle" .. math.random(1, 14) .. ".wav", 100, 100)
-    elseif ran == 2 then
-        ply:EmitSound("npc/zombie/zombie_pain" .. math.random(1, 6) .. ".wav", 100, 100)
-    elseif ran == 3 then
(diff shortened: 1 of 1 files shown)


DBotThePony:
spaces

gamemode/modules/hobo/sv_hobo.lua
```diff
@@ -1,21 +1,37 @@
-local function MakeZombieSoundsAsHobo(ply)
-    if ply:EntIndex() == 0 then return end
-    if not ply.nospamtime then
-        ply.nospamtime = CurTime() - 2
-    end
+local sounds = {
+	[1] = {
```
FPtje, line 2: Why assign these numbers explicitly? Wouldn't it be less error prone to leave them out?

FPtje (changes requested):
Thanks for the PR! It does look a bit cleaner. Some comments:

DarkRP uses 4 spaces for indentation rather than tabs.

cresterienvogel (commented):
just some old habit i should get rid of, you are right tho

FlorianLeChat:
Could you replace all the `pl` with `ply`? It's insignificant but it keeps some consistency with the DarkRP code.

gamemode/modules/hobo/sv_hobo.lua
```diff
@@ -1,21 +1,37 @@
...
+    {
+        range = 3,
+        path = "npc/zombie/zombie_alert"
+    }
+}
+
 local function MakeZombieSoundsAsHobo(ply)
-    if ply:EntIndex() == 0 then return end
+    if ply:EntIndex() == 0 then 
+        return
```
FPtje, line 18: Trailing whitespace on both lines. With most modern editors you can turn on a setting to "trim trailing whitespace", which automatically removes spaces and tabs at the end of lines upon saving a file.

gamemode/modules/hobo/sv_hobo.lua
```diff
@@ -1,21 +1,37 @@
...
+        return 
+    end
+
     if not ply.nospamtime then
         ply.nospamtime = CurTime() - 2
     end
 
-    local Team = ply:Team()
-    if not RPExtraTeams[Team] or not RPExtraTeams[Team].hobo or CurTime() < (ply.nospamtime + 1.3) or (ply:GetActiveWeapon():IsValid() and ply:GetActiveWeapon():GetClass() ~= "weapon_bugbait") then
+    local team = ply:Team()
```
FPtje, line 25: `team` is a global in GMod that has a real meaning. Using it as a local variable can accidentally break things like `team.GetName` if you're not careful. It's better to avoid using globals as local variables.

gamemode/modules/hobo/sv_hobo.lua
```diff
@@ -1,21 +1,37 @@
...
-    if ran == 1 then
-        ply:EmitSound("npc/zombie/zombie_voice_idle" .. math.random(1, 14) .. ".wav", 100, 100)
-    elseif ran == 2 then
-        ply:EmitSound("npc/zombie/zombie_pain" .. math.random(1, 6) .. ".wav", 100, 100)
-    elseif ran == 3 then
-        ply:EmitSound("npc/zombie/zombie_alert" .. math.random(1, 3) .. ".wav", 100, 100)
-    end
+
+    local ran = math.random(1, 3)
+    local sound = sounds[ran]
```
FPtje, line 33: Same for `sound`. Some editors have a GMod Lua plugin (Sublime, VS Code and Atom do), which highlight

FPtje (approved):
Thanks for the changes. There are some minor comments, but I'll fix them post merge :+1:

PR #3080, opened 2021-01-06
Add playerGotLicense hook

djtb2924:
Hi ! There is no hook when a player got a gun license.

Params : player, mayor


FPtje:
Thanks! Could you add a hook stub?

FPtje:
Thanks!

PR #3085, opened 2021-01-15
add method customEntityCount

### LINKED ISSUE #3084: Expose maxEntities
[//]: # (GO TO THE DISCORD INSTEAD: https://darkrp.page.link/discord )

### Description of the bug
I was wanting to restrict access to cop jobs if players already had money printers spawned.
Exposing the maxEntities table as a global would solve my issue.
https://github.com/FPtje/DarkRP/blob/35ca64069be4e090604544d9fe8a4b41aafeba87/gamemode/modules/base/sv_entityvars.lua#L223

This is a feature request.

### Lua errors
N/A

It's not possible to achieve what I want without editing the core files, which the developer of DarkRP strongly suggests against.

someoneyouspite:
Resolves #3084

### DIFF
--- a/gamemode/modules/base/sv_entityvars.lua
+++ b/gamemode/modules/base/sv_entityvars.lua
@@ -245,6 +245,11 @@ function meta:customEntityLimitReached(entTable)
     return max ~= 0 and maxEntities[self][entTable.cmd] >= max
 end
 
+function meta:customEntityCount(entTable)
+    local entities = maxEntities[self]
+    return (entTable and entities) and entities[entTable.cmd] or 0
+end
+
 hook.Add("PlayerDisconnected", "removeLimits", function(ply)
     maxEntities[ply] = nil
     net.Start("DarkRP_DarkRPVarDisconnect")
--- a/gamemode/modules/base/sv_entityvars.lua
+++ b/gamemode/modules/base/sv_entityvars.lua
@@ -222,16 +222,12 @@ Maximum entity values
 ---------------------------------------------------------------------------]]
 local maxEntities = {}
 function meta:addCustomEntity(entTable)
-    if not entTable then return end
-
     maxEntities[self] = maxEntities[self] or {}
     maxEntities[self][entTable.cmd] = maxEntities[self][entTable.cmd] or 0
     maxEntities[self][entTable.cmd] = maxEntities[self][entTable.cmd] + 1
 end
 
 function meta:removeCustomEntity(entTable)
-    if not entTable.cmd then return end
-
     maxEntities[self] = maxEntities[self] or {}
     maxEntities[self][entTable.cmd] = maxEntities[self][entTable.cmd] or 0
     maxEntities[self][entTable.cmd] = maxEntities[self][entTable.cmd] - 1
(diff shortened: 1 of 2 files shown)


FPtje:
Thanks! Could you add a function stub!

FPtje:
That exclamation mark was supposed to be a question mark* :see_no_evil:

someoneyouspite:
Will do.

Kefta:
You should not treat the parameter as optional, it will just lead to silent errors. Instead, you should implement it as
```Lua
local entities = maxEntities[self]
if entities == nil then return 0 end

entities = entities[entTable.cmd]
if entities == nil then return 0 end

return entities
```
This is the most efficient impl and the most intuitive to read imo. You could also exchange the last ``== nil`` statement for ``return entities or 0``, but this is not type safe since Lua doesn't have real ternary operators, but to each their own.

The optional parameter discrepancy should also be fixed in ``addCustomEntity`` and ``removeCustomEntity``.

someoneyouspite:
It could still error if passed a non-table for entTable, should I make sure it's type safe like the following?

```lua
    local entities = maxEntities[self]
    if entities == nil then return 0 end

    if type(entTable) == "table" then
        entities = entities[entTable.cmd]
    end
    if entities == nil then return 0 end

    return entities
```

Kefta:
No, the DarkRP API does almost no type checking for its arguments across the codebase so you should assume the passed arg is a table. If a table isn't passed, an error (not a very helpful one but it's better than no error) will occur. This should change in the future when proper CheckType functions are added to GMod's API.

FoksVHox:
> No, the DarkRP API does almost no type checking for its arguments across the codebase so you should assume the passed arg is a table. If a table isn't passed, an error (not a very helpful one but it's better than no error) will occur. This should change in the future when proper CheckType functions are added to GMod's API.

Has there been any talk about that?

Kefta:
> > No, the DarkRP API does almost no type checking for its arguments across the codebase so you should assume the passed arg is a table. If a table isn't passed, an error (not a very helpful one but it's better than no error) will occur. This should change in the future when proper CheckType functions are added to GMod's API.
> 
> Has there been any talk about that?

I have added them with the new sandboxed GMod Lua API, but there are currently issues with pushing it as it relies on C++20 which is incompatible with Steam runtime. I would not count on them being released immediately, so for now, the solution I posted above is the best implementation.

someoneyouspite:
Do you want me to fix the optional parameters in `addCustomEntity`, `removeCustomEntity`, and `customEntityLimitReached` as well?
They all contain this snippet.
```lua
maxEntities[self] = maxEntities[self] or {}
maxEntities[self][entTable.cmd] = maxEntities[self][entTable.cmd] or 0
```
I was considering replacing them with the following.
```lua
if maxEntities[self] == nil then
    maxEntities[self] = {}
end
if maxEntities[self][entTable.cmd] == nil then
    maxEntities[self][entTable.cmd] = 0
end
```

FoksVHox:
> ```lua
> if maxEntities[self] == nil then
>     maxEntities[self] = {}
> end
> if maxEntities[self][entTable.cmd] == nil then
>     maxEntities[self][entTable.cmd] = 0
> end
> ```
That's just a mess, the one above would result in cleaner code.

someoneyouspite:
I would agree in this case, yet I'm not suggesting the fix.

Kefta:
I was referring to
```Lua
if not entTable then return end
```
and
```Lua
if not entTable.cmd then return end
```
Those lines should be removed.

Kefta (approved):

someoneyouspite:
Fixed the bad commit message, shouldn't be doing pull requests at 4 am. 🤦

FPtje:
Thanks for the effort! Code looks good!

PR #3086, opened 2021-01-20
Hashing player positions into a grid in O(N) before computing every interaction for proximity chat

JurisBog:
This implementation aims to help the voice radius implementation scaling by separating each player into a grid (based on a hash). There is no reason why we should check if two players can talk to each other if they are on opposite sides of the map, this method ensures that by clustering. Additionally, if there are two clusters of 10 people, there would be less computation as only 10^2+10^2=200 would be checked instead of 20^2=400 without considering symmetry.

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -334,11 +334,13 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
     grid = {}
 
     local plyPos = {}
+    local eyePos = {}
 
     -- Get the grid position of every player O(N)
     for _, ply in ipairs(players) do
         local pos = ply:GetPos()
         plyPos[ply] = pos
+        eyePos[ply] = ply:EyePos()
         local x = floor(pos.x / gridSize)
         local y = floor(pos.y / gridSize)
 
@@ -359,7 +361,8 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
     for _, ply1 in ipairs(players) do
         local gridX = plyToGrid[1][ply1]
         local gridY = plyToGrid[2][ply1]
-        local ply1Pos
+        local ply1Pos = plyPos[ply1]
+        local ply1EyePos = eyePos[ply1]
 
         for x = gridX, gridX + 1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
             local row = grid[x]
@@ -372,11 +375,9 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
 
                 -- Check if every player can talk
                 for _, ply2 in ipairs(cell) do
-                    ply1Pos = ply1Pos or plyPos[ply1]
-                    local ply2Pos = plyPos[ply2]
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
-                        (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
-                            (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room
+                        (ply1Pos:DistToSqr(plyPos[ply2]) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                            (not dynv or IsInRoom(ply1EyePos, eyePos[ply2], ply2))) -- Dynamic voice is on and players are in the same room
 
                     DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
                     DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
@@ -393,11 +394,9 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
             for i = 1, count do
                 for j = i + 1, count do
                     local ply1, ply2 = cell[i], cell[j]
-                    local ply1Pos, ply2Pos = plyPos[ply1], plyPos[ply2]
-
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
-                        (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
-                            (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room
+                        (plyPos[ply1]:DistToSqr(plyPos[ply2]) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                            (not dynv or IsInRoom(eyePos[ply1], eyePos[ply2], ply2))) -- Dynamic voice is on and players are in the same room
 
                     DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
                     DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -364,24 +364,25 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
         local ply1Pos = plyPos[ply1]
         local ply1EyePos = eyePos[ply1]
 
-        for x = gridX, gridX + 1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
+        for i = 0, 3 do
+            local vOffset = 1 - ((i >= 3) and 1 or 0)
+            local hOffset = -(i % 3-1)
+            local x = gridX + hOffset
+            local y = gridY + vOffset
+
             local row = grid[x]
             if not row then continue end
 
-            for y = gridY, gridY + 1 do   -- Get every neighbouring cell (including the actual cell) (y axis)
-                if x == gridX and y == gridY then continue end -- Exclude the actual cell as it needs a separate computation
-                local cell = row[y]
-                if not cell then continue end -- Check that values exist
+            local cell = row[y]
+            if not cell then continue end
 
-                -- Check if every player can talk
-                for _, ply2 in ipairs(cell) do
-                    local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
-                        (ply1Pos:DistToSqr(plyPos[ply2]) < voiceDistance and -- voiceradius is on and the two are within hearing distance
-                            (not dynv or IsInRoom(ply1EyePos, eyePos[ply2], ply2))) -- Dynamic voice is on and players are in the same room
+            for _, ply2 in ipairs(cell) do
+                local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
+                    (ply1Pos:DistToSqr(plyPos[ply2]) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                        (not dynv or IsInRoom(ply1EyePos, eyePos[ply2], ply2))) -- Dynamic voice is on and players are in the same room
 
-                    DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
-                    DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
-                end
+                DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
+                DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
             end
         end
     end
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -329,8 +329,8 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
 
 
     -- Clear old values
-    plyToGrid[0] = {}
     plyToGrid[1] = {}
+    plyToGrid[2] = {}
     grid = {}
 
     -- Get the grid position of every player O(N)
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -333,9 +333,12 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
     plyToGrid[2] = {}
     grid = {}
 
+    local plyPos = {}
+
     -- Get the grid position of every player O(N)
     for _, ply in ipairs(players) do
         local pos = ply:GetPos()
+        plyPos[ply] = pos
         local x = floor(pos.x / gridSize)
         local y = floor(pos.y / gridSize)
 
@@ -369,8 +372,8 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
 
                 -- Check if every player can talk
                 for _, ply2 in ipairs(cell) do
-                    ply1Pos = ply1Pos or ply1:GetPos()
-                    local ply2Pos = ply2:GetPos()
+                    ply1Pos = ply1Pos or plyPos[ply1]
+                    local ply2Pos = plyPos[ply2]
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
                         (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                             (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room
@@ -390,7 +393,7 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
             for i = 1, count do
                 for j = i + 1, count do
                     local ply1, ply2 = cell[i], cell[j]
-                    local ply1Pos, ply2Pos = ply1:GetPos(), ply2:GetPos()
+                    local ply1Pos, ply2Pos = plyPos[ply1], plyPos[ply2]
 
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
                         (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -392,8 +392,9 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
         for _, cell in pairs(row) do
             local count = #cell
             for i = 1, count do
+                local ply1 = cell[i]
                 for j = i + 1, count do
-                    local ply1, ply2 = cell[i], cell[j]
+                    local ply2 = cell[j]
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
                         (plyPos[ply1]:DistToSqr(plyPos[ply2]) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                             (not dynv or IsInRoom(eyePos[ply1], eyePos[ply2], ply2))) -- Dynamic voice is on and players are in the same room
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -304,6 +304,7 @@ local vrad = GM.Config.voiceradius
 local dynv = GM.Config.dynamicvoice
 local deadv = GM.Config.deadvoice
 local voiceDistance = GM.Config.voiceDistance * GM.Config.voiceDistance
+local timeDelay = GM.Config.voiceCheckTimeDelay
 local DrpCanHear = {}
 
 -- Recreate DrpCanHear after Lua Refresh
@@ -312,29 +313,38 @@ for _, ply in pairs(player.GetHumans()) do
     DrpCanHear[ply] = {}
 end
 
-local gridSize = GM.Config.voiceDistance*2 -- Grid cell size is equal to the size of the diamater of player talking
+local gridSize = GM.Config.voiceDistance * 2 -- Grid cell size is equal to the size of the diamater of player talking
 local floor = math.floor -- Caching floor as we will need to use it a lot
 
 -- Grid based position check
-timer.Create("DarkRPCanHearPlayersVoice", 1, 0, function()
-    DrpCanHear = {}
+local grid
+local plyToGrid = {
+    {},
+    {}
+}
+
+
+timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
     local players = player.GetHumans()
 
-    -- Create Grid
-    local grid = {}
-    local plyToGrid = {
-        {},
-        {}
-    }
+
+    -- Clear old values
+    plyToGrid[0] = {}
+    plyToGrid[1] = {}
+    grid = {}
 
     -- Get the grid position of every player O(N)
     for _, ply in ipairs(players) do
-        local pos = ply:GetShootPos()
-        local x = floor(pos.x/gridSize)
-        local y = floor(pos.y/gridSize)
-        grid[x] = grid[x] or {}
-        grid[x][y] = grid[x][y] or {}
-        grid[x][y][#grid[x][y] + 1] = ply
+        local pos = ply:GetPos()
+        local x = floor(pos.x / gridSize)
+        local y = floor(pos.y / gridSize)
+
+        local row = grid[x] or {}
+            local cell = row[y] or {}
+                table.insert(cell, ply)
+            row[y] = cell
+        grid[x] = row
+
         plyToGrid[1][ply] = x
         plyToGrid[2][ply] = y
 
@@ -348,22 +358,22 @@ timer.Create("DarkRPCanHearPlayersVoice", 1, 0, function()
         local gridY = plyToGrid[2][ply1]
         local ply1ShootPos
 
-        for x = gridX, gridX+1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
+        for x = gridX, gridX + 1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
             local row = grid[x]
             if not row then continue end
 
-            for y = gridY, gridY+1 do   -- Get every neighbouring cell (including the actual cell) (y axis)
+            for y = gridY, gridY + 1 do   -- Get every neighbouring cell (including the actual cell) (y axis)
                 if x == gridX and y == gridY then continue end -- Exclude the actual cell as it needs a separate computation
                 local cell = row[y]
                 if not cell then continue end -- Check that values exist
 
                 -- Check if every player can talk
                 for _, ply2 in ipairs(cell) do
-                    ply1ShootPos = ply1ShootPos or ply1:GetShootPos()
-                    local ply2ShootPos = ply2:GetShootPos()
+                    ply1Pos = ply1Pos or ply1:GetPos()
+                    local ply2Pos = ply2:GetPos()
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
-                        (ply1ShootPos:DistToSqr(ply2ShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
-                            (not dynv or IsInRoom(ply1ShootPos, ply2ShootPos, ply2))) -- Dynamic voice is on and players are in the same room
+                        (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                            (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room
 
                     DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
                     DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
@@ -375,31 +385,32 @@ timer.Create("DarkRPCanHearPlayersVoice", 1, 0, function()
     -- Doing a pass-through inside every cell to compute the interactions inside of the cells.
     -- Each grid check is O(N(N+1)/2) where N is the number of players inside the cell.
     for _, row in pairs(grid) do
-        for _, cellPlayers in pairs(row) do
-            local count = #cellPlayers
+        for _, cell in pairs(row) do
+            local count = #cell
             for i = 1, count do
                 for j = i + 1, count do
-                    local ply1 = cellPlayers[i]
-                    local ply2 = cellPlayers[j]
-                    local ply1ShootPos = ply1:GetShootPos()
-                    local ply2ShootPos = ply2:GetShootPos()
+                    ply1Pos = ply1Pos or ply1:GetPos()
+                    local ply2Pos = ply2:GetPos()
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
-                        (ply1ShootPos:DistToSqr(ply2ShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
-                            (not dynv or IsInRoom(ply1ShootPos, ply2ShootPos, ply2))) -- Dynamic voice is on and players are in the same room
+                        (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                            (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room
 
                     DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
-                    DrpCanHear[ply2] = DrpCanHear[ply2] or {}
-                    DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive())
+                    DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
                 end
             end
         end
     end
 end)
 
+hook.Add("PlayerDisconnect", "DarkRPCanHear", function(ply)
+    DrpCanHear[ply] = nil -- Clear to avoid memory leaks
+end)
+
 function GM:PlayerCanHearPlayersVoice(listener, talker)
     if not deadv and not talker:Alive() then return false end
 
-    return DrpCanHear[listener][talker], threed
+    return (not vrad) or DrpCanHear[listener][talker] == true, threed
 end
 
 function GM:CanTool(ply, trace, mode)
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -356,7 +356,7 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
     for _, ply1 in ipairs(players) do
         local gridX = plyToGrid[1][ply1]
         local gridY = plyToGrid[2][ply1]
-        local ply1ShootPos
+        local ply1Pos
 
         for x = gridX, gridX + 1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
             local row = grid[x]
@@ -389,8 +389,9 @@ timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
             local count = #cell
             for i = 1, count do
                 for j = i + 1, count do
-                    ply1Pos = ply1Pos or ply1:GetPos()
-                    local ply2Pos = ply2:GetPos()
+                    local ply1, ply2 = cell[i], cell[j]
+                    local ply1Pos, ply2Pos = ply1:GetPos(), ply2:GetPos()
+
                     local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
                         (ply1ShootPos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                             (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
     DrpCanHear[ply] = {}
 end
 
--- proxy function to take load from PlayerCanHearPlayersVoice, which is called a quadratic amount of times per tick,
--- causing a lagfest when there are many players
-local function calcPlyCanHearPlayerVoice(listener)
-    if not IsValid(listener) then return end
-    DrpCanHear[listener] = DrpCanHear[listener] or {}
-    if listener:IsBot() then return end
-    local shootPos = listener:GetShootPos()
-    for _, talker in ipairs(player.GetHumans()) do
-        local talkerShootPos = talker:GetShootPos()
-        DrpCanHear[listener][talker] = not vrad or -- Voiceradius is off, everyone can hear everyone
-            (shootPos:DistToSqr(talkerShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
-                (not dynv or IsInRoom(shootPos, talkerShootPos, talker))) -- Dynamic voice is on and players are in the same room
-    end
-end
-hook.Add("PlayerInitialSpawn", "DarkRPCanHearVoice", function(ply)
-    calcPlyCanHearPlayerVoice(ply)
-    if ply:IsBot() then return end
-    timer.Create(ply:UserID() .. "DarkRPCanHearPlayersVoice", 0.5, 0, fn.Curry(calcPlyCanHearPlayerVoice, 2)(ply))
-end)
+local gridSize = GM.Config.voiceDistance*2 -- Grid cell size is equal to the size of the diamater of player talking
+local floor = math.floor -- Caching floor as we will need to use it a lot
 
-hook.Add("PlayerDisconnected", "DarkRPCanHearVoice", function(ply)
-    DrpCanHear[ply] = nil
-    if ply:IsBot() then return end
-    for _, v in ipairs(player.GetHumans()) do
-        if not DrpCanHear[v] then continue end
-        DrpCanHear[v][ply] = nil
+-- Grid based position check
+timer.Create("DarkRPCanHearPlayersVoice", 1, 0, function()
+    DrpCanHear = {}
+    local players = player.GetHumans()
+
+    -- Create Grid
+    local grid = {}
+    local plyToGrid = {
+        {},
+        {}
+    }
+
+    -- Get the grid position of every player O(N)
+    for _, ply in ipairs(players) do
+        local pos = ply:GetShootPos()
+        local x = floor(pos.x/gridSize)
+        local y = floor(pos.y/gridSize)
+        grid[x] = grid[x] or {}
+        grid[x][y] = grid[x][y] or {}
+        grid[x][y][#grid[x][y] + 1] = ply
+        plyToGrid[1][ply] = x
+        plyToGrid[2][ply] = y
+
+        DrpCanHear[ply] = {} -- Initialize output variable
     end
-    timer.Remove(ply:UserID() .. "DarkRPCanHearPlayersVoice")
+
+    -- Check all neighbouring cells for every player.
+    -- We are only checking in 1 direction to avoid duplicate check of cells
+    for _, ply1 in ipairs(players) do
+        local gridX = plyToGrid[1][ply1]
+        local gridY = plyToGrid[2][ply1]
+        local ply1ShootPos
+
+        for x = gridX, gridX+1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
+            local row = grid[x]
+            if not row then continue end
+
+            for y = gridY, gridY+1 do   -- Get every neighbouring cell (including the actual cell) (y axis)
+                if x == gridX and y == gridY then continue end -- Exclude the actual cell as it needs a separate computation
+                local cell = row[y]
+                if not cell then continue end -- Check that values exist
+
+                -- Check if every player can talk
+                for _, ply2 in ipairs(cell) do
+                    ply1ShootPos = ply1ShootPos or ply1:GetShootPos()
+                    local ply2ShootPos = ply2:GetShootPos()
+                    local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
+                        (ply1ShootPos:DistToSqr(ply2ShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                            (not dynv or IsInRoom(ply1ShootPos, ply2ShootPos, ply2))) -- Dynamic voice is on and players are in the same room
+
+                    DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
+                    DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
+                end
+            end
+        end
+    end
+
+    -- Doing a pass-through inside every cell to compute the interactions inside of the cells.
+    -- Each grid check is O(N(N+1)/2) where N is the number of players inside the cell.
+    for _, row in pairs(grid) do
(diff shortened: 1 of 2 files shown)


gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
...
-hook.Add("PlayerInitialSpawn", "DarkRPCanHearVoice", function(ply)
-    calcPlyCanHearPlayerVoice(ply)
-    if ply:IsBot() then return end
-    timer.Create(ply:UserID() .. "DarkRPCanHearPlayersVoice", 0.5, 0, fn.Curry(calcPlyCanHearPlayerVoice, 2)(ply))
-end)
+local gridSize = GM.Config.voiceDistance*2 -- Grid cell size is equal to the size of the diamater of player talking
+local floor = math.floor -- Caching floor as we will need to use it a lot
+
+-- Grid based position check
+timer.Create("DarkRPCanHearPlayersVoice", 1, 0, function()
```
FPtje, line 319: If this is indeed significantly more efficient in practice, we might want to lower this from 1 second to like 0.3 or something. I think this 1 second delay is actually noticeable for players. Someone racing by with their microphone on, and you either don't hear them, or for a full second.

FPtje (changes requested):
This optimization is absolutely wonderful. It attacks a real bottleneck, and provides an elegant solution. It reminds me of an exercise in university, where a similar solution was needed to optimize a purposely badly coded simulation of tanks shooting at one another.

There are quite a few comments, but the algorithm looks sound to me. Great work!

Some questions:

- What tests have you run to make sure this works (and also in edge cases)
- Have you measured the difference? You could temporarily change `player.GetHumans()` with `player.GetAll()`, start a local server with a ton of bots and test away. I'm actually really excited to know what the speed difference is between the two. I reckon it can get quite big, especially with the maximum of 128 players.


Large server owners will be particularly happy about this. IIRC they have a problem with this very code you optimised.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
...
+            if not row then continue end
+
+            for y = gridY, gridY+1 do   -- Get every neighbouring cell (including the actual cell) (y axis)
+                if x == gridX and y == gridY then continue end -- Exclude the actual cell as it needs a separate computation
+                local cell = row[y]
+                if not cell then continue end -- Check that values exist
+
+                -- Check if every player can talk
+                for _, ply2 in ipairs(cell) do
+                    ply1ShootPos = ply1ShootPos or ply1:GetShootPos()
```
Kefta, line 362: I don't think the ShootPos should be used for this since it's based on the model attachment - GetPos would be more suitable.

gamemode/modules/base/sv_gamemode_functions.lua
Kefta: The time should just be a config variable.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
...
+        {},
+        {}
+    }
+
+    -- Get the grid position of every player O(N)
+    for _, ply in ipairs(players) do
+        local pos = ply:GetShootPos()
+        local x = floor(pos.x/gridSize)
+        local y = floor(pos.y/gridSize)
+        grid[x] = grid[x] or {}
```
Kefta, line 335: You should localise table accesses (you can reuse the same local if the previous value can be discarded).

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
...
+        grid[x][y] = grid[x][y] or {}
+        grid[x][y][#grid[x][y] + 1] = ply
+        plyToGrid[1][ply] = x
+        plyToGrid[2][ply] = y
+
+        DrpCanHear[ply] = {} -- Initialize output variable
```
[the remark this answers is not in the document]
JurisBog: According to [this stackoverflow suggestion](https://stackoverflow.com/questions/16239563/is-it-better-to-set-a-table-to-empty-or-set-all-elements-of-a-table-to-nil) it is better to clear keys the keys. I am not sure if there is a function, but I am happy to loop over all keys with `pairs`

gamemode/modules/base/sv_gamemode_functions.lua
JurisBog: > If this is indeed significantly more efficient in practice, we might want to lower this from 1 second to like 0.3 or something. I think this 1 second delay is actually noticeable for players. Someone racing by with their microphone on, and you either don't hear them, or for a full second.

That is my bad, it used to be 0.5 seconds.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
...
+    }
+
+    -- Get the grid position of every player O(N)
+    for _, ply in ipairs(players) do
+        local pos = ply:GetShootPos()
+        local x = floor(pos.x/gridSize)
+        local y = floor(pos.y/gridSize)
+        grid[x] = grid[x] or {}
+        grid[x][y] = grid[x][y] or {}
+        grid[x][y][#grid[x][y] + 1] = ply
```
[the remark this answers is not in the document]
FPtje: Only if you specify a position. Otherwise it just inserts at the end, which is O(1). Though I suppose calling the function and having it do the check is more expensive than manually adding to the table.

gamemode/modules/base/sv_gamemode_functions.lua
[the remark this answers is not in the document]
FPtje: There is [`table.Empty`](https://wiki.facepunch.com/gmod/table.Empty). I thought it was a LuaJIT primop, but it's just a for loop.

Im not sure if it's better or worse in this specific situation. It might need some measurements.

gamemode/modules/base/sv_gamemode_functions.lua
JurisBog: I am not quite sure what you mean by this.

JurisBog:
> Large server owners will be particularly happy about this. IIRC they have a problem with this very code you optimised.
That's exactly what got me into it. :)

> What tests have you run to make sure this works (and also in edge cases)
No explicit tests have been done, but it currently runs on my server with 80-ish+ people right now.

> Have you measured the difference?
I will get back to you on that one, but I am not sure how to benchmark beyond simply using FProfile. If I could get a hand with that, I'd appreciate it greatly!

gamemode/modules/base/sv_gamemode_functions.lua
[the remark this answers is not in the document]
JurisBog: It also mentions that this is only true when there are 32+ entries, which I think is unlikely for a single person to check against 32 other people.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,34 +312,88 @@ for _, ply in pairs(player.GetHumans()) do
...
+                    local ply1 = cellPlayers[i]
+                    local ply2 = cellPlayers[j]
+                    local ply1ShootPos = ply1:GetShootPos()
+                    local ply2ShootPos = ply2:GetShootPos()
+                    local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
+                        (ply1ShootPos:DistToSqr(ply2ShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
+                            (not dynv or IsInRoom(ply1ShootPos, ply2ShootPos, ply2))) -- Dynamic voice is on and players are in the same room
+
+                    DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
+                    DrpCanHear[ply2] = DrpCanHear[ply2] or {}
```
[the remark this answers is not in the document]
JurisBog: Sure would be!

gamemode/modules/base/sv_gamemode_functions.lua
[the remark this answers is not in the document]
Kefta: LuaJIT inlines it if you don't specify a position. It's faster than actually adding it manually unless you already have the length cached.

gamemode/modules/base/sv_gamemode_functions.lua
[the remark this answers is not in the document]
Kefta: table.Empty loops thru the table to empty it, it's faster to recreate the table.

gamemode/modules/base/sv_gamemode_functions.lua
Kefta: ``grid[x] = grid[x] or {}``
``grid[x][y] = grid[x][y] or {}``

You can store grid[x] to a local so that you can index it just once/twice instead of three times.

mcNuggets1:
Can we have some data about this?
Like statistics comparing the old version's runtime speed vs the new versions runtime speed

gamemode/modules/base/sv_gamemode_functions.lua
FPtje: Isn't GetPos at players' feet? Voices will be blocked by the tiniest of differences in height if that's the case.

Kefta: > Isn't GetPos at players' feet? Voices will be blocked by the tiniest of differences in height if that's the case.

It is, but it's comparing the distance between player's feet which is in a consistent position. GetShootPos can change to a large degree depending on the model and even more importantly, the rotation of the player. At least GetPos will never disqualify a player from hearing another based on them turning away from another.

gamemode/modules/base/sv_gamemode_functions.lua
FPtje: > The time should just be a config variable.

Why? It's an implementation detail, this config variable would _very_ likely only confuse people. It's difficult to understand what it means, and even more difficult what the consequences will be if you set it too low or too high. This should be a value that's just good for everyone. If this code is fast enough, setting it to a low value should work for everyone, even for servers with the maximum amount of players.

Kapppa: As if a server wishes to change the value because it is causing an issue or something. They have to change the core gamemode which is highly un-recommended. This leads them to if they update and forget to change the value, they may think the update of DarkRP broke something and not use it. (I have seen this happen) By making it a config and they change it, every update will use the config value causing less issues.

FPtje: The entire point is that it won't be causing issues if it's set to the right value. People will tinker with config values without the slightest clue about what they mean. Then they suddenly notice that voice doesn't work, and I get complaints about voice chat being broken. I'll then have no clue whatsoever as to why the voice would be broken, because there are no error messages.

Having a config value for this will actively cause people to break voice, and I will get the issues. It is not happening.

Kefta: Explain what the config variable does in the comment. This sort of hardcoded refresh rate is just going to lead people who DO know what they're doing to alter core files. For example, on 128 player servers, a 0.5 to even 1 second refresh rate might be preferred while at 32 players, 0.1 seconds would be. There is no reason to not provide power user features because it would "confuse" people when the simple solution is to explain what the config variable does in a comment. You could even add "-- DO NOT ALTER THIS UNLESS YOU KNOW WHAT YOU'RE DOING!"

FPtje: If the value needs to be dependent on player count, then that should be fixed by DarkRP itself, not by demanding server hosters know the intricacies of DarkRP.

Although with this PR the performance will be more dependent on how many players are near one another. If everyone is in hearing range of one another (e.g. Through noclip), you'll still get bad performance. Do you think people "who know what they're doing" should fill in this value for how often they expect many players to be near one another?

> than it is their fault for not reading.

Sure, but I'll be left not knowing it's their fault because in a year I'll have forgotten there's a config option to either completely break voice chat or slow the server to a crawl.

> lead people who DO know what they're doing to alter core files

People who know what they're doing will hopefully be people who recognise a problem in DarkRP itself. Any issue or pull request they create will be much preferred over the ones who break shit through not knowing what a config variable does, causing a problem that is extremely difficult to trace back to the source. 

The point is, no one should want to edit that value, if they do, it's either their incompetence or a bug in DarkRP. Config options are not for working around bugs.

Kefta: People have different needs. Some servers tend to have a lot of people near each other a lot (events) and thus would want a greater value. Some have more serious RP servers and thus only have a handful of people near each other at a time, thus they'd want the value as small as possible (maybe even 0) to be accurate. It's not a situation of there being one best value, but rather giving users the option for the gamemode to be versatile. Let users who know what they're doing decide whether they want accuracy or efficiency.

FPtje: With "people have different needs" one can argue for spending the rest of their lives adding config options. Why not have one for picking the algorithm itself? Maybe for really small servers the simple O(n^2) algorithm is faster. Why not have a config option for using GetShootPos or GetPos in the calculation? Why not a config option for the name of the timer? A config option for the names of the hooks? 

The line on what should be a config value and what shouldn't be  might be vague, but some things clearly should be, and others clearly shouldn't. This to me firmly falls in the latter category.

> a lot of people near each other often (events)

If darkrp is a performance bottleneck in this case, then the responsibility for that lies with DarkRP. A config option here is a statement of inability to write performance code, and a lazy excuse towards anyone who has performance problems with DarkRP. 

This config option is harmfuld for everyone. For people who don't know what they're doing it'll break the game. For the people who do know what they're doing it's a slap in the face, an insult. "Go capture your server's behavior by twiddling with a single float."

This is a PR that greatly optimises the performance of this feature. Your insistence on having a config option implies a mistrust in its performance gain. We can run some tests, I reckon 0.1 or 0.3 might be good for a 128 player server. We can then merge it and see if people start having trouble with many players near another. If it does, we can find new clever tricks like the author of this PR did here. Let's not go for a dirty cop out.

Kefta: A hardcoded number like this seems like config 101, but I give up. Let's see if the creator ever comes back to make the corrections needed.

JurisBog: Sorry for the delay.

I personally stand with the config option, entirely because those are left at specific defaults pre-defined by the gamemode. We get best of both worlds as DarkRP the default setting should be a good value to be used in most cases, and those with more intricate knowledge can update the value to whatever they think is better.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,29 +313,38 @@ for _, ply in pairs(player.GetHumans()) do
...
-        grid[x][y][#grid[x][y] + 1] = ply
+        local pos = ply:GetPos()
+        local x = floor(pos.x / gridSize)
+        local y = floor(pos.y / gridSize)
+
+        local row = grid[x] or {}
+            local cell = row[y] or {}
+                table.insert(cell, ply)
+            row[y] = cell
+        grid[x] = row
```
Kapppa, line 346: Why is there weird indentation?

JurisBog: I personally find it easier to read, don't know how others feel about it.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,40 +313,104 @@ for _, ply in pairs(player.GetHumans()) do
...
+
+        DrpCanHear[ply] = {} -- Initialize output variable
+    end
+
+    -- Check all neighbouring cells for every player.
+    -- We are only checking in 1 direction to avoid duplicate check of cells
+    for _, ply1 in ipairs(players) do
+        local gridX = plyToGrid[1][ply1]
+        local gridY = plyToGrid[2][ply1]
+        local ply1ShootPos
```
FPtje, line 359: `ply1ShootPos` is declared and used, but not defined

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,40 +313,104 @@ for _, ply in pairs(player.GetHumans()) do
...
+
+    -- Doing a pass-through inside every cell to compute the interactions inside of the cells.
+    -- Each grid check is O(N(N+1)/2) where N is the number of players inside the cell.
+    for _, row in pairs(grid) do
+        for _, cell in pairs(row) do
+            local count = #cell
+            for i = 1, count do
+                for j = i + 1, count do
+                    ply1Pos = ply1Pos or ply1:GetPos()
+                    local ply2Pos = ply2:GetPos()
```
FPtje, line 393: `ply1` and `ply2` here are not defined

FPtje:
Here's some raw data for the timings.
[Timings voice chat export.xlsx](https://github.com/FPtje/DarkRP/files/5940089/Timings.voice.chat.export.xlsx)

I've modified the code a bit, created the following two functions in `sv_gamemode_functions.lua`:

<details>
<summary>Functions code</summary>

```lua
function oldCalcVoice()
    for _, listener in pairs(player.GetAll()) do
        if not IsValid(listener) then return end
        DrpCanHear[listener] = DrpCanHear[listener] or {}
        local shootPos = listener:GetShootPos()
        for _, talker in ipairs(player.GetAll()) do
            local talkerShootPos = talker:GetShootPos()
            DrpCanHear[listener][talker] = not vrad or -- Voiceradius is off, everyone can hear everyone
                (shootPos:DistToSqr(talkerShootPos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                    (not dynv or IsInRoom(shootPos, talkerShootPos, talker))) -- Dynamic voice is on and players are in the same room
        end
    end
end

local gridSize = GM.Config.voiceDistance * 2 -- Grid cell size is equal to the size of the diamater of player talking
local floor = math.floor -- Caching floor as we will need to use it a lot

-- Grid based position check
local grid
local plyToGrid = {
    {},
    {}
}

function newCalcVoice()
    local players = player.GetAll()


    -- Clear old values
    plyToGrid[0] = {}
    plyToGrid[1] = {}
    grid = {}

    -- Get the grid position of every player O(N)
    for _, ply in ipairs(players) do
        local pos = ply:GetPos()
        local x = floor(pos.x / gridSize)
        local y = floor(pos.y / gridSize)

        local row = grid[x] or {}
            local cell = row[y] or {}
                table.insert(cell, ply)
            row[y] = cell
        grid[x] = row

        plyToGrid[1][ply] = x
        plyToGrid[2][ply] = y

        DrpCanHear[ply] = {} -- Initialize output variable
    end

    -- Check all neighbouring cells for every player.
    -- We are only checking in 1 direction to avoid duplicate check of cells
    for _, ply1 in ipairs(players) do
        local gridX = plyToGrid[1][ply1]
        local gridY = plyToGrid[2][ply1]
        local ply1Pos

        for x = gridX, gridX + 1 do   -- Get every neighbouring cell (including the actual cell) (x axis)
            local row = grid[x]
            if not row then continue end

            for y = gridY, gridY + 1 do   -- Get every neighbouring cell (including the actual cell) (y axis)
                if x == gridX and y == gridY then continue end -- Exclude the actual cell as it needs a separate computation
                local cell = row[y]
                if not cell then continue end -- Check that values exist

                -- Check if every player can talk
                for _, ply2 in ipairs(cell) do
                    ply1Pos = ply1Pos or ply1:GetPos()
                    local ply2Pos = ply2:GetPos()
                    local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
                        (ply1Pos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                            (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room

                    DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
                    DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
                end
            end
        end
    end

    -- Doing a pass-through inside every cell to compute the interactions inside of the cells.
    -- Each grid check is O(N(N+1)/2) where N is the number of players inside the cell.
    for _, row in pairs(grid) do
        for _, cell in pairs(row) do
            local count = #cell
            for i = 1, count do
                local ply1 = cell[i]
                for j = i + 1, count do
                    local ply2 = cell[j]
                    ply1Pos = ply1Pos or ply1:GetPos()
                    local ply2Pos = ply2:GetPos()
                    local canTalk = not vrad or -- Voiceradius is off, everyone can hear everyone
                        (ply1Pos:DistToSqr(ply2Pos) < voiceDistance and -- voiceradius is on and the two are within hearing distance
                            (not dynv or IsInRoom(ply1Pos, ply2Pos, ply2))) -- Dynamic voice is on and players are in the same room

                    DrpCanHear[ply1][ply2] = canTalk and (deadv or ply2:Alive())
                    DrpCanHear[ply2][ply1] = canTalk and (deadv or ply1:Alive()) -- Take advantage of the symmetry
                end
            end
        end
    end
end

```

</details>

Some changes with the original code:

- Changed `player.GetHumans` to `player.GetAll`, because I'm testing with bots
- Quickly fixed the two minor mistakes mentioned above

Some notes:

- "Spread" means very few if not none of all the players were within hearing range of one another, so at most 1 or 2 players per cell
- "nearby" was lazy, I just used `FAdmin tp *`. FAdmin has an algorithm to make sure people don't get placed into one another, so not everyone will be in the same cell or in hearing range of every other player. Placing every player in exactly the same spot is a bit _too_ artificial a test.
- The benchmark was a quick `lua_run for i = 1, 10 do local a = SysTime() for j = 1, 10000 do old/newCalcVoice() end print(SysTime() - a) end`


Then tested with as many iterations as I found feasible. With few players that's 100,000 iterations, but with 128 players I had to go down to 100 👀.

I haven't done many statistics on the results, but it's clear from a glance that the difference is absolutely wonderful when the players are spread. When all players are nearby, it looks like the new algorithm is a constant factor slower than the old algorithm (6-10 times?). Emphasis on _looks like_, some more statistics are needed to actually confirm that.

JurisBog:
Not sure how great of an idea this is - but a potential solution would be to add a "sparsity" threshold (either number of cells created or max players per cell) below which the grid is avoided for X number of seconds/iterations?

gamemode/modules/base/sv_gamemode_functions.lua
Kefta: I think that sort of indentation should only be used for scopes and stack operations, it looks weird in this context.

FPtje:
It could be, but the old algorithm has terrible performance in the same situation as well. I wonder if there are some clever solutions for that specific situation. Maybe we can do something with the knowledge/assumption that players are near one another when in the same cell. 

Either way I think this decrease in performance in this very specific situation is not a blocker for merging the PR. By far in most situations it will be much better. Only in some very specific cases it's a bit worse.

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,40 +313,105 @@ for _, ply in pairs(player.GetHumans()) do
...
+
+
+    -- Clear old values
+    plyToGrid[0] = {}
+    plyToGrid[1] = {}
+    grid = {}
+
+    -- Get the grid position of every player O(N)
+    for _, ply in ipairs(players) do
+        local pos = ply:GetPos()
```
Kapppa, line 338: As you calculate the player position at the beginning, you can just cache it. This will mean you can index it later instead of calling the position loads of times. The only thing I can think that is a downside is that it will be a bit outdated 
by only a minuscule amount.

Kefta: They could have teleported in that time. It should just stay as is.

mcNuggets1:
Checking GetPos will result in players not hearing one and another if there is a minimal slope.
EyePos would be a better variant.

TomDotBat:
> EyePos would be a better variant.

Agreed, using EyePos would actually add realism as it's closer to the player's ears.

Kefta:
> > EyePos would be a better variant.
> 
> Agreed, using EyePos would actually add realism as it's closer to the player's ears.

The problem with anything but GetPos is that it's rotation dependent which is not ideal.

gamemode/modules/base/sv_gamemode_functions.lua
JurisBog: I presume you mean to cache it through each execution of this function. I believe that the entire concept works on the assumption are 'roughly' in the same area for 0.3+ seconds in which case, the minor change for caching positions is negligible.

FPtje:
> The problem with anything but GetPos is that it's rotation dependent which is not ideal.

Perhaps we can use GetPos for the distance check, EyePos for the trace that checks whether there is a wall in between (`IsInRoom`).

gamemode/modules/base/sv_gamemode_functions.lua
FPtje: If the positions are cached at the start of the function, and used later in the same call tree, players can not be teleported. I presume @Kapppa is talking about the uses of GetPos on lines 373 and 393.

Dannelor:
I may be wrong, but does this implementation cover the case where two players are in the upper left cell or bottom right cell from each other?
Also, would a hexagon grid be beneficial in this case rather than a square grid? After working through a rough version, I think it might be able to reduce the area that will be checked for each cell, at the cost of some extra preprocessing. I'd be happy to talk over what I've found regarding it.

Kapppa (changes requested):

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,40 +313,105 @@ for _, ply in pairs(player.GetHumans()) do
...
+}
+
+
+timer.Create("DarkRPCanHearPlayersVoice", timeDelay, 0, function()
+    local players = player.GetHumans()
+
+
+    -- Clear old values
+    plyToGrid[0] = {}
+    plyToGrid[1] = {}
```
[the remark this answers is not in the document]
Kefta: LuaJIT will still compile 0-starting tables the same, but it should probably still be changed for consistency.

JurisBog:
> I may be wrong, but does this implementation cover the case where two players are in the upper left cell or bottom right cell from each other?
> Also, would a hexagon grid be beneficial in this case rather than a square grid? After working through a rough version, I think it might be able to reduce the area that will be checked for each cell, at the cost of some extra preprocessing. I'd be happy to talk over what I've found regarding it.

I believe so, it should be in lines 361-366. I am not entirely sure how the hexagonal grid would improve things so please do tell!

gamemode/modules/base/sv_gamemode_functions.lua
```diff
@@ -312,40 +313,107 @@ for _, ply in pairs(player.GetHumans()) do
...
+    end
+
+    -- Doing a pass-through inside every cell to compute the interactions inside of the cells.
+    -- Each grid check is O(N(N+1)/2) where N is the number of players inside the cell.
+    for _, row in pairs(grid) do
+        for _, cell in pairs(row) do
+            local count = #cell
+            for i = 1, count do
+                for j = i + 1, count do
+                    local ply1, ply2 = cell[i], cell[j]
```
Kapppa, line 396: As @FPtje said, you can move the ply1 down a scope. This will save a lot of indexing.

Dannelor:
> 
> 
> > I may be wrong, but does this implementation cover the case where two players are in the upper left cell or bottom right cell from each other?
> > Also, would a hexagon grid be beneficial in this case rather than a square grid? After working through a rough version, I think it might be able to reduce the area that will be checked for each cell, at the cost of some extra preprocessing. I'd be happy to talk over what I've found regarding it.
> 
> I believe so, it should be in lines 361-366. I am not entirely sure how the hexagonal grid would improve things so please do tell!

I did some debugging on it, and I think there may be a dead-zone in 2 of the corners, I've attached some images where I set the players/bots color based on if they were being checked within the cell. Two of the corners do not show detection for either player or bot and I believe would be skipped over in the cell checks.

Also verifying by doing PrintTable(DrpCanHear) in PlayerCanHearPlayersVoice shows that they are not able to hear each other.

<details>
<summary> Images examples </summary>

bot_detects
corner
corner_2
player_detects

</details>

JurisBog:
> I did some debugging on it, and I think there may be a dead-zone in 2 of the corners, I've attached some images where I set the players/bots color based on if they were being checked within the cell. Two of the corners do not show detection for either player or bot and I believe would be skipped over in the cell checks.
> 
> Also verifying by doing PrintTable(DrpCanHear) in PlayerCanHearPlayersVoice shows that they are not able to hear each other.

I really like the tool and visualisation you provided! Would you be able to share it so I can do some further testing and identify the issue? From my understanding, it should be processed for either of the players (whichever one is on the 'bottom-right' corner)

>Regarding using a hexagon grid, doing some quick napkin math if our voice radius is length 1 unit and we need to check 5 squares of area we'd have to check a total area 20 unit^2, for a grid of hexagons with Apothem of 1 unit (Area 3.464 unit^2) we would need to check 4 hexagons cells with a total area of 13.856 unit^2. This does require some extra initial steps to convert from world coordinates to hex-grid coordinates though and I haven't tested the difference in performance due to this.

Could we get some visuals for the working out, please?

Dannelor:
Yeah I think there just needs to be 1 extra cell check

And here's the code I used for the visual, and I'll get back to you on the working out for the hexagon grid.
<details>

<summary> Visual Code</summary>

This can be thrown into any clientside file

Show box around cell
```
local talk = GM.Config.voiceDistance * 2
hook.Add("PostDrawOpaqueRenderables","testing",function ()
    local pos = LocalPlayer():GetPos()
    local x = math.floor(pos.x / talk) 
    local y = math.floor(pos.y / talk)

    render.DrawWireframeBox(Vector(x * talk,y * talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
end)
```

In sv_gamemode_function add these and change code to check for bots

Color Players/Bots when checking neighbor cells

```
Line 340: ply:SetMaterial("models/debug/debugwhite")
               ply:SetColor(Color(255,255,255))

line 383: ply2:SetMaterial("models/debug/debugwhite")
              ply2:SetColor(Color(0,255,0))   

```

</details>

Dannelor:
> Could we get some visuals for the working out, please?

Here is a quick write-up of my findings regarding the use of a hexagonal grid

We can skip testing 3 of the 7 tiles due to checking all players, this is similarly implemented in the square grid.

<details>

<summary> Hexagon Tile skipping duplicates</summary>

Tiles with blue are checked by yellow tiles, and tiles with green are checked by the center tile.
If players are in neighboring tiles, we can safely skip 3 contiguous sides as it will be checked by another player.
This leaves only 4 total hexagons that need to be eventually checked.

hexagon_removing_duplicates

We can show that on a hexagon grid a player will not be able to have their voice be in unchecked tiles

</details>

Player Voice distance compared to checked locations in hexagon grid

<details>

<summary> Hexagon Player Voice radius </summary>

For any player with voice radius distance r on a hexagon grid with Apothem distance of r, the below image shows that we do not intersect any tiles that are not within tiles that will eventually be checked due to above rule.

hexagon_player_radius

</details>


We also decrease the total surface area checked by the cells algorithm by using hexagon tiles

<details>

<summary> Decreased validation area on hexagon grid vs square grid </summary>

The image below shows an overlayed square grid vs hexagon grid with equal Apothem, the white areas are hexagon cells that will be checked.
The Red area shows area that will be checked by a square grid, but not a hexagon grid.
The blue area onward is unchecked by both.

Should both square grid and hexagon grid have Apothem of 1 Unit we can say:
Area of square = 4 unit^2
Area of hexagon = 3.464 unit^2

% of area =  (3.464 unit^2 * 4 hexagons) / (4 unit^2 * 5 squares) = 69% (nice)

hexagon_square_overlay

</details>

The formulas for converting to hexagon tile coordinates is linear in time

<details>

<summary> World to Hex Conversion </summary>

These are borrowed from the hexagonal grids [Webpage](https://www.redblobgames.com/grids/hexagons/#pixel-to-hex)
These use functions for simplicity, but we can rewrite the code for efficiency and many of the computed constants can be cached

This code would replace the current implementation for finding square grids position from player coordinates.

```
function pixel_to_pointy_hex(point):
    var q = (sqrt(3)/3 * point.x  -  1./3 * point.y) / size
    var r = (                        2./3 * point.y) / size
    return hex_round(Hex(q, r))

function cube_round(cube):
    var rx = round(cube.x)
    var ry = round(cube.y)
    var rz = round(cube.z)

    var x_diff = abs(rx - cube.x)
    var y_diff = abs(ry - cube.y)
    var z_diff = abs(rz - cube.z)

    if x_diff > y_diff and x_diff > z_diff:
        rx = -ry-rz
    else if y_diff > z_diff:
        ry = -rx-rz
    else:
        rz = -rx-ry

    return Cube(rx, ry, rz)

function hex_round(hex):
    return cube_to_axial(cube_round(axial_to_cube(hex)))
```

</details>

Once we have the Axial coordinates the rest of the code remains nearly unchanged as we use only a q and r values to determine the players hex grid position.

When iterating over neighbor cells, it is necessary to use cached constant offset coordinates to compute the location of each neighbor such as '[[0, -1], [1, -1], [1, 0]]'

JurisBog:
> Yeah I think there just needs to be 1 extra cell check

You were absolutely right, there are 8 cells all around a given cell. I was checking 6 of them, skipping 'top-right'

After adding an extra cell to check, this should be resolved.

JurisBog:
> Here is a quick write-up of my findings regarding the use of a hexagonal grid

This is a brilliant write-up. Given that you have already started on the maths side of it, I wondered what the consequences of increasing the number of sides are. Let's say we switch from hexagon to octagons. Would we enjoy the same benefit?

Furthermore,  could we approximate a circle using some finer resolution, so that we only need to iterate over some cells around a given player without distance checking therfore reducing the complexity to O(N) (in which case it is likely to come with a hefty constant).

TomDotBat:
> Let's say we switch from hexagon to octagons. Would we enjoy the same benefit?

Surely this wouldn't work as octagons don't tesselate?

JurisBog:
Yeah, you're right - my bad, rookie mistake.

FPtje:
This discussion is absolutely amazing. @Dannelor thanks a lot for the write-up. It is detailed, clearly communicates the benefits and it makes me enthusiastic. It sounds like a great next step to make. I wonder what difference there would be between square grid and hexagon grid.

For now I think the best approach is finishing up the square grid approach, merging that and continuing with hexagon in a new branch. I'll create an issue 👍

mcNuggets1:
For further optimisations, one could cache the GetPos, EyePos function itself.

Like this
`
local reg = debug.getregistry()
local EyePos = reg.Player.EyePos

usage: EyePos(Player)
`

Kefta:
It's best to not do that at the moment. It's a minor optimisation that I'm working on getting LuaJIT to recognise for all code to be affected.

Kapppa:
> For further optimisations, one could cache the GetPos, EyePos function itself.
> 
> Like this
> 
> local reg = debug.getregistry()
> local EyePos = reg.Player.EyePos
> 
> usage: EyePos(Player)

I think using this would be nicer as this returns what you did internally, that being if we were going to do it.
```
local pMeta = FindMetaTable('Player')
local EyePos = pMeta.EyePos
```

Kapppa:
I have been doing some testing and here is the information I have found out.

- The most expensive check is the IsInRoom check.
- IsInRoom uses util.TraceLine
- util.TraceLine doesn't use Lua

So if we reduce the amount of data we check in the TraceLine the check will be faster. Currently we only filter the current player we are checking, but we only return false if trace hit the world. This means any Entity we hit will cause the TraceLine to add that as a hit but we don't care about them as they are not the world. So what I did was change `IsInRoom(eyePos[ply1], eyePos[ply2], ply2)` to `IsInRoom(eyePos[ply1], eyePos[ply2], players)` as util.TraceLine can take in a table as the filter. This means it will ignore all players in the trace meaning less hits in the trace.

I believe this to be one of the causes of the long execution time that @FPtje found in his nearby testing. As it reduced the execution time is half when testing with 32 players (31 bots) and only by 0.5% when with 16 players (15 bots) if @FPtje could test using the changes I suggested, it would be nice to see the impact it would have.

FPtje:
That `IsInRoom` is the most expensive is something I had expected. Traces are much more expensive than squared distance checks. 

Though instead of filtering out all players, wouldn't it be better to set the right [trace mask](https://wiki.facepunch.com/gmod/Enums/MASK#MASK_NPCSOLID)?

Also, if the trace can currently hit entities, would that mean one can hear another player through walls when they put a prop in between?

Kapppa:
I thought about using the Mask, but I have never used them before. So I used the filter as I knew it worked. 

And about the Prop thing, possibly if it stops the check as soon as it hits something.

Dannelor:
> 
> 
> That `IsInRoom` is the most expensive is something I had expected. Traces are much more expensive than squared distance checks.
> 
> Though instead of filtering out all players, wouldn't it be better to set the right [trace mask](https://wiki.facepunch.com/gmod/Enums/MASK#MASK_NPCSOLID)?
> 
> Also, if the trace can currently hit entities, would that mean one can hear another player through walls when they put a prop in between?

Yeah I agree on this. It looks like MASK_SOLID is the default option for traces, and it looks like there are some CONTENTS_* flags that might be better being taken out or replaced with better options?

<details>

<summary> CONTENTS_* flags used </summary>

These 2 seem necessary to handle the majority of IsInRoom cases and will ignore hitting players
CONTENTS_SOLID
CONTENTS_MOVEABLE

These 3 I'm unsure about for checking
CONTENTS_WINDOW: Should players be able to hear through glass? I'm partial to yes because of yelling.
CONTENTS_MONSTER: Currently it looks like the trace will stop on NPC's and return true without actually checking if there is world in-between. After removing this it is not checking on Bots, but I'm not sure if it works on "real" players?
CONTENTS_GRATE: Should players be able to talk through a grate? I'm partial to yes, as this includes anything a bullets and sight can pass through it.

</details>

After looking I think the CONTENTS_SOLID | CONTENTS_MOVEABLE may work well for this case. If someone else if able to verify?

Kapppa:
I have done some more testing and have found out by adding `tracedata.collisiongroup = COLLISION_GROUP_WORLD` to the tracedata makes it faster. Using a mask either didn't change the performance or made it worse. So changing the collision group it changed from 

(1 player, 63 bots, 100 executions repeated 10 times)
With Mask as MASK_SOLID
```
4.584387824
4.565724747
4.553101728
4.551065472
4.548138535
4.568526015
4.561934656
4.561743877
4.572804425
4.561016255
```

With CollisionGroup as COLLISION_GROUP_WORLD
```
1.973841018
1.910615148
1.910703852
1.884380241
1.899916767
1.903296116
1.916958618
1.911684587
1.911032594
1.862566451
```

Adding both of them together made the performance worse than just having the CollisionGroup. What I have noticed is that I get marginally worse execution time than @FPtje . This is tested on an i9-10900k

WilliamVenner:
Trace table micro optimisation

https://github.com/JurisBog/DarkRP/pull/1

FPtje:
As I understand it correctly, we only really need the extra cell check right?

Is there anything else needed for this PR before merging?

Kapppa:
> As I understand it correctly, we only really need the extra cell check right?
> 
> Is there anything else needed for this PR before merging?

The use of the collision group as well.

FPtje:
> The use of the collision group as well.

I think it's best to consider that tangential and optional. I'm fine with it being included or not included in this PR. If it's not included I'll look at it :+1:

JurisBog:
Here's a small snippet I used for testing that 4 out of the 8 surrounding cells are checked:
```lua
local gridX = 0
local gridY = 0
for i = 0, 3 do
    local vOffset = 1 - ((i >= 3) and 1 or 0)
    local hOffset = -(i % 3-1)
    local x = gridX + hOffset
    local y = gridY + vOffset
    print(x, y)
end
```

JurisBog:
I believe I am done with these square checks. Is there anything that I may have missed?

FPtje (approved):

FPtje:
Looks good! I've merged the changes! 🎉🎉🎉🥳  

Thanks a lot @JurisBog for this contribution. This really improves the performance in DarkRP. 

Thanks also to @Kapppa, @Dannelor, @WilliamVenner and everyone else contributing to this pull request.
I've made some changes. Notably, the config value has been changed to `DarkRP.voiceCheckTimeDelay`, to communicate that it is not configuration, but an implementation detail. Lua scripters who know what they're doing will know how to set that value. 

-----

Also, I just tried @Kapppa's idea to set the CollisionGroup of the trace. I tested it with one bot, with both a wall in between and no wall in between:

<details>
<summary>Timing results</summary>

Code run with `lua_run`:

```lua
falco = Player(2) botPos=Player(3):GetShootPos() falcoPos=falco:GetShootPos() f = DarkRP.debugVars.IsInRoom for j = 1, 10 do local t = SysTime() for i = 1, 100000 do f(botPos, falcoPos, falco) end print(SysTime() - t) end
```

```
Old - no wall in between:
0.74880469999994
0.70692809999991
0.70216959999993
0.72021880000011
0.70153789999995
0.71913530000006
0.7229797
0.74114089999989
0.73185189999981
0.70470569999998

Old - wall in between:
0.61419940000019
0.6012581
0.62747059999992
0.66948000000002
0.69400759999985
0.64819780000016
0.6823010999999
0.65127380000013
0.64967309999997
0.69032259999994

Set only collision group - no wall in between:
0.49964839999984
0.50850620000006
0.50281969999992
0.49940029999993
0.50132059999987
0.52350100000012
0.49857689999999
0.49584129999994
0.4971336000001
0.49809559999994

Set only collision group - wall in between:
0.51606149999998
0.49375639999994
0.50140240000019
0.49682759999996
0.50382420000005
0.50317199999995
0.49283840000021
0.48513509999998
0.5004507000001
0.49951329999999

Set collision group and mask - no wall in between:
0.51532970000017
0.46681089999993
0.46502109999983
0.46688540000014
0.4706927999996
0.46574349999992
0.50935089999984
0.47376329999997
0.4737296000003
0.47510740000007

Set collision group and mask - wall in between:
0.52643949999992
0.47864630000004
0.49079540000002
0.47870490000014
0.48870259999967
0.46867680000014
0.47434730000032
0.47267160000001
0.48432390000016
0.47931389999985
```


</details>

Looks like setting the collision group is indeed faster.

Interestingly, I could reproduce the "exploit" that players can hear one another through walls when putting a prop in between, both with `COLLISION_GROUP_WORLD` and without. However, when also setting the mask, the exploit disappeared. It also doesn't seem to have a significant effect on performance.

FPtje:
I was wondering, `gridSize` is set to the diameter of the voice radius, rather than the radius:
https://github.com/FPtje/DarkRP/blob/df96dd6e00ca70de291a63a9184d19942acc574b/gamemode/modules/base/sv_gamemode_functions.lua#L318-L318

Why is that? Wouldn't setting it to the radius work as well?

Here are some screenshots with the difference:

<details>
<summary>Diameter</summary>

hl2_2021-03-13_19-20-57


</details>

<details>



<summary>Radius</summary>

hl2_2021-03-13_19-20-40


</details>

<details>
<summary>Script used</summary>

```lua
local talk = GAMEMODE.Config.voiceDistance * 2
hook.Add("PostDrawOpaqueRenderables","testing",function ()
    local pos = LocalPlayer():GetPos()
    local x = math.floor(pos.x / talk)
    local y = math.floor(pos.y / talk)

    render.DrawWireframeBox(Vector(x * talk,y * talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk + talk,y * talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk,y * talk + talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk + talk,y * talk + talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk,y * talk - talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk - talk,y * talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk - talk,y * talk - talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk + talk,y * talk - talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)
    render.DrawWireframeBox(Vector(x * talk - talk,y * talk + talk,pos.z), Angle(0,0,0), Vector(0,0,-50), Vector(talk,talk,50), Color(255,255,255), true)

    render.DrawWireframeSphere(pos, GAMEMODE.Config.voiceDistance, 100, 100, Color(255, 0, 0, 255), true)
end)

```

</details>

By setting it to the radius, you're still 100% sure that the players you can hear are either in your own cell or a neighboring one. Halving the grid size has a huge effect on the area it has to check, making false checks far less likely. Am I missing something?

Dannelor:
Yeah I think this may be a good option, just a few things I want to mention on it.

There are edge cases that pop-up for players on boundaries that would depend on how the algorithm determines the square they're in. If two players on the the boundaries of a square 1R apart there are points that they could be in range of each other, but also 2 boxes apart. I think we can safely ignore these points though for an extremely tiny error margin.

Also one difference that pops up is now the player can have their voice within all neighboring cells at the same time, I don't think this changes anything though, but might be a good idea to research and make sure if it hasn't been looked into already.

FPtje:
On the first point,source units are really small, like an inch or so. Besides, with box side length equals voice radius, a player two boxes away won't be in range, even when standing on the border 👍

On the second point, that will always happen when you stand in a corner of a cell I reckon, regardless of its size, right?

JurisBog:
This definitely looks like a good way of doing it! Looks like I was playing
it a bit too safe

Dannelor:
Yeah I think my edge cases would only come up under extremely specific circumstances and since the current algorithm will have a bias towards one direction if a player is perfectly on a boundary it shouldn't be an issue.

For the second one, with the diameter length the worst case is a player has their voice visible within 3 neighbors at once. 
For the radius one, a player at the center has their radius within all neighbors simultaneously.
I don't think this changes anything since it handles duplicate checks already, but wanted to note it down since it is different from previous.

JurisBog:
>Also one difference that pops up is now the player can have their voice within all neighboring cells at the same time, I don't think this changes anything though, but might be a good idea to research and make sure if it hasn't been looked into already.

Currently, all of the surrounding cells are checked against one another.

JurisBog:
Here is an image of how the existing check is decided. Red is "active" cell and green ones are checked. You can convince yourself by exhaustion that if any of the remaining surrounding cells become "active" they will include the red cell in their checks.

image

FPtje:
Alright! The `gridSize` is now set to the voice radius rather than the diameter. With the size halved, the area it checks should be divided by 4, generally catching much fewer players in unnecessary checks.

Thanks for the feedback!

WilliamVenner:
Is the `filter` necessary if we are using `MASK_SOLID_BRUSHONLY`?

https://github.com/FPtje/DarkRP/blob/2d96d4a493bdf596b920873b8afc020ef283254d/gamemode/modules/base/sv_gamemode_functions.lua#L297-L299

Kefta:
It is the rarest of cases, but players could technically have a brush model which would be caught by the mask.

WilliamVenner:
Wow I've never heard of that, do you have an example? Why would they even have a brush model?

Kefta:
It requires manually setting the player's model to a brush ("*number"). I do not know if this even works anymore but there's no harm in having a filter.

Kapppa:
I dev for a large "Roleplay" server, we have ran into issues with the current mask selected as thing you used to be able to talk through are no longer possible. For example, we have a prop that is transparent(acting like glass) in a building that blocks the lobby and receptionist. The current mask prevented the two people hearing each other. We had to change to `MASK_VISIBLE` as it only did stuff that blocked the LOS. It may allow for voice to go through props but its better to hear when you shouldn't than when should be able to hear and you can't. We are also not using the filter.

FlorianLeChat:
> For example, we have a prop that is transparent(acting like glass) in a building that blocks the lobby and receptionist.

It's not only the glasses by the way, but also the fences and other elements where the bullets can pass are currently considered as completely solid elements and block voice communications (it's a rare problem but it happens sometimes). By adding [CONTENTS_GRATE](https://wiki.facepunch.com/gmod/Enums/CONTENTS#CONTENTS_GRATE) to the mask, the problem seems to be fixed with no negative effect, but I don't think it's a durable solution to this kind of problem.

PR #3087, opened 2021-01-24
Create GitHub workflow to run glualint on the DarkRP repository

FPtje:
This is mostly a test PR to see what happens when there are warnings/errors in a PR.

The given warnings are from files that should be ignored, because they are in the ignore list of `.glualint.json`. Due to a bug in glualint the warnings are thrown anyways. That has been fixed in the latest master of glualint, but for now I'll use it to test warnings.

### DIFF
--- a/.github/workflows/run-glualint.yml
+++ b/.github/workflows/run-glualint.yml
@@ -1,0 +1,14 @@
+name: run-glualint
+on:
+  push:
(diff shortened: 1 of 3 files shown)


FPtje:
CC @saibotk, it's thanks to his work that this has become easy.

PR #3090, opened 2021-01-28
Change the behavior with the "FAdmin_StoreBan" hook

FlorianLeChat:
Currently, it is possible to return a value in the `FAdmin_StoreBan` hook, but this is totally useless because it has no effect. With these changes it will be possible to stop the ban from being saved. However, I wonder if this is not too much since the introduction of the `FAdmin_CanBan` hook, considering that `FAdmin_StoreBan` is used for the registration and not the "**action**" of the ban.
If you consider this change to be unnecessary then you should simply remove the return value from the `FAdmin_StoreBan` hook.


FPtje:
I think this is a good fix. Thanks!

PR #3093, opened 2021-02-04
An even smarter solution

MyDimeIsUp:
Smarter than #3078. Allows for more sounds to be easily added in case someone wants to do that.


Kefta:
Move the sounds table to a config var then.

FPtje:
The config var sounds like a good idea, though it can be seen separately from this PR. Thanks for the improvement!

PR #3094, opened 2021-02-09
Allow for instant job respawning

Down-s:
I see some custom servers do this, would make things a lot easier if we could just do this with one config option.

### DIFF
--- a/gamemode/modules/jobs/sv_jobs.lua
+++ b/gamemode/modules/jobs/sv_jobs.lua
@@ -137,6 +137,8 @@ function meta:changeTeam(t, force, suppressNotification, ignoreMaxMembers)
         DarkRP.resetLaws()
     end
 
+    local DoEffect = false
+
     self:SetTeam(t)
     hook.Call("OnPlayerChangedTeam", GAMEMODE, self, prevTeam, t)
     DarkRP.log(self:Nick() .. " (" .. self:SteamID() .. ") changed to " .. team.GetName(t), nil, Color(100, 0, 255))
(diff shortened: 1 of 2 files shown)


gamemode/modules/jobs/sv_jobs.lua
```diff
@@ -156,7 +156,21 @@ function meta:changeTeam(t, force, suppressNotification, ignoreMaxMembers)
         gamemode.Call("PlayerSetModel", self)
         gamemode.Call("PlayerLoadout", self)
     else
-        self:KillSilent()
+        if GAMEMODE.Config.instantjob then
```
FPtje, line 159: Could you add this config option to the `config.lua` file?

gamemode/modules/jobs/sv_jobs.lua
```diff
@@ -156,7 +156,21 @@ function meta:changeTeam(t, force, suppressNotification, ignoreMaxMembers)
...
+            local vPoint = self:GetShootPos() + Vector(0,0,50)
+            local effectdata = EffectData()
+            effectdata:SetEntity(self)
+            effectdata:SetStart(vPoint)
+            effectdata:SetOrigin(vPoint)
+            effectdata:SetScale(1)
+            util.Effect("entity_remove", effectdata)
```
FPtje, line 166: This code duplicates the effect code of the `norespawn` if-branch. Could you unify the duplicate code?

FPtje (changes requested):
Sensibe addition, thanks! Some comments.

Down-s:
Sure, I'll do that now

gamemode/modules/jobs/sv_jobs.lua
```diff
@@ -137,26 +137,41 @@ function meta:changeTeam(t, force, suppressNotification, ignoreMaxMembers)
...
+
+            self:StripWeapons()
+            self:RemoveAllAmmo()
+            self:Spawn()
+        else
+            self:KillSilent()
+        end
+    end
+
+    if DoEffect then
```
FPtje, line 167: Wouldn't this effect be run _after_ the player has spawned on line 161? The effect would appear at the player's spawn position, rather than where they were.

Down-s: I made it like that on purpose as I feel it looks better spawning in with the effect, I can change it if you like.

FPtje (approved):
LGTM! Thanks!

Issue #3005, opened 2020-04-09
The old usermessage library is bad for optimization.


AlternativeGM:
Sorry no form (phone).
Create a separate DarkRP release with the replacement of the usermessages library with the net library or make a choice between methods in the config. Usermessages break down optimization on weak computers.

P.S. Kind of like you are forbidden to edit the source files in DarkRP. Sorry for using a translator.

Kefta:
This has already been discussed ad nauseum in https://github.com/FPtje/DarkRP/pull/1988 and many other issues. Garry's Mod will eventually change the implementation of umsg to net internally, but you can just use a umsg->net Lua script.

Bo98:
> Usermessages break down optimization on weak computers.

If there is a particular case which is having such a serious impact on people, let us know.

umsg isn't even used in the most heavily invoked parts of DarkRP networking - like DarkRPVars or weapons.

AlternativeGM:
> > Usermessages break down optimization on weak computers.
> 
> If there is a particular case which is having such a serious impact on people, let us know.
> 
> umsg isn't even used in the most heavily invoked parts of DarkRP networking - like DarkRPVars or weapons.

Notification... 
Yep...
*censored* notifications!

Bo98:
Notifications are the hardest to change without breaking backwards compatibility. Because the implementation is in the HUD module, many custom HUDs copy-pasted it because people just disable the HUD module entirely rather than using `HUDShouldDraw`. (Actual customisation of how it looks is typically done via overriding the GMod's own documented functions & hooks.)

matiasg90:
I am using the github version and my server never stops laging.
https://i.gyazo.com/0ef04afae758964adabf1a0de3aa9520.png

when installing the workshop version January 22, magically the error went away, I suppose we are talking about the same

FPtje:
At one point I did change notifications to net. I swiftly reverted it when I got tons of complaints of addons breaking.

It's not going to happen, unless gmod changes the implemention to net, maybe.

AlternativeGM:
> At one point I did change notifications to net. I swiftly reverted it when I got tons of complaints of addons breaking.
> 
> It's not going to happen, unless gmod changes the implemention to net, maybe.

I can edit core files? I need a net libraly.

FPtje:
Why do think you need that?
You can disable modules and recreate them in the darkrp mod.

AlternativeGM:
Synchronization of addon and game mode.
[...]

PR #3095, opened 2021-02-09
Switch notifications to use net messages & add config option for notification sound

Down-s:
I noticed that the notification system used the deprecated umsg system. I switched this out for the net message system and added a config option for changing the notification sound.

### DIFF
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -323,6 +323,9 @@ GM.Config.chatCommandPrefix = "/"
 GM.Config.F1MenuHelpPage = "https://darkrp.miraheze.org/wiki/Main_Page"
 GM.Config.F1MenuHelpPageTitle = "DarkRP Wiki"
 
+-- The sound that plays when you get a DarkRP notification
+GM.Config.NotifSound = "buttons/lightswitch2.wav"
+
 -- Put Steam ID's and ranks in this list, and the players will have that rank when they join.
 GM.Config.DefaultPlayerGroups = {
     ["STEAM_0:0:00000000"] = "superadmin",
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -324,7 +324,7 @@ GM.Config.F1MenuHelpPage = "https://darkrp.miraheze.org/wiki/Main_Page"
 GM.Config.F1MenuHelpPageTitle = "DarkRP Wiki"
 
 -- The sound that plays when you get a DarkRP notification
-GM.Config.NotifSound = "buttons/lightswitch2.wav"
+GM.Config.notificationSound = "buttons/lightswitch2.wav"
 
 -- Put Steam ID's and ranks in this list, and the players will have that rank when they join.
 GM.Config.DefaultPlayerGroups = {
--- a/gamemode/modules/base/sv_util.lua
+++ b/gamemode/modules/base/sv_util.lua
@@ -1,3 +1,4 @@
+util.AddNetworkString("_Notify")
 function DarkRP.notify(ply, msgtype, len, msg)
     if not istable(ply) then
         if not IsValid(ply) then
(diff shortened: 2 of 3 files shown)


FPtje:
Sorry, the notification system is knowingly kept on `umsgs` because external addons depend on it. I actually changed it to net messages before, only to revert it because of the large amount of addons that started breaking.

TomDotBat:
Usermessages should have been removed years ago, DarkRP is probably the main reason why they weren't. I think both usermessages and the net library should be supported for now, and then slowly start forcing people to use the net library as time goes on. If you give everyone a reasonable amount of time to convert their addons it should work fine.

FPtje:
Most of those addons aren't maintained anymore. The downside of breaking many DarkRP servers clearly outweigh the benefits of moving to net messages. I know umsgs have lower size limits, but we're not hitting them. Umsgs work perfectly fine for DarkRP notifications.

This is typically something where Lua devs argue that umsgs are bad and should always be replaced, but server owners will start grabbing pitch forks when we actually go ahead and replace them. 

Previous issues:
https://github.com/FPtje/DarkRP/issues/3005
https://github.com/FPtje/DarkRP/pull/2983
https://github.com/FPtje/DarkRP/pull/1988

TomDotBat:
> This is typically something where Lua devs argue that umsgs are bad and should always be replaced, but server owners will start grabbing pitch forks when we actually go ahead and replace them.

That's why we don't replace them, at least not instantly.
By avoiding the deprecation warnings we're essentially halting the progression of the game's development.

Down-s:
Couldn't we just send both a umsg and net?

TomDotBat:
> Couldn't we just send both a umsg and net?


That's exactly what I just mentioned:
> I think both usermessages and the net library should be supported for now, and then slowly start forcing people to use the net library as time goes on.

FPtje:
This PR is an instant replacement in my book.

> By avoiding the deprecation warnings we're essentially halting the progression of the game's development.

I find this hard to believe. How would the presence of some deprecated functions halt progression of development? Are GMod devs seriously bound and chained by the inability to remove some functions?

Honestly, hating on usermessages seems more like a long running meme than something that has actual noticeable effect on servers and/or players. I know net messages are superior in pretty much every way, but when existing, unmaintained addons depend on it, switching might still not be the way to go.

> Couldn't we just send both a umsg and net?

We could have done so in 2015, maybe earlier, and still "finishing" that transition now would break about as many addons as it would with this sudden replacement. Only the addons since that point _might_ have started using the net messages if they didn't copy other people's code. There is no incentive for developers to switch, especially when they stopped maintaining their addons altogether.

Down-s:
Hopefully it doesn't sound too dumb, but could we just override umsg and check if they are trying to access _Notify

FPtje:
That sounds really dirty, but I like the idea. We should try to make it somewhat efficient, though. Good thinking!

TomDotBat:
>This PR is an instant replacement in my book.

It is, and this is clearly not the solution to the problem.

>I find this hard to believe. How would the presence of some deprecated functions halt progression of development? Are GMod devs seriously bound and chained by the inability to remove some functions?

When I say it's halting the progression of the game's development, I'm regarding the modernisation of community made content. No one should be required to use a library that was marked deprecated several years ago, and so that's why adding net support **alongside** usermessages is probably the best way we can start this transition.

FlorianLeChat:
> Hopefully it doesn't sound too dumb, but could we just override umsg and check if they are trying to access _Notify

I think it would be the best way at this point to have an instant and totally transparent transition. Unfortunately, people would still continue to copy code containing deprecated libraries..

Down-s:
> Unfortunately, people would still continue to copy code containing deprecated libraries.
Some sort of warning could be added to let developers know they are using deprecated shit

FPtje:
> No one should be required to use a library that was marked deprecated several years ago

Overriding the DarkRP notify functions has always been the best way to go for a long time. By listening to the umsgs they depend on an implementation detail that we now can't change because they did.

> Unfortunately, people would still continue to copy code containing deprecated libraries..

That's fine by me, we can keep that workaround around forever.

> Some sort of warning could be added to let developers know they are using deprecated shit

Those warnings might not ever reach the developers who stopped playing gmod years ago, only the server owners that still run the scripts. :thinking:

TomDotBat:
> Those warnings might not ever reach the developers who stopped playing gmod years ago, only the server owners that still run the scripts. 🤔

They might not, but that certainly doesn't stop server owners from mentioning it to people who can fix it for them. I'm not saying it will happen, but I'm sure there are plenty of people in the community who are willing to revamp the scripts that show these warnings.

Kapppa:
You could have a look at this if you want the gamemode to use the Net Library. 

https://github.com/Kapppa/UMsg-to-Net

gamemode/config/config.lua
```diff
@@ -323,6 +323,9 @@ GM.Config.chatCommandPrefix = "/"
 GM.Config.F1MenuHelpPage = "https://darkrp.miraheze.org/wiki/Main_Page"
 GM.Config.F1MenuHelpPageTitle = "DarkRP Wiki"
 
+-- The sound that plays when you get a DarkRP notification
+GM.Config.NotifSound = "buttons/lightswitch2.wav"
```
FPtje, line 327: Could you change this to `GM.Config.notificationSound`? The capitalization is a bit of a mess, but I think this would be most consistent with the other options.

FPtje (changes requested):
For this PR the change from umsg to net should be reverted. We can then look at the other idea in a follow-up issue. The notification sound is a good idea, one minor comment about the name.👍

gamemode/config/config.lua
Down-s: Done 👍

FPtje (changes requested):
The net message conversion is still there. Could you revert that?

PR #3101, opened 2021-02-16
gamemode/init.lua optimisations

BullyHunter32:
Replaced pairs for ipairs where appropriate and cached 'SortedPairs' as it is a frequently used yet expensive function


mcNuggets1:
Ah yes, the initial 0.02 ms save on loading the map.
Instead of doing such minor optimisations in several commits, wouldn't 1 big commit with micro optimisations rather help?

Don't get me wrong, I'm all about micro optimisations, but this is marginal.

PR #3102, opened 2021-02-17
Allow newer version of tmysql

Pawsative:
If you used a newer version of tmysql from SuperiorServer's github (https://github.com/SuperiorServers/gm_tmysql4), MySQLite.query() would return no data, and tmysql.initialize() is no longer a function so that would error too.

Changed tmysql.initialize to tmysql.Connect, but made tmysql.Connect use tmysql.Initialize if they're using an older version of tmysql.

Newer version of tmysql4 requires you to have tmysql:Poll() in a think hook also (https://github.com/SuperiorServers/dash/blob/master/lua/dash/libraries/server/mysql.lua#L85-L87)

Tested on these versions of tmysql, and working:
https://github.com/SuperiorServers/gm_tmysql4 (Newer version)
https://github.com/bkacjios/gm_tmysql4/ (Older version)

### DIFF
--- a/gamemode/libraries/mysqlite/mysqlite.lua
+++ b/gamemode/libraries/mysqlite/mysqlite.lua
@@ -98,7 +98,6 @@
     DatabaseInitialized
         Called when a successful connection to the database has been made.
 ]]
-
 local debug = debug
 local error = error
 local ErrorNoHalt = ErrorNoHalt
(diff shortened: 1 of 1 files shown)


Pawsative:
Oh, I'll add this to the mysqlite git instead

PR #3103, opened 2021-02-17
sink 20% of the money dropped on death

7aninhos:
with the intention of reducing inflation

not tested, I wasn't sure if the function could accept "amount * 0.8" directly, it could be a new GM.Config as well

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -469,7 +469,8 @@ function GM:PlayerDeath(ply, weapon, killer)
 
         if amount > 0 then
             ply:addMoney(-amount)
-            DarkRP.createMoneyBag(ply:GetPos(), amount)
+            local death_sink = amount * 0.8
+            DarkRP.createMoneyBag(ply:GetPos(), death_sink)
         end
     end
 


DBotThePony:
what

7aninhos:
> what
> 
> why did you just came straight with this pr, without even a config option. Also, do money really need to vanish into void?

yes, that is the idea of a money sink. I was expecting this to get accepted first in order to do the config option, he might not even be interest on money sinks at all, so why bother

FPtje:
There already is a `GAMEMODE.Config.deathfee` option. While that is constant, you could make it so that values between 0 and 1 turn it into a relative option.

> I was expecting this to get accepted first

It seems like that might just piss a couple of people off.

DBotThePony:
> There already is a `GAMEMODE.Config.deathfee` option. While that is constant, you could make it so that values between 0 and 1 turn it into a relative option.
This pr is about *removing* lsome of lost money from player instead of dropping it all on ground. This should be defined as separate option

FPtje:
That makes sense

7aninhos:
darkrp is doomed to inflation, because money is constantly being created of thin air, so removing some of it from the economy helps keep it's value

MrBrax:
isn't this what the tax thing is for?

7aninhos:
yes, but the more ways the better, so we don't have to remove a lot of money in just one or a few ways; darkrp already has a few ways, but those are far from enough, and mine is better than taxes because the later usually feels like robbery, since you're losing money just because the time is passing. this money sink shouldn't feel like robbery because the person who dies usually doesn't gets his money back, and the one who got the dropped money wouldn't ever know that the amount is changed

FPtje:
That's actually pretty clever, though there are problems with drop money on death in the first place. It really, really endless encourages revenge killing 😬

That's a different story, though. Could you add an option to config.lua, and default to it being inactive?

7aninhos:
sure, it's better to have flexibility on the value, but I don't understand why someone wouldn't use this

just to add, having a bank addon changes things, especially now that arcbank is free, as people could just deposit their money to prevent losing it on death, although I haven't given much thought to revenge killing

MrBrax:
also, isn't the reason for inflation in base darkrp money printers? the first thing i did was disable those and add other ways of earning money. it's hard coming up with more ways, but it sure as hell slows inflation down

7aninhos:
I agree, and yes, money printers and salaries are the biggest reasons of inflation

PR #3104, opened 2021-02-19
Replace #RPExtraTeams by jobCount

JarosLucky:
In the basic essence, in this context, it has an equal role, but what was is an unnecessary expensive action that can be avoided


FoksVHox:
This could be a breaking change, as there are some addon that uses `#RPExtraTeams`

FPtje:
Yeah, I'm not sure why there's a separation between `#RPExtraTeams` and the manual job count, but it might indeed break something if there is job removing in between the job creation calls.

Besides, I wouldn't call a `#RPExtraTeams` unnecessarily expensive.

JarosLucky:
Maybe then we made .team = Team?

пт, 19 февр. 2021 г., 16:52 Falco Peijnenburg <notifications@github.com>:

mcNuggets1:
Didn't you once say. One should use table.Count for RPExtraTeams?

PR #3106, opened 2021-02-21
Use glualint-1.17.2 on ubuntu-20.04

FPtje:
The latest version should fix the libffi.so.6 issue described in the comment.

See https://github.com/FPtje/GLuaFixer/issues/96


FPtje:
Tests succeed. Merging.

PR #3113, opened 2021-03-10
Fix error with almost no information

mcNuggets1:
This fixes an error, when an entity doesn't have the DarkRPItem table for some reason anymore.

It now happens to me and I don't have any clue what causes it, or which entity.

Also with a little mini optimisation included.

### DIFF
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -696,7 +696,9 @@ local function restoreReconnectedEnts(ply)
             e:Setowning_ent(ply)
         end
 
-        ply:addCustomEntity(e.DarkRPItem)
+        if e.DarkRPItem then
+            ply:addCustomEntity(e.DarkRPItem)
+        end
     end
 
     queuedForRemoval[sid] = nil
@@ -969,8 +971,9 @@ local function collectRemoveEntities(ply)
         remClasses[string.lower(customEnt.ent)] = true
     end
 
+    local sid = ply.SID
     for _, v in ipairs(ents.GetAll()) do
-        if v.SID ~= ply.SID or not v:IsVehicle() and not remClasses[string.lower(v:GetClass() or "")] then continue end
+        if v.SID ~= sid or not v:IsVehicle() and not remClasses[string.lower(v:GetClass() or "")] then continue end
 
         table.insert(collect, v)
     end


mcNuggets1:
[ERROR] gamemodes/darkrp/gamemode/modules/base/sv_entityvars.lua:226: attempt to index local 'entTable' (a nil value)
  1. addCustomEntity - gamemodes/darkrp/gamemode/modules/base/sv_entityvars.lua:226
   2. restoreReconnectedEnts - gamemodes/darkrp/gamemode/modules/base/sv_gamemode_functions.lua:699
    3. unknown - gamemodes/darkrp/gamemode/modules/base/sv_gamemode_functions.lua:725

Sadly this error doesn't say anything.
And also it errors for basically no reason.

It assumes, DarkRPItem is set, just because it's in the DarkRP table.
But it can be removed manually and there is purpose to do so.

FPtje:
You just gotta look further than just the direct point of error. Follow me along my path of reasoning:

Starting point: For some entity `e` in `queuedForRemoval[sid]`, `e.DarkRPItem` is `nil`.
- Something must have put an entity into `queuedForRemoval[sid]` with this property.
- `queuedForRemoval[sid]` is set here: https://github.com/FPtje/DarkRP/blob/d6edf40edb59cfaeca9596147c7915975aad8e19/gamemode/modules/base/sv_gamemode_functions.lua#L938-L950, to some `entList`
- that `entList` comes from here: https://github.com/FPtje/DarkRP/blob/d6edf40edb59cfaeca9596147c7915975aad8e19/gamemode/modules/base/sv_gamemode_functions.lua#L995-L996
- and are collected here: https://github.com/FPtje/DarkRP/blob/d6edf40edb59cfaeca9596147c7915975aad8e19/gamemode/modules/base/sv_gamemode_functions.lua#L961-L986
- It looks at `DarkRPEntities` to find ent classes that should be collected for removal. Every such entity of which `ent.SID == ply.SID`, gets put in the table. It also adds a player's lawboards. With a bit of reasoning, you can find that there appatenly is some entity that has an `SID`, but not a `DarkRPEntity`. 
- Maybe it's the lawboards? Maybe there's an entity that can be bought, but also spawned in another way. When it is spawned in another way, it sets the `SID`, but not the `DarkRPItem`.


Now you: When this error happens, run a Lua statement on the server that loops through all entities. If `ent.SID ~= nil and ent.DarkRPItem == nil`, then you've found the culprit. See where it comes from to more fully understand the issue.

mcNuggets1:
It's prop_vehicle_jeep, it's not even an DarkRP entity, nor has anywhere DarkRPItem set to it or is EVER used via addCustomEntity.
Obviously it doesn't have DarkRPItem on it.

So will this PR now then merged?

FPtje:
That's very important information. That makes the difference between hiding an error you don't understand, and accounting for behavior that you now _do_ understand.

PR #3114, opened 2021-03-18
Fix 'Attempt to index camEnd3D2D (A nil value)'

Samsterminator:
Currently camEnd3D2D() is a nil value, which breaks printers


FPtje:
Thanks! This shouldn't have been broken.

CC @BullyHunter32

PR #3119, opened 2021-04-14
Force pitch of healing sound to 25, if below 25.

mcNuggets1:
Otherwise, the sound is crackling and extremely long / ear raping.
It sounds much better this way.


FPtje:
Thanks!

PR #3120, opened 2021-04-21
Color being called on every frame

7aninhos:
seems negligible but will stack up indefinitely since no one cares to fix it

same as the player.GetAll one

### DIFF
--- a/entities/entities/spawned_money/cl_init.lua
+++ b/entities/entities/spawned_money/cl_init.lua
@@ -1,5 +1,8 @@
 include("shared.lua")
 
+local color_1 = Color(140, 0, 0, 100)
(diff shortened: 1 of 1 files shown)


entities/entities/spawned_money/cl_init.lua
```diff
@@ -1,5 +1,8 @@
 include("shared.lua")
 
+local color_1 = Color(140, 0, 0, 100)
```
FPtje, line 3: Could you name the colors? They're dark red and white

7aninhos: done

FPtje: Thanks!

PR #3122, opened 2021-05-06
Reduce the Amount of Color and Vector calls in Draw/Think hooks

Kapppa:
- Store most static Color calls in local var
- Store most static Vector offsets in local var
- Makes some code more readable
- Cached Model call in Meteor

<details>
    <summary>Other Stuff</summary>
    <p>There were more changes but they were either in FAdmin or changing them made the code worse off either by being less readable, worse performance or were rarely used.</p>
    <p>I have some modification for `entities\entities\meteor\init.lua` but they are for another commit </br>I have some modification for `gamemode\modules\chatindicator\cl_init.lua` but I didn't like the changes.</p>
</details>

### DIFF
--- a/entities/entities/darkrp_laws/cl_init.lua
+++ b/entities/entities/darkrp_laws/cl_init.lua
@@ -2,10 +2,16 @@ include("shared.lua")
 
 local Laws = {}
 
+ENT.DrawPos = Vector(1, -111, 58)
+
+local color_navy_200 = Color(0, 0, 70, 200)
+local color_red = Color(255, 0, 0, 255)
+local color_white = color_white
+
 function ENT:Draw()
     self:DrawModel()
 
-    local DrawPos = self:LocalToWorld(Vector(1, -111, 58))
+    local DrawPos = self:LocalToWorld(self.DrawPos)
 
     local DrawAngles = self:GetAngles()
     DrawAngles:RotateAroundAxis(self:GetAngles():Forward(), 90)
@@ -16,14 +22,13 @@ function ENT:Draw()
         surface.SetDrawColor(0, 0, 0, 255)
         surface.DrawRect(0, 0, 558, 290)
 
-        draw.RoundedBox(4, 0, 0, 558, 30, Color(0, 0, 70, 200))
+        draw.RoundedBox(4, 0, 0, 558, 30, color_navy_200)
 
-        draw.DrawNonParsedSimpleText(DarkRP.getPhrase("laws_of_the_land"), "Roboto20", 279, 5, Color(255, 0, 0, 255), TEXT_ALIGN_CENTER)
+        draw.DrawNonParsedSimpleText(DarkRP.getPhrase("laws_of_the_land"), "Roboto20", 279, 5, color_red, TEXT_ALIGN_CENTER)
 
-        local col = Color(255, 255, 255, 255)
         local lastHeight = 0
         for k, v in ipairs(Laws) do
-            draw.DrawNonParsedText(string.format("%u. %s", k, v), "Roboto20", 5, 35 + lastHeight, col)
+            draw.DrawNonParsedText(string.format("%u. %s", k, v), "Roboto20", 5, 35 + lastHeight, color_white)
             lastHeight = lastHeight + (fn.ReverseArgs(string.gsub(v, "\n", "")) + 1) * 21
         end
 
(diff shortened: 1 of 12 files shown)


entities/entities/darkrp_laws/cl_init.lua
```diff
@@ -2,10 +2,16 @@ include("shared.lua")
 
 local Laws = {}
 
+ENT.DrawPos = Vector(1, -111, 58)
+
+local color_navy_200 = Color(0, 0, 70, 200)
+local color_red = Color(255, 0, 0, 255)
+local color_white = color_white
```
FPtje, line 9: This works for me, but where is it defined? I'm not sure we can depend on it.

entities/entities/drug_lab/init.lua
```diff
@@ -24,9 +25,9 @@ function ENT:canUse(activator)
 end
 
 function ENT:createItem(activator)
-    local drugPos = self:GetPos()
+    local drugPos = self:GetPos() + self.SpawnOffset
```
FPtje, line 28: Adding two vectors still creates a new vector, so it's not really an optimisation. I wouldn't make any further steps to optimise it though, the micro change would not be worth the decrease in readability.

This does have the nice effect of making the spawn offset configurable.

entities/entities/meteor/cl_init.lua
```diff
@@ -1,6 +1,8 @@
 ENT.Spawnable = false
 ENT.AdminSpawnable = false
 
+ENT.ParticleModel = Model("particles/smokey")
```
FPtje, line 4: This forces the caching of the model on server startup, adding to the first frame lag that's already there. I think it's better to cache it upon first use, especially since barely anyone ever uses the meteor.

FPtje (changes requested):
Nice work, some comments

entities/entities/darkrp_laws/cl_init.lua
FlorianLeChat: This is defined internally in Gmod: https://wiki.facepunch.com/gmod/Global_Variables#misc

FPtje: Ah, alright 👍

entities/entities/drug_lab/init.lua
mcNuggets1: Serverside functions which are called only once don't need this sort of optimisation.

entities/entities/meteor/cl_init.lua
mcNuggets1: It is automatically cached on use, so removing it, would be more suitable.

entities/entities/drug_lab/init.lua
Kapppa: It is faster and also makes the code more readable in my opinion.

Ran 10000 times as smaller runs gives scientific notation
```
Vector(x,y,z + 30)
0.0072376109999368
Vec + Vec
0.0021952729999839
```

FPtje: That's adding a vector of itself. That might not give the same performance as adding two distinct vectors. Either way, these functions aren't being run 10,000 times, so there is no performance difference.

Like I said, though, damn the `0.0000005` seconds of performance difference for every drug lab creation. Being able to set the variable is nice enough.

Kefta: Vector:Add(Vector) preserves the original object and is readable. They are significantly faster operations than creating a new object which also adds all the vector pool operations.

Kapppa: > That's adding a vector of itself. That might not give the same performance as adding two distinct vectors. Either way, these functions aren't being run 10,000 times, so there is no performance difference.
> 
> Like I said, though, damn the `0.0000005` seconds of performance difference for every drug lab creation. Being able to set the variable is nice enough.

Sorry I should have labeled them better. That is adding two distinct vectors.

```
local vec = Vector(100, 100, 100)
local vec2 = Vector(0,0,30)
local t

print("Vector(x,y,z + 30)")
local time = SysTime()
for i=1,10000 do
	t = Vector(vec.x,vec.y,vec.z + 30)
end
print(SysTime() - time)

print("Vec + Vec")
local time = SysTime()
for i=1,10000 do
	t = vec + vec2
end
print(SysTime() - time)
```

I did this mainly for the fact it had better readability over the previous one, with the added benefit you are able to change the offset. The performance benefits are just a nice side effect.

FPtje (approved):
Nice work! Thanks!

PR #3124, opened 2021-05-17
Allow editing the eating sound of food entities

mcNuggets1:
I want to make custom eating noises and hooking into the creation of a food object and then assigning the new sound might just fit.

### DIFF
--- a/entities/entities/food/init.lua
+++ b/entities/entities/food/init.lua
@@ -45,7 +45,7 @@ function ENT:Use(activator, caller)
     umsg.End()
 
     self:Remove()
-    activator:EmitSound("vo/sandwicheat09.mp3", 100, 100)
+    activator:EmitSound(self.UseSound, 100, 100)
 end
 
 function ENT:OnRemove()
--- a/entities/entities/food/init.lua
+++ b/entities/entities/food/init.lua
@@ -45,7 +45,7 @@ function ENT:Use(activator, caller)
     umsg.End()
 
     self:Remove()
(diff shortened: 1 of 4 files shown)


entities/entities/food/init.lua
```diff
@@ -45,7 +45,7 @@ function ENT:Use(activator, caller)
     umsg.End()
 
     self:Remove()
-    activator:EmitSound("vo/sandwicheat09.mp3", 100, 100)
+    activator:EmitSound(self.UseSound, 100, 100)
```
FPtje, line 48: How does `EatSound` turn into `UseSound`?

FPtje (changes requested):
I appreciate your contributions, but I'd _really_ appreciate it if you test your changes. I know I'm primarily responsible for errors being introduced into DarkRP, but your PRs have introduced errors into DarkRP before. Please test your changes, including simple ones like this one.

entities/entities/food/init.lua
mcNuggets1: Didn't i fix this within a few hours?
I'm running this on my server for days now.
No problems.

mcNuggets1: Ah sorry, just seen it.
I wasn't sure how to name it.

And the food entity isn't used that much on our server, so I wouldn't know.

Fixed it, and again, very sorry.
I hate doing these mistakes.

FPtje: Testing is best done by ensuring the code path is run and watching  the results. In this case it would be spawning some food and trying to eat it, see if it makes a sound.

PR #3125, opened 2021-05-27
Change ents.FindInSphere to player.GetHumans and vector:DistToSqr

Kapppa:
Its 10x faster with a small amount of people. With about 60 people its 5x as fast. Will probably be better for server with lots of entities as it doesn't check them anymore.

### DIFF
--- a/gamemode/modules/base/sv_util.lua
+++ b/gamemode/modules/base/sv_util.lua
@@ -50,12 +50,15 @@ end
 util.AddNetworkString("DarkRP_Chat")
 
 function DarkRP.talkToRange(ply, PlayerName, Message, size)
-    local ents = ents.FindInSphere(ply:EyePos(), size)
+    local ents = player.GetHumans()
     local col = team.GetColor(ply:Team())
     local filter = {}
 
+    local plyPos = ply:EyePos()
+    local sizeSqr = size ^ 2
+
     for _, v in ipairs(ents) do
-        if v:IsPlayer() and not v:IsBot() and (v == ply or hook.Run("PlayerCanSeePlayersChat", PlayerName .. ": " .. Message, false, v, ply) ~= false) then
+        if (v:EyePos():DistToSqr(plyPos) <= sizeSqr) and (v == ply or hook.Run("PlayerCanSeePlayersChat", PlayerName .. ": " .. Message, false, v, ply) ~= false) then
             table.insert(filter, v)
         end
     end


FlorianLeChat:
Could we please have a benchmark to see the difference?

Kapppa:
> Could we please have a benchmark to see the difference?

With 4 bots
```
old version took: 0.795556
new version took: 0.067966000000013
```

With ~70 bots
```
old version took: 0.99436599999999
new version took: 0.277388
```

Kefta:
Should've binded player.FindIn* before I left.

gamemode/modules/base/sv_util.lua
```diff
@@ -50,12 +50,15 @@ end
 util.AddNetworkString("DarkRP_Chat")
 
 function DarkRP.talkToRange(ply, PlayerName, Message, size)
-    local ents = ents.FindInSphere(ply:EyePos(), size)
+    local ents = player.GetHumans()
     local col = team.GetColor(ply:Team())
     local filter = {}
 
+    local plyPos = ply:EyePos()
+    local sizeSqr = size ^ 2
```
FPtje, line 58: I can't believe I'm saying this, but IIRC, `size * size` is faster than `size ^ 2`. This is because power calculations can get difficult to compute if the power component is less trivial than `2`.

FPtje:
In your benchmark, how many times was the function run? I don't suppose a single run with 4 bots took `0.8` seconds, did it? :sweat_smile:

Kapppa:
> In your benchmark, how many times was the function run? I don't suppose a single run with 4 bots took `0.8` seconds, did it? 😅

10000 times and yes that is seconds

gamemode/modules/base/sv_util.lua
Kapppa: size ^ 2 is faster I swear.

mcNuggets1:
How many entities were present?

Kapppa:
> How many entities were present?
> Also how is the performance with player.GetAll?

This was on an empty map with just the world props. So like doors and stuff, so not many. 

This had to be tested with player.GetAll as I was using bots. 

<details><summary>CODE</summary>
<p>

```lua
function check(func, name)
	local timee = SysTime()
	for i=1, 10000 do
		func()
	end
	print(name.." took: "..SysTime() -  timee)
end


local function performance()
    local ply, PlayerName, Message, size = Player(3), "Jimbo", "My great message", 500
	local ents = ents.FindInSphere(ply:EyePos(), size)
    local col = team.GetColor(ply:Team())
    local filter = {}

    for _, v in ipairs(ents) do
        if v:IsPlayer() and (v == ply or hook.Run("PlayerCanSeePlayersChat", PlayerName .. ": " .. Message, false, v, ply) ~= false) then
            table.insert(filter, v)
        end
    end

    return filter
end
check(performance, "old version")

local function performance2()
    local ply, PlayerName, Message, size = Player(3), "Jimbo", "My great message", 500
	
	local ents = player.GetAll()
    local col = team.GetColor(ply:Team())
    local filter = {}
	
    local plyPos = ply:EyePos()
    local sizeSqr = size ^ 2

    for _, v in ipairs(ents) do
        if (v:EyePos():DistToSqr(plyPos) <= sizeSqr) && (v == ply or hook.Run("PlayerCanSeePlayersChat", PlayerName .. ": " .. Message, false, v, ply) ~= false) then
            table.insert(filter, v)
        end
    end

    return filter
end
check(performance2, "new version")
```
</p>
</details>

If you want I can test with lots of props.

gamemode/modules/base/sv_util.lua
Kapppa: > I can't believe I'm saying this, but IIRC, `size * size` is faster than `size ^ 2`. This is because power calculations can get difficult to compute if the power component is less trivial than `2`.

The reason I said it was faster is because I saw these comments on a Git post. https://github.com/Facepunch/garrysmod/pull/1755#issuecomment-773738032

I saw them change x * x to x ^ 2 so I assumed it was faster. But your comments made me test it. These tests had to be ran 1 million times to get out of scientific notation so these results are literally the worst case scenario and any differences are amplified. 

```
number * number took: 0.00029512599576265
number * number took: 0.00027080500149168
number * number took: 0.00027066399343312
number * number took: 0.0002706310042413
number * number took: 0.00028311900678091
number ^ 2 took: 0.00025741501303855
number ^ 2 took: 0.00024428300093859
number ^ 2 took: 0.00024426299205516
number ^ 2 took: 0.00024425800074823
number ^ 2 took: 0.00024425798619632

number * number took: 0.00030000000000996
number * number took: 0.00027100000002633
number * number took: 0.00054099999999835
number * number took: 0.00027099999999791
number * number took: 0.00048899999998753
number ^ 2 took: 0.00028500000001941
number ^ 2 took: 0.00027099999999791
number ^ 2 took: 0.00027900000000614
number ^ 2 took: 0.00027099999999791
number ^ 2 took: 0.00028500000001941
```

Each test was ran 1 million times, so 20 million in total. As you can see number ^ 2 is faster, but only marginally in most cases but sometimes number * number is drastically worse. But that could be something else causing it, as it should be consistent. 

I can change to number * number to keep consistent with the codebase if you would like but 🤷

FlorianLeChat: From what I remember, he didn't change it because it was performing better but it was mathematically wrong. Referring to Spar's benchmark website: https://gitspartv.github.io/LuaJIT-Benchmarks/#test7, the performance is the **same** except if the code is **interpreted and not compiled**, where `n * n` is more efficient than `n ^ 2`. Personally, I generally keep `n * n` instead of `^ 2` or `math.pow`.

FPtje: Looks like LuaJIT has a special case for power 2. If it compiles to the exact same set of instructions then there's not going to be any difference. It looks like `x * x` is indeed better if the code isn't compiled.

Issue #3131, opened 2021-06-23
F4 Team Model Select Tooltip Produces Errors on Mouseover


kackalack:
When selecting a model on a job that has multiple models associated with it I am spammed with errors while the cursor hovers over the icon. It seems to fail filling in the panel of the tooltip displayed for the model path.

[ERROR] lua/skins/default.lua:1014: attempt to call field 'Tooltip' (a nil value)
1. SkinHook - lua/skins/default.lua:1014
2. unknown - lua/vgui/dtooltip.lua:94 (x1475)

In order to produce this bug, just hover your cursor over the icon of any model of a team in the f4 jobs menu.

I can replicate this error on a fresh install with no other modifications.  Darkrp seems to be the only one responsible for this bug.

FlorianLeChat:
I can also confirm that. However I think it's related to a recent update of Garry's Mod because it's too much for a bug that has been around for a long time without any reports.

FPtje:
Fixed by #3132

PR #3132, opened 2021-06-24
Fix tooltip errors with the DarkRP derma skin

FlorianLeChat:
This is a possible fix for the issue #3131.

Here is the in-game result:
image


FPtje:
Brilliant. Thanks!

PR #3135, opened 2021-07-01
Add support for item skin on f4 menu

ceifa:
I'm new at darkrp, idk if I have do doc it somewhere, but it's a feature I'd like to see.


FPtje (approved):
I think it's fine. Thanks!

PR #3137, opened 2021-07-10
Allow opening the pocket menu via reload in singleplayer mode.

EntranceJew:
Because you couldn't, previously. Uses the same workaround that keys use.

### DIFF
--- a/entities/weapons/pocket/cl_menu.lua
+++ b/entities/weapons/pocket/cl_menu.lua
@@ -61,7 +61,7 @@ function DarkRP.openPocketMenu()
     reload()
     frame:SetSkin(GAMEMODE.Config.DarkRPSkin)
 end
-
+usermessage.Hook("PocketMenu", DarkRP.openPocketMenu)
 
 --[[---------------------------------------------------------------------------
 UI
--- a/entities/weapons/pocket/cl_menu.lua
+++ b/entities/weapons/pocket/cl_menu.lua
@@ -61,9 +61,7 @@ function DarkRP.openPocketMenu()
     reload()
     frame:SetSkin(GAMEMODE.Config.DarkRPSkin)
 end
-net.Receive( "DarkRP_PocketMenu", function( len, pl )
-    DarkRP.openPocketMenu()
-end )
+net.Receive("DarkRP_PocketMenu", DarkRP.openPocketMenu)
 
 --[[---------------------------------------------------------------------------
 UI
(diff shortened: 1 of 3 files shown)


entities/weapons/pocket/cl_menu.lua
```diff
@@ -61,7 +61,7 @@ function DarkRP.openPocketMenu()
     reload()
     frame:SetSkin(GAMEMODE.Config.DarkRPSkin)
 end
-
+usermessage.Hook("PocketMenu", DarkRP.openPocketMenu)
```
FPtje, line 64: There are existing usermessages that I cannot convert to net messages, but usermessages _are_ deprecated. Could you change this to use net messages rather than usermessages? Thanks!

FPtje (changes requested):
Nice work, one comment

FPtje:
Oh, I've enabled the checks, and I see that glualint fails with a trailing whitespace warning. Could you fix that as well?

entities/weapons/pocket/cl_menu.lua
```diff
@@ -61,7 +61,11 @@ function DarkRP.openPocketMenu()
     reload()
     frame:SetSkin(GAMEMODE.Config.DarkRPSkin)
 end
-usermessage.Hook("PocketMenu", DarkRP.openPocketMenu)
+net.Receive( "DarkRP_PocketMenu", function( len, pl )
+	if not ( IsValid( pl ) and pl:IsPlayer() ) then
```
FPtje, line 65: Why does the player need to be valid? It's always LocalPlayer right?

EntranceJew: I misunderstood the documentation; I thought it would have been possible to receive messages from other players (via some means) and therefore would have wanted to ensure we weren't allowing other people to force the local player to open their pockets. I see that clients can only use SendToServer. Removing it seems to function correctly in both singleplayer and multiplayer so it should be good after this.

entities/weapons/pocket/shared.lua
```diff
@@ -118,9 +118,14 @@ function SWEP:SecondaryAttack()
 end
 
 function SWEP:Reload()
-    if not CLIENT then return end
+    if CLIENT then
+      DarkRP.openPocketMenu()
```
FPtje, line 122: Won't this open the pocket menu twice on a server? Once through calling `DarkRP.openPocketMenu` here, and once through the net message received from the server?

Perhaps just add a `game.SinglePlayer()` check to sending the message?

FPtje (approved):
I've pushed the requested changes. Thanks for the contribution!

PR #3139, opened 2021-07-22
Replace activator with ply

Mekrip:
This should send the "You cannot afford this" notify to the right recipient.


FPtje:
Nice! Thanks

PR #3142, opened 2021-08-02
Stop people from dropping weapons they spawn with by default.

liondadev:
I still see new server owners leaving this on without even noticing it. This would erase that issue. This pretty much allows duping of weapons you spawn with if the server has an inventory system. Even though most servers have it disabled, some people forget or don't know to disable it. This would also require a pull request to darkrpmodification to disable it in that config file but Ill do that if this one is accepted.

### DIFF
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -86,7 +86,7 @@ GM.Config.droppocketdeath               = true
 -- dropweapondeath - Enable/disable whether people drop their current weapon when they die.
 GM.Config.dropweapondeath               = false
 -- Whether players can drop the weapons they spawn with.
-GM.Config.dropspawnedweapons            = true
+GM.Config.dropspawnedweapons            = false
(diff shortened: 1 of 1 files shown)


ReliableSolus:
Why do we need developers who can't figure out how to set up basic things? By that logic, you should disable all things that can be used for evil. For example, client scripts and the sleep module.

FlorianLeChat:
Personally, I think it's more a problem of server owners not spending the time/doing it than an issue with the gamemode. Following this logic, there will be no change in attitudes and the problem will move somewhere else again and again.

FPtje:
I think the better question here is "what is the right default?"

In a barebones DarkRP setup with no changed settings or addons, which option would provide the best experience? Setting it to false enforces job boundaries by making it harder to share job weapons with friends who don't have the job.

Setting it to true allows weapons to circulate more freely. From that sense, I think restricting it might be a good idea. It prevents by script that which admins would otherwise have to punish for.

FPtje:
I say it's a good default.

PR #3143, opened 2021-08-08
Removed DarkRP.removeJob ( 'jobCount' (a nil value) )

nodgear:
2015'ish changes had a top level jobCount local variable (and used on create/remove functions).
This local variable doesn't exist anymore, yet, the function is trying to reduce the jobCount.

Removed the counter, it's not used anywhere else in the function, and the local variable is not even present anymore


FPtje:
This was actually changed quite recently. Thanks for fixing it!

PR #3145, opened 2021-08-24
Optimise server's side of entityvars.lua

mcNuggets1:
- The client already has this optimisation in place and this results in much faster access time
FProfiler has shown, that "getDarkRPVar" is one of the most expensive bottlenecks

- Renamed unnamed hook
This hook fooled me a few times already into searching for it, because I didn't know what it did, while working with hook.GetTable()[index] to find a hook returning something

This PR resolves this.

Fixing this problem ALONE is not possible without modifying DarkRP core files.

### DIFF
--- a/gamemode/modules/base/sv_entityvars.lua
+++ b/gamemode/modules/base/sv_entityvars.lua
@@ -88,7 +88,7 @@ function meta:sendDarkRPVars()
 
             local vars = {}
             for var, value in pairs(DarkRPVars[target] or {}) do
-                if self ~= target and (privateDRPVars[target] or {})[var] then continue end
+                if self ~= target and (privateDarkRPVars[target] or {})[var] then continue end
                 table.insert(vars, var)
             end
 
--- a/gamemode/modules/base/sv_entityvars.lua
+++ b/gamemode/modules/base/sv_entityvars.lua
@@ -255,12 +255,16 @@ function meta:customEntityCount(entTable)
 end
 
 hook.Add("PlayerDisconnected", "DarkRP_VarRemoval", function(ply)
-    DarkRPVars[ply] = nil
-    privateDarkRPVars[ply] = nil
-
     maxEntities[ply] = nil
 
     net.Start("DarkRP_DarkRPVarDisconnect")
         net.WriteUInt(ply:UserID(), 16)
     net.Broadcast()
+end)
+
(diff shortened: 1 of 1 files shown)


mcNuggets1:
Should work fine now.
Runs on my servers as an experiment now too.
If anything is wrong, I will fix it immediately as I see it.

mcNuggets1:
Everything seems to be working now.

I did a few tests using prints too.
https://i.imgur.com/W0pMJTo.png

FPtje:
I finally have some time to look at this PR. Sorry for the delay. Thanks again for the contribution. Some comments.

> FProfiler has shown, that "getDarkRPVar" is one of the most expensive bottlenecks

> It was place 25 in bottlenecks, serverside

If something is in place 25 of a profiler, I wouldn't call that a bottleneck, much like I wouldn't call 25cm from the top of a wine bottle the bottleneck. Generally with profiling, the cost centers decrease exponentially from top to bottom. You can see that in this example profile, where the 25th spot is highlighted.

hl2_2021-09-11_15-58-56

Let's look at the `total time in ms`. The 25th spot is at 0.74ms. The top spot, is at 14.57ms. That means it is **19.69** times slower. By comparison, the 25th spot is nothing, not even worth looking at. The reason why talking about performance is often talking about bottlenecks, is because of the analogy. With a literal bottleneck, only a small part of the bottle is the cause of the reduced amount of liquid that can flow through it. The same is generally true for performance: generally only a small set of instructions are responsible for slowing the whole shebang down. 

--- 

Nevertheless, good contributions regarding performance are always welcome. Some code comments:

There's a reference to `ply.DarkRPVars` still left here:

https://github.com/FPtje/DarkRP/blob/7e06296120f4b6891fc65fd325018e3391a603e4/gamemode/modules/base/sv_gamemode_functions.lua#L715-L715

There also isn't backwards compatibility. I'm quite sure addons that use `ply.DarkRPVars` will break. We all know they shouldn't, but we also all know many people don't care. There should be some backwards compatibility, like there is clientside:

https://github.com/FPtje/DarkRP/blob/b8d395def9cb84a7b413c350ecc8fe2d27ebade1/gamemode/modules/base/cl_entityvars.lua#L22-L25

I've fixed that in 2d96d4a493bdf596b920873b8afc020ef283254d 👍

PR #3146, opened 2021-09-17
Don't put bots in AFK-Mode

Uschipanzer:
Except bots from afk system because they don't trigger "KeyPress" hook.


FlorianLeChat:
Can you put a space below this check? This will make the code a little more readable.

FPtje:
Shouldn't the check be in the function `KillAFKTimer`? After all, that's where the timer runs. By putting it in this function it makes it impossible to set a bot to being AFK, even if they were to run the `/afk` command. The latter is useful for testing.

PR #3147, opened 2021-09-20
Add canGiveLicense hook

luca1197:
This hook would be useful to further restrict giving a gun license.
Currently, it is only possible to use the `police = true` property in a job to allow giving a license, leaving it out / setting it not `false` disallows it. In my case, I would like to restrict it to a specific police job because there are several of them in my case.

Of course, using the `canChatCommand` hook is possible, but I think a `canGiveLicense` hook makes the whole thing easier and more obvious, I only came across the existence of this hook when I wrote a commit message from this pull request because it's not necessarily something you think of when wanting to restrict these cases.

A small note about the `LookingAt` stuff, I moved it to the beginning of the function to make it accessible to the hook, I don't think it will cause any issues.

### DIFF
--- a/gamemode/modules/police/sv_commands.lua
+++ b/gamemode/modules/police/sv_commands.lua
@@ -297,30 +297,16 @@ end
 DarkRP.defineChatCommand("requestlicense", RequestLicense)
 
 local function GiveLicense(ply)
-    local noMayorExists = fn.Compose{fn.Null, fn.Curry(fn.Filter, 2)(ply.isMayor), player.GetAll}
-    local noChiefExists = fn.Compose{fn.Null, fn.Curry(fn.Filter, 2)(ply.isChief), player.GetAll}
-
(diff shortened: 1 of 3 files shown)


mcNuggets1:
Wonderful

FPtje:
Nice work! I've changed some things in the implementation:

- The default check for mayor/chief/cp check is now in the default hook handler. This means you can override this behavior by explicitly returning true. This is especially useful if you want to base this decision on something other than mayor/chief/cp, e.g. if you want gun dealers to be able to grant licenses
- The default message for not granting a license is no longer `DarkRP.getPhrase("unable", "/givelicense", cantGiveReason)`. The listener of the hook should do the translation.

> A small note about the LookingAt stuff, I moved it to the beginning of the function to make it accessible to the hook, I don't think it will cause any issues.

This is indeed not an issue. Thanks for being vigilant :+1:

PR #3151, opened 2021-10-01
Prevent adding too much ammo from shipment

JurisBog:
According to the [SWEP wiki](https://wiki.facepunch.com/gmod/Structures/SWEP), the `DefaultClip` should be the only item specifying additional bullets. "Making it higher than ClipSize will give player additional ammo on spawn".

The existing implementation fills the Clip in its entirety and tops it up with `DefaultClip`.

### DIFF
--- a/entities/entities/spawned_shipment/init.lua
+++ b/entities/entities/spawned_shipment/init.lua
@@ -92,8 +92,8 @@ function ENT:Use(activator, caller)
 
     local canUse, reason = hook.Call("canDarkRPUse", nil, activator, self, caller)
     if canUse == false then
-        if reason then DarkRP.notify(activator, 1, 4, reason) end
-        return
+      if reason then DarkRP.notify(activator, 1, 4, reason) end
+      return
     end
 
     hook.Call("playerOpenedShipment", nil, activator, self)
@@ -106,6 +106,25 @@ function ENT:Use(activator, caller)
         if not IsValid(self) then return end
         self.SpawnItem(self)
     end)
+end
+
+local function calculateAmmo(class, shipment)
+    local clip1, ammoadd = shipment.clip1, shipment.ammoadd
+
+    local defaultClip, clipSize
+    local wep_tbl = weapons.Get(class)
+    if wep_tbl and wep_tbl.Primary then
+        defaultClip = wep_tbl.Primary.DefaultClip
+        clipSize = wep_tbl.Primary.ClipSize
+    end
+    ammoadd = ammoadd or defaultClip
+
+    -- If the clip is empty, fill it with additional bullets
+    if not clip1 then
+        clip1 = (ammoadd > 0) and math.min(clipSize, ammoadd) or 0
+        ammoadd = ammoadd - clip1
+    end
+    return ammoadd, clip1
 end
 
 function ENT:SpawnItem()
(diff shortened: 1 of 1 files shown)


entities/entities/spawned_shipment/init.lua
```diff
@@ -92,8 +92,8 @@ function ENT:Use(activator, caller)
 
     local canUse, reason = hook.Call("canDarkRPUse", nil, activator, self, caller)
     if canUse == false then
-        if reason then DarkRP.notify(activator, 1, 4, reason) end
-        return
+      if reason then DarkRP.notify(activator, 1, 4, reason) end
+      return
```
FlorianLeChat, line 96: I think there is an indentation problem here.

entities/entities/spawned_shipment/init.lua
```diff
@@ -108,6 +108,25 @@ function ENT:Use(activator, caller)
     end)
 end
 
+local function calculateAmmo(class, shipment)
+    local clip1, ammoadd = shipment.clip1, shipment.ammoadd
+
+    local defaultClip, clipSize
+    local wep_tbl = weapons.Get(class)
+    if wep_tbl and wep_tbl.Primary then
```
FPtje, line 116: When this condition is `false`, `defaultClip` and `clipSize` will remain `nil`. That will cause errors with the `math.min` call and arithmetic below.

entities/entities/spawned_shipment/init.lua
```diff
@@ -108,6 +108,25 @@ function ENT:Use(activator, caller)
...
+        defaultClip = wep_tbl.Primary.DefaultClip
+        clipSize = wep_tbl.Primary.ClipSize
+    end
+    ammoadd = ammoadd or defaultClip
+
+    -- If the clip is empty, fill it with additional bullets
+    if not clip1 then
```
FPtje, line 123: This condition does not do what the comment says it does. The comment mentions empty, but the condition is about `nil` or `false`. In Lua, `not 0` evaluates to `false`.

entities/entities/spawned_shipment/init.lua
FPtje: This code is not modified besides the indentation. I think this change can be undone.

FPtje (commented):
Thanks for looking into this. In my experience, working with ammo, clips and whatnot in GMod is very complicated. That sort of code is to be treated with care.

I'm not sure what the problem is that is solved by this pull request. Could you explain in which situation the shipment would add too much ammo? Does it have something to do with turning `spawned_weapon`s into shipments?

FPtje:
What's the status of this PR?

entities/entities/spawned_shipment/init.lua
JurisBog: I meant to specify if it was undefined, define it to a full clip (existing behaviour), but remove the extra ammo.

entities/entities/spawned_shipment/init.lua
Kefta: Weapon.Primary does not have to be a table. Type check before indexing.

entities/entities/spawned_shipment/init.lua
```diff
@@ -108,6 +108,26 @@ function ENT:Use(activator, caller)
     end)
 end
 
+local function calculateAmmo(class, shipment)
+    local clip1, ammoadd = shipment.clip1, shipment.ammoadd
+
+    local defaultClip, clipSize = 0, 0
+    local wep_tbl = weapons.Get(class)
+    if wep_tbl and wep_tbl.Primary then
+        defaultClip = wep_tbl.Primary.DefaultClip
```
Kefta, line 117: or 0

entities/entities/spawned_shipment/init.lua
```diff
@@ -108,6 +108,26 @@ function ENT:Use(activator, caller)
...
+        clipSize = wep_tbl.Primary.ClipSize
```
Kefta, line 118: or 0

PR #3152, opened 2021-10-14
Fixed console ban

doxzter:
Fixed FAdmin bug with console ban. When admin using "fadmin ban" with non-quoted steamid, rule breaker isn't getting kicked immediately, This commit fixes it

### DIFF
--- a/gamemode/modules/fadmin/fadmin/playeractions/kickban/sv_init.lua
+++ b/gamemode/modules/fadmin/fadmin/playeractions/kickban/sv_init.lua
@@ -128,6 +128,7 @@ local function Ban(ply, cmd, args)
             args[1] = targets[1]
             args[2] = args[6]
             args[3] = args[7]
+            targets = FAdmin.FindPlayer(args[1]) or targets
             for i = 2, #args do
                 if i >= 4 then args[i] = nil end
             end


gamemode/modules/fadmin/fadmin/playeractions/kickban/sv_init.lua
```diff
@@ -128,6 +128,7 @@ local function Ban(ply, cmd, args)
             args[1] = targets[1]
             args[2] = args[6]
             args[3] = args[7]
+            targets = FAdmin.FindPlayer(args[1]) or targets
```
FPtje, line 131: `args[1]` is equal to `"STEAM_0"`, as shown by the check on line 126. This means that `FAdmin.FindPlayer will either find everyone or no one. Probably the former. I don't reckon this is what we want :grimacing:

FPtje (changes requested):
Thanks for the PR! I don't think this solution is quite it. Could you provide an example of where the FAdmin ban command would go wrong?

doxzter:
Here's record of this bug. The core problem is that admins can use this bug to ban higher-level ones, bypassing fadmin immunity
https://youtu.be/J9FLwHJmaRs

gamemode/modules/fadmin/fadmin/playeractions/kickban/sv_init.lua
FPtje: Ah of course, in that case I think `targets[1]` is clearer. I'll edit it.

FPtje:
I've merged the pull request. Thanks for the contribution!

PR #3153, opened 2021-11-09
Fix clip size bug

mark27205:
This is unexpected behaviour, and broke some stuff on my server. It is not necessary (particularly if it breaks things) to return 0, if ammoadd is not higher than 0.
Ammoadd is spareammo in createShipment. Some people did spareammo = 0. So this code will literally remove the first clip entirely, eventough ClipSize and DefaultClip are correct.

### DIFF
--- a/entities/entities/spawned_shipment/init.lua
+++ b/entities/entities/spawned_shipment/init.lua
@@ -122,7 +122,7 @@ local function calculateAmmo(class, shipment)
     -- If the clip is undefined (default behaviour), default it to a full clip
     -- But remove the bullets used for it.
     if not clip1 then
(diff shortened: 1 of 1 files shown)


mark27205:
Actually if you think further, my changes also don't solve the problem. I'll commit a new solution later this day, to fix this behaviour.

mark27205:
Completely forgot about this. I pushed my changes now. I actually reverted the behaviour that was introduced. Because why change it in shipments to gmod behaviour but NOT in single weapons? Quite inconvenient in my opinion.

mcNuggets1:
+1

FPtje:
Sounds good. Thanks!

PR #3155, opened 2021-11-21
exploit fix

Karlend:
.5


Karlend:
I think you'll understand. I don't want to give more info to prevent servers to be "attacked".
More info - telegram @karlend

PR #3156, opened 2021-11-21
exploit fix

Karlend:
.5

I think you'll understand. I don't want to give more info to prevent servers to be "attacked".
More info - telegram @Karlend

### DIFF
--- a/gamemode/modules/money/sh_money.lua
+++ b/gamemode/modules/money/sh_money.lua
@@ -3,7 +3,7 @@ local entMeta = FindMetaTable("Entity")
 
 function pMeta:canAfford(amount)
     if not amount or self.DarkRPUnInitialized then return false end
-    return math.floor(amount) >= 0 and (self:getDarkRPVar("money") or 0) - math.floor(amount) >= 0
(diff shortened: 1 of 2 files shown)


FlorianLeChat:
I find it totally counterproductive not to describe the reproduction method or at least a brief description to replicate this supposed exploit. Also, if I can give you any confidence at all, some of the DarkRP servers today are not using a recent version. Why? Because server owners download the gamemode once on GitHub and never update it, while the safest way is to use the workshop version because it takes advantage of the latest changes (with a certain time lag to make sure the latest changes are stable). So sooner or later, obsolete servers will face this problem and will have to update, no worries.

FPtje:
The fix does give some good hints as to the direction of the exploit, though. I appreciate that the exact details are hidden. DarkRP doesn't really have a responsible disclosure procedure, so I think this method is fine. Thanks for the report.

Kapppa:
The CreateCheque change isn't needed as it calls canAfford in the method.

FPtje:
Also added a `DarkRP.toInt` function to be used instead of `tonumber` for user input.

PR #3158, opened 2021-12-22
Pass self into various custom checks for opportunities of optimisation

BullyHunter32:
```lua
TEAM_BULLYHUNTER = DarkRP.createJob("BullyHunter", {
    color = Color(255, 255, 255, 255),
    model = {
        "models/player/Group03/Female_01.mdl",
        "models/player/Group03/Female_02.mdl"
    },
    description = [[This text will serve as the description of
                    this team.]],
    weapons = {"weapon_p2282"},
    command = "example",
    max = 0.7, -- at most 70% of the players can have this job. Set to a whole number to set an absolute limit.
    salary = 45,
    admin = 0,
    vote = false,
    hasLicense = false,
    customCheckSteamIDs = {
        ["STEAM_0:1:60438167"] = true,
    },
    customCheck = function(ply, self)
        return self.customCheckSteamIDs[ply:SteamID()]
    end
})
```

### DIFF
--- a/gamemode/modules/base/sh_createitems.lua
+++ b/gamemode/modules/base/sh_createitems.lua
@@ -18,7 +18,7 @@ local function declareTeamCommands(CTeam)
         if CTeam.admin == 1 and not ply:IsAdmin() or CTeam.admin == 2 and not ply:IsSuperAdmin() then return false end
         if isnumber(CTeam.NeedToChangeFrom) and plyTeam ~= CTeam.NeedToChangeFrom then return false end
         if istable(CTeam.NeedToChangeFrom) and not table.HasValue(CTeam.NeedToChangeFrom, plyTeam) then return false end
-        if CTeam.customCheck and CTeam.customCheck(ply) == false then return false end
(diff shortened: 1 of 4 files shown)


Kefta:
I hope no addons call customChecks manually. Good change.

FlorianLeChat:
> I hope no addons call customChecks manually. Good change.

I just checked some popular F4 menus from both the GmodStore and the Workshop and basically almost all addons do that unfortunately.

Some examples: (these are the same checks each time)
image
image
image
image
image

The current problem is the DarkRP doesn't offer verification mechanisms which can be used outside on other addons so it's pretty normal for F4 menu creators to make their own copies of these checks.

FPtje:
Thanks for the PR. I like the idea in general, but it indeed won't work with addons. I suppose it's best to just use the global variable. 

> The current problem is the DarkRP doesn't offer verification mechanisms which can be used outside on other addons so it's pretty normal for F4 menu creators to make their own copies of these checks.

Even if we added it now, F4 menus wouldn't start using it.

PR #3163, opened 2022-01-26
Hide Entity Display When Entity Invisible

ryanm2711:
This is useful for various scenarios, such as a staff member being cloaked.

### DIFF
--- a/gamemode/modules/hud/cl_hud.lua
+++ b/gamemode/modules/hud/cl_hud.lua
@@ -332,7 +332,8 @@ local function DrawEntityDisplay()
     local aimVec = localplayer:GetAimVector()
 
     for _, ply in ipairs(players or player.GetAll()) do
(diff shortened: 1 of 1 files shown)


Kefta:
The render mode being transalpha does not inherently mean they're invisible. You also have to check if their alpha is 0.

mcNuggets1:
Cloak your staffmembers using SetNoDraw(true), dont cloak them via colors.

FPtje:
If we want to discuss proper cloaking, I've always argued that making invisible is not the right way. Clients will always receive the cloaked player's position. Use spectate with roaming instead

ryanm2711:
Thanks for the info, I'll do some changes accordingly.

mcNuggets1:
> 

Wasn't it possible for players to stop the networking to other players?

Kefta:
`and` has priority over `or`, the parentheses should be around the `or` clause.

FPtje:
Thanks for the change. I've fixed the parentheses, this is now ready to be merged 👍

ryanm2711:
Ah sorry, I didn't see the additional message to fix formatting. Thanks for doing it for me, I didn't know about all of that so thanks for teaching me something new! 😃

PR #3166, opened 2022-03-06
Allow the font to display glyphs that are outside of the Latin-1 range.

MelowFX:


FPtje:
Thanks! The files from FPP and FSpectate are not supposed to be edited, so I reverted that.

PR #3170, opened 2022-04-17
Fix admin variable

Wgis:
If you don't set admin value, it's not being work and simplerr throw error. 
This is do the same, but work as must work.

### DIFF
--- a/gamemode/modules/base/sh_createitems.lua
+++ b/gamemode/modules/base/sh_createitems.lua
@@ -414,10 +414,11 @@ function DarkRP.createJob(Name, colorOrTable, model, Description, Weapons, comma
 
     local CustomTeam = tableSyntaxUsed and colorOrTable or
         {color = colorOrTable, model = model, description = Description, weapons = Weapons, command = command,
-            max = maximum_amount_of_this_class, salary = Salary, admin = admin or 0, vote = tobool(Vote), hasLicense = Haslicense,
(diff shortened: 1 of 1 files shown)


FPtje:
It looks like this change will always break when using the table way of creating jobs (i.e. `DarkRP.createJob("name", {....})`), as it will override whatever you set in the table to `0`.

Could you show me an example of the simplerr error you mentioned?

FPtje:
The simplerr error of the `admin` field missing is desired behavior, so I think I can close the issue. If the provided example proves that something is wrong, I'll reopen the issue.

Wgis:
You can add isnumber for additional check. 
If you forgot adding/don't want adding `admin = 0 ` everythere - is the best solution. 
But now all roles must have `admin = 0 `...

Wgis:
If you set admin = 1 where it's need - it must work well.
You can delete all `admin = 0 ` in jobrelated.lua and just leave `admin = 1 ` and etc.

FPtje:
Ah, so you want to make the `admin` parameter optional. That is better done in `sh_checkitems`. See da39648cb59b782bc5c3213d94d608bcc6b553f6:
https://github.com/FPtje/DarkRP/blob/da39648cb59b782bc5c3213d94d608bcc6b553f6/gamemode/modules/base/sh_checkitems.lua#L182-L188

Wgis:
Yep. I've see it. Use is bad idea. But you answer is true)

PR #3171, opened 2022-04-18
Fix earthquake tremor setup

brandonsturgeon:
This has bothered me for awhile. By default, DarkRP creates a `env_physexplosion` immediately when it's loaded.
This isn't a big deal, it doesn't break anything, but gmod reports an error because DarkRP creates an entity too early.
```
Trying to create entities too early! (env_physexplosion)
```

This PR just delays the initial function call (and `PostCleanupMap` hook because some addons run that hook too early) until `InitPostEntity`, making sure the game is ready for new entities.

After testing, it appears earthquakes still work as expected and the error is absent from the startup logs.
gmod_CVqwCSjjpM

### DIFF
--- a/gamemode/modules/events/sv_events.lua
+++ b/gamemode/modules/events/sv_events.lua
@@ -100,8 +100,11 @@ local function createTremor()
     tremor.nodupe = true
     tremor:Spawn()
 end
-createTremor()
-hook.Add("PostCleanupMap", "DarkRP_events", createTremor)
+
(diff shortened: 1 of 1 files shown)


FPtje:
Nice. Thanks!

FPtje:
> InitPostEntity errors potentially cause this to never be added

InitPostEntity breakage (or addon returning) already breaks Darkrp beyond recognition. A solution for that would solve a **lot** of problems.

PR #3172, opened 2022-04-21
HUDShouldDraw DarkRP_VoiceChat

bailmast:
The ability to disable the annoying voice chat icon via HUDShouldDraw


Kefta:
Since you're already editing the file, you should pass self from GM:HUDPaint and use that instead of GAMEMODE in all the local functions.

bailmast:
Do you mean every single one, even GM.Config?

> Since you're already editing the file, you should pass self from GM:HUDPaint and use that instead of GAMEMODE in all the local functions.

Kefta:
self is only predefined in GM functions. You must pass it as an argument to the local functions so that the hook calls can access it.

FPtje (approved):
Thanks for the contribution!

PR #3177, opened 2022-06-10
GM.Config.allowPickupEquippedWeapons

cresterienvogel:
Added a config bool to enable/disable whether players can pick up weapons they already have. It's really annoying when you accidentally press +use and lose an entire weapon.

### DIFF
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -33,6 +33,8 @@ GM.Config.allowsprays                   = true
 GM.Config.allowvehicleowning            = true
 -- allowvnocollide - Enable/disable the ability to no-collide a vehicle (for security).
 GM.Config.allowvnocollide               = false
+-- allowPickupEquippedWeapons - Enable/disable whether players can pick up weapons they already have.
+GM.Config.allowPickupEquippedWeapons    = true
 -- alltalk - Enable for global chat, disable for local chat.
 GM.Config.alltalk                       = false
 -- antimultirun - Disallow people joining your server(s) twice on the same account.
--- a/gamemode/modules/base/sv_gamemode_functions.lua
+++ b/gamemode/modules/base/sv_gamemode_functions.lua
@@ -585,7 +585,7 @@ local adminCopWeapons = {
 function GM:PlayerCanPickupWeapon(ply, weapon)
     if ply:isArrested() then return false end
     if weapon.PlayerUse == false then return false end
-    
+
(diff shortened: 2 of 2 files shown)


mcNuggets1:
Please no HasWeapon check in a frequently called hook

cresterienvogel:
> Please no HasWeapon check in a frequently called hook

I don't think it's called that often

mcNuggets1:
Just put the HasWeapon check behind your config check

FPtje:
Thanks!

FlorianLeChat:
Sorry to mention just now when this has been merge. It seems this setting may prevent the player from being able to have extra grenades and SLAMs from Half-Life 2. In fact, these two "weapons" do not have ammunition and require the player to spawn new weapons (`weapon_frag`, `weapon_slam`).

You can do the test easily:
- Start a singleplayer session with the updated DarkRP gamemode after this pull request.
- Spawn a grenade and/or a SLAM.
- If you want to have more than one grenade or one SLAM, you can't.

mcNuggets1:
Dont use the config option?
Add a whitelist?

FlorianLeChat:
I'll wait for a response from the creator of the pull request to find out how he can deal with it. Because this issue applies to all the consumable weapons and we all know that this kind of weapons are regularly used on DarkRP servers. There needs to be a compromise between adding this kind of restriction but allowing it for certain classes of weapons (for example with a whitelist or a more advanced detection).

cresterienvogel:
probably whitelist will be the best option

Kefta:
Whitelist seems fragile. Why not just check !IsScripted() && SWEP.Primary/Secondary.Ammo == -1?

FPtje:
I think this option is great for servers that want to separate ammo from guns. While that doesn't technically work with grenades and SLAMs, it does work in the general case. Would it be an idea to leave more complicated behavior to custom hooks?

Kefta:
I think it's trivial to improve with the check I just provided. Many other custom SWEPs will be affected by this. No need to leave to leave it to addons that will never reach mass popularity and which will just open up question issues going forward. I don't understand the rationale of this being true by default in the first place, but if it is, non-ammo weapons should function as they did.

FPtje:
Oh whoops, it definitely shouldn't be true by default

FPtje:
Reverted for now. Thanks for spotting.

cresterienvogel:
The true value meant that players could pick up weapons the same way as before

PR #3182, opened 2022-06-22
Micro fix in cam.Start workaround

AMD-NICK:


Kefta:
The workaround isn't even needed anymore, this is done in the engine.

DBotThePony:
Engine fix is different from this workaround though, because engine checks scene balance after each rendered frame.

<details>
<summary>Offtopic</summary>

*thanks Rubat, i love my `Calling cam.End2D because someone forgot to!` and right after that `cam.End2D undeflow` spam*

</details>

Either way, this workaround should be removed completely.

DBotThePony:
> It errors for underflow immediately, just as this workaround does.

It (workaround) doesn't, because it does not end dangling scenes at the end of current frame (workaround only prevent direct underflows caused by direct cam.End calls), but the engine does, so workaround and engine fix are different.

Kefta:
When I say "this is done in the engine," I mean that this workaround is already done in the engine. The ending of dangling scenes is a completely different fix made at the same time, but that's not what's being discussed here so there's no reason to bring it up.

FPtje:
I must say I was suspicious, @Kefta. I remember there being claims of this being fixed before, and the fix was only a very specific case. It was still very easy to crash the game with a single bad `cam` call. However, this seems to be really fixed properly in the game now. That's nice.

I've removed the workaround from DarkRP and the DarkRPMod 👍

PR #3184, opened 2022-07-08
correct mass when pocketing sents, not just props

Aws0mee:


Kefta:
The default is already phys:GetMass() when the variable is unset, and I don't see it set anywhere in the codebase. Is this a fix for some addons?

Aws0mee:
> The default is already phys:GetMass() when the variable is unset, and I don't see it set anywhere in the codebase. Is this a fix for some addons?

phys:GetMass() appears to return 1 under some conditions which allows entities that should not be pocketed to be pocketed. I saw a fix was already in place for props (which are usually in the pocket blacklist anyway) so I thought the code would be better off in PlayerSpawnedSENT as well. 

I'd also like to implement the following code to fix this in entities purchased from the F4, but I'm not sure where it would be acceptable to put this code.
```lua
hook.Add("playerBoughtCustomEntity", "testHook", function(ply, tblEnt, ent, cost)
    local phys = ent:GetPhysicsObject()
    if IsValid(phys) then
        ent.RPOriginalMass = phys:GetMass()
    end
end)
```

Kefta:
What conditions is the mass set to 1? There are no SetMass calls that apply to general SENTs in the codebase so it seems like it should be the SENT's job to set RPOriginalMass before they call SetMass.

Aws0mee:
> What conditions is the mass set to 1? There are no SetMass calls that apply to general SENTs in the codebase, so it seems like it should be the SENT's job to set RPOriginalMass before they call SetMass and clear it after.

If you pick up an entity with the gravgun and then check its mass it should return 1. I believe there are other situations where it would also return 1 but I'm not sure. I don't know if this is a game bug or if it was intended, but I'd be happy to report it to facepunch if it's an issue in gmod.

Aws0mee:
@Kefta I made a video if you want to see what I'm talking about exactly https://youtu.be/zVnfdrRMqpI. Like I said before, I'm not sure if this is a game issue, or if it's intended. If it's not intended just let me know and I'll report it. If it is intended, then I think the code in this PR will be necessary.

FPtje:
Thanks for the PR! I think setting mass to 1 on gravgun pickup makes sense. After all, while holding things, you shouldn't be able to bash other players' heads in.

That would also mean that putting this in SENT spawning code wouldn't always work, though, as not all entities are SENTs. Would it instead be an idea to do something with gravgun pickup? E.g. Disallow pocketing props that are being held by gravgun?

Aws0mee:
> Thanks for the PR! I think setting mass to 1 on gravgun pickup makes sense. After all, while holding things, you shouldn't be able to bash other players' heads in.
Ah, never thought of that. I guess it is intended then.


> That would also mean that putting this in SENT spawning code wouldn't always work, though, as not all entities are SENTs. Would it instead be an idea to do something with gravgun pickup? E.g. Disallow pocketing props that are being held by gravgun?

I can add a hook for GravGunOnPickedUp and GravGunOnDropped instead which should work for all entities, but which file do you think would be appropriate for these hooks. Also, should the code in PlayerSpawnedProp be removed too since it basically accomplishes the same goal?

FPtje:
On gravgun picked up might be too late. That would need to be tested. Otherwise the "can gravgun pickup" could be abused to store that variable. @Kefta what do you think?

Kefta:
I'll test tomorrow, the ordering is a bit finicky.

Aws0mee:
The following code prevents any entity from being pocketed while it's being gravgunned. I tested and it seems to work. However, this will result in no entities being able to be pocketed while gravgunned even if they're small enough and should be able to be. Because of this, I'm not sure if this is the best fix, but it does work nonetheless. 

```lua
hook.Add("GravGunOnPickedUp", "DarkRPPocketFix", function(ply, ent)
    ent.beingGravGunned = true
end)

hook.Add("GravGunOnDropped", "DarkRPPocketFix", function(ply, ent)
    ent.beingGravGunned = nil
end)

hook.Add("canPocket", "DarkRPPocketFix", function(ply, ent)
    if ent.beingGravGunned then return false end
end)
```

FPtje:
I'm looking back at this issue. @Kefta Have you been able to test the ordering? Thanks!

FPtje:
The gravgun pickup and drop seem to be reliable enough to work even when a player is killed or kicked while holding an entity. Commit https://github.com/FPtje/DarkRP/commit/ccd1f23f8e806b1ae75ff6eaddabf87718ce1562 forbids pocketing an entity when it is held by the gravgun.

I also tested printing an entity's mass on gravgunpickup: it always prints `1`, which means that this hook is indeed too late to get an entity's mass. I could mess with the `Can...` hook, but honestly I'd rather avoid that entirely, as it is a messy workaround.

Unless people _really_ have a good reason to pick up things being held by the gravgun I think this is a good solution.
Thanks @Aws0mee for reporting the issue and providing an initial solution, and thanks @Kefta for the feedback 👍

Kefta:
Sorry I haven't had a computer recently, meant to test before I left home but I can verify the changes you made are solid next week.

FPtje:
No worries. I'm grateful for any help and support you give 👍

MrBrax:
https://youtu.be/EUPnvxklls4?t=295

neat coincidence

PR #3187, opened 2022-08-04
Add policeRequest hook

Down-s:
Added this hook as I'm making an addon that needs to check when a player sends a /999 request.

### DIFF
--- a/gamemode/modules/police/sh_interface.lua
+++ b/gamemode/modules/police/sh_interface.lua
@@ -213,7 +213,7 @@ DarkRP.hookStub{
 }
 
 DarkRP.hookStub{
-    name = "combineRequest",
+    name = "policeRequest",
     description = "When a player sends a police request.",
     parameters = {
         {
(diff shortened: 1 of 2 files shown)


WilliamVenner:
Police aren't necessarily combine

Down-s:
> Police aren't necessarily combine

I'm just keeping it consistent with the function name, combine seems to be used interchangeably with police in the code for whatever reason.

WilliamVenner:
Eh there's an additional ambiguity as well, `combineRequest` can be easily read as `combine request`... combining what???!?!?!

FPtje:
Personally I'm against making hooks for every individual chat command. Every chat command could have its own hook, but it's more bloat than it is worth. It's better to override the chat commands or to use these two hooks:

https://darkrp.miraheze.org/wiki/Hooks/Server/canChatCommand

https://darkrp.miraheze.org/wiki/Hooks/Server/onChatCommand

PR #3189, opened 2022-08-07
"Optimization" of hungermod

Seef0x:


Kefta:
Lua doesn't evaluate 0 as false. You need to keep the == 0 check.

Seef0x:
Oh my bad, there was always the == 0 check but i had to get confused with the second pull request

mcNuggets1:
I just woke up, but it shouldn't work even after that, because the energy var is substracted a line up.
Meaning, you would use the uneditted var.

Seef0x:
Lol i didn't see
Is caching the calculation and using it on the 2 lines more optimized than re recup?

FPtje:
> Is caching the calculation and using it on the 2 lines more optimized than re recup?

Not really, getting the DarkRPVar is just a table lookup. 

I'll close the issue. Thanks for the attempt at optimisation.

Kefta:
It does save a hashtable lookup and function call so it is marginally more efficient, but you should've fixed the issue I pointed out in time.

PR #3191, opened 2022-08-17
CustomShipments table is sequential, so ipairs should be used.

Aws0mee:
CustomShipments should always be sequential, so ipairs should be used to iterate it.

```lua
shipByName[string.lower(name or "")] = table.insert(CustomShipments, customShipment)
```


FPtje:
It's possible for people to remove shipments, even after they have been loaded in. This makes `ipairs` iterate over only a part of the table. That breaks a bunch of stuff.

PR #3193, opened 2022-08-18
Fix error when tmysql4 is preferred but not installed

### LINKED ISSUE #3192: error when tmysql4 is preferred but not installed
When tmysql4 is the preferred module, but it is not installed an error will be thrown which will prevent the rest of the code in the file from running and will break the gamemode. 
```
[ERROR] gamemodes/darkrp/gamemode/libraries/mysqlite/mysqlite.lua:143: attempt to index global 'tmysql' (a nil value)
  1. loadMySQLModule - gamemodes/darkrp/gamemode/libraries/mysqlite/mysqlite.lua:143
   2. unknown - gamemodes/darkrp/gamemode/libraries/mysqlite/mysqlite.lua:153
    3. include - [C]:-1
     4. unknown - gamemodes/darkrp/gamemode/init.lua:42
 ```

### PR DESCRIPTION
**Pull request by Aws0mee**
Fixes #3192

### DIFF
--- a/gamemode/libraries/mysqlite/mysqlite.lua
+++ b/gamemode/libraries/mysqlite/mysqlite.lua
@@ -125,7 +125,7 @@ local function loadMySQLModule()
     local moo, tmsql = file.Exists("bin/gmsv_mysqloo_*.dll", "LUA"), file.Exists("bin/gmsv_tmysql4_*.dll", "LUA")
 
     if not moo and not tmsql then
-        error("Could not find a suitable MySQL module. Supported modules are MySQLOO and tmysql4.")
+        error("Could not find a suitable MySQL module. Please either:\nInstall tmysql. It can be obtained from https://github.com/SuperiorServers/gm_tmysql4\nInstall MySQLOO. It can be obtained from https://github.com/FredyH/MySQLOO\nDue to this error, MySQL is disabled. This means that SQLite is used instead to store data.")
     end
     moduleLoaded = true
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Kefta** [comment]
You made a typo.

**Comment by Kefta** [comment]
Also I disagree with how the fix is implemented. tmysql should be checked to see if it's an existing table before usage, but an error should still be thrown that the preferred module isn't installed.

**Comment by Aws0mee** [comment]
> You made a typo.

I checked again and I don't see it? Where is the typo so I can fix it?

**Comment by Aws0mee** [comment]
> Also I disagree with how the fix is implemented. tmysql should be checked to see if it's an existing table before usage, but an error should still be thrown that the preferred module isn't installed instead of silently moving on.


I can reverse the checks in the ig statement when I get back to my pc. I agree with you about throwing an error, should probably call ErrorNoHalt if the preferred module is not installed. I just wanted to fix the current implementation without making any changes. I can add this too if need be.

**Comment by FPtje** [comment]
Please note that MySQLite has its own repo: https://github.com/FPtje/mysqlite

This is why the check-modified-subtree check fails.

**Comment by Kefta** [comment]
> > You made a typo.
> 
> I checked again and I don't see it? Where is the typo so I can fix it?

`tmsql`

**Comment by Aws0mee** [comment]
@Kefta that looks good to me, am I mistaken?

```lua
    local moo, tmsql = file.Exists("bin/gmsv_mysqloo_*.dll", "LUA"), file.Exists("bin/gmsv_tmysql4_*.dll", "LUA")
```
 
 ```lua
     if MySQLite_config.Preferred_module == "tmysql4" and tmsql then
 ```

Aws0mee:
> Please note that MySQLite has its own repo: https://github.com/FPtje/mysqlite
> 
> This is why the check-modified-subtree check fails.

Do you want me to close this and make a PR in that repo instead? Or would you like me to make an identical PR in that repo as well as this one? Also, should I reverse the order of the checks and throw an error like @Kefta suggested?

FPtje:
> Do you want me to close this and make a PR in that repo instead? Or would you like me to make an identical PR in that repo as well as this one? Also, should I reverse the order of the checks and throw an error like @Kefta suggested?

A PR in the other repo would work. I have a script to update DarkRP, so I'll take care of that 👍 

I also agree with @kefta, in that you should throw an error instead of silently moving on. The current error might not be as clear to people, but at least it shows something is wrong. With this PR, the code pretends that all is ok. This can lead people to think that the MySQL connection is working, and that all is well, even though everything is still being done with the SQLite file. 

An error should explain what's wrong and more importantly, what a user can do to fix it. Something like:

> The preferred module for MySQL is selected to be tmysql4. However, tmysql4 does not appear to be installed. Please either:
> - Install tmysql. It can be obtained from https://github.com/SuperiorServers/gm_tmysql4.
> - Select MySQLOO as the preferred module for MySQL. This module appears to be installed.
> 
> Due to this error, MySQL is disabled. This means that SQLite is used instead to store data.

This can indeed be an `ErrorNoHalt`. You can throw a similar error for MySQLOO. Judging by the code, we're already past the check that _either_ is installed. This error would just be that the wrong one is selected.

In the same vein, the `require` silently goes past the problem where the wrong one is selected. That should be fixed as well.

Aws0mee:
@FPtje What do you think of the changes I just made? If they look good I will make the same changes on the mysqlite repo.

FPtje:
Looks good for a PR on the other repo. I'll take a closer look later :+1:

Aws0mee:
> Looks good for a PR on the other repo. I'll take a closer look later 👍

Made the PR in the other repo, I think it would be easiest for you to simply merge both since they have the same code, but you can do whatever you like. Let me know if you want any additional changes.

FPtje:
With the PR in MySQLite, I'm closing this PR.

See https://github.com/FPtje/MySQLite/pull/16

PR #3194, opened 2022-08-31
Update sv_gamemode_functions.lua

Aws0mee:
Not a big deal since it's only called after lua refresh, but might as well.


jadedblue7442:
OMG FINALLY!!!! AWESOME IS THE BEST HE MADE THE LIGHTSABER ADDON AS WELL AS DARKRP FOR GARRYSMOD

FPtje:
I feel like I should get a little more credit for that DarkRP part 🤷‍♂️

PR #3196, opened 2022-08-31
Fix notification messages

### LINKED ISSUE #3195: Incorrect notification messages
When attempting to spawn vehicles, NPCs, or SENTs without the proper permissions will give a notification that is grammatically incorrect and does not make much sense.

image

Aws0mee:
fixes #3195

### DIFF
--- a/gamemode/modules/language/sh_english.lua
+++ b/gamemode/modules/language/sh_english.lua
@@ -267,9 +267,9 @@ local my_language = {
 
     -- Notifies
     disabled = "%s has been disabled! %s",
-    gm_spawnvehicle = "The spawning of vehicles",
-    gm_spawnsent = "The spawning of scripted entities (SENTs)",
-    gm_spawnnpc = "The spawning of Non-Player Characters (NPCs)",
+    gm_spawnvehicle = "spawn vehicles",
+    gm_spawnsent = "spawn scripted entities (SENTs)",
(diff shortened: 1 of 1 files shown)


FPtje:
Could you check to see if those translations are used elsewhere as well? With the language system I made one big mistake, which is to think that I could translate just parts of phrases, and then combine different bits to make a single message. That does not work with translation at all, and you get wacky situations like this, where the grammar is completely off.

The proper solution might be to give that message its own translation entirely.

Aws0mee:
> Could you check to see if those translations are used elsewhere as well? With the language system I made one big mistake, which is to think that I could translate just parts of phrases, and then combine different bits to make a single message. That does not work with translation at all, and you get wacky situations like this, where the grammar is completely off.
> 
> The proper solution might be to give that message its own translation entirely.

I already searched each one with GitHub search and they all only showed one result. GitHub search has been bad for me in the past, so it’s possible it was leaving out other marches. I can check with something better when I get back on my PC tomorrow, but I’m pretty sure it’s only used in this one place.

Aws0mee:
Already checked with GitHub search and double checked with VS Code, looks like all 3 are only used in that single message.

image
image
image

FPtje:
Thanks for checking! Could you also make this change in the darkrpmod? Thanks!

https://github.com/FPtje/darkrpmodification/blob/master/lua/darkrp_language/english.lua

PR #3199, opened 2022-09-15
chatlisteners: optimize drawChatReceivers

NathanX-S:


FPtje:
I'm afraid that if-statement doesn't do anything useful. In that function, it is already certain that the thing should be drawn. The function is not even called otherwise.

You can see it below in the code, where the hook is added when a player starts speaking, and removed when they stop.

NathanX-S:
@FPtje DBugR reports the function as being hit always. Something worth looking into.

FPtje:
DBugR is wrong, but I can see why. The hook itself is added and removed when someone respectively starts and stops chat. I don't think it would be possible for DBugR to see this reasoning. It just sees that function on HUDPaint.

So basically DBugR sees a function on HUDPaint, that for all it knows is called on every HUDPaint. That is true, but only when the player has their chat box open. Once they close it, the function is "suddenly" gone, and DBugR cannot keep track of it anymore.

PR #3200, opened 2022-09-24
Add message for /unownalldoors with no doors owned

Aws0mee:
I noticed when you use /unownalldoors it won't say anything if you don't have any doors which made me think I typed the command wrong. Now it will notify you that you don't have any doors instead of doing nothing.

### DIFF
--- a/gamemode/modules/doorsystem/sv_doors.lua
+++ b/gamemode/modules/doorsystem/sv_doors.lua
@@ -461,7 +461,7 @@ local function UnOwnAll(ply, cmd, args)
         cost = cost + GiveMoneyBack
     end
 
-    if amount == 0 then return "" end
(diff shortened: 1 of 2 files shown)


FPtje:
Thanks! 

For the next PR, could you improve your commit messages? Right now they just say "Update sv_doors.lua", "Update sv_english.lua" and "fix". This is not very clear. A better commit message would be the title of this PR, or "fix missing comma in sh_english.lua"

Aws0mee:
> Thanks!
> 
> For the next PR, could you improve your commit messages? Right now they just say "Update sv_doors.lua", "Update sv_english.lua" and "fix". This is not very clear. A better commit message would be the title of this PR, or "fix missing comma in sh_english.lua"

Sure thing, sorry about that.

PR #3201, opened 2022-10-09
Fix lottery

Wgis:
Lottery has been broken, if gamemode translated on other languages. 
This fix helps, and work propertly. 
I check it.


FPtje:
Thanks! Could you explain how it was broken?

mcNuggets1:
Or how this fixes it?

Wgis:
It's broken because there must be unique id, but text replaced with language symbols. For example: лотерея1. After all, player cant answer and lottery says nobody not enter. But this no need. It just technical ID and nobody dont see it.

PR #3204, opened 2022-10-11
Fixed question answer for another player

doxzter:
Every player could answer on other player's question by using "ans" concommand with unique questionID. For example I used command "ans lottery1 1" and first player in player.GetAll() table entered lottery. This PR fixes that


SendLua:
👍

FPtje (approved):
Whoa, that's been in there all along? Nice find!

PR #3206, opened 2022-11-13
Add currency formatting options

djtb2924:
https://en.wikipedia.org/wiki/Decimal_separator

English ppl uses :
`GM.Config.currencyDecimalSeparator      = "."`
`GM.Config.currencyThousandSeparator     = ","`

"internationally recommended" (french ppl) uses :
`GM.Config.currencyDecimalSeparator      = ","`
`GM.Config.currencyThousandSeparator     = " "`

### DIFF
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -325,7 +325,6 @@ GM.Config.lockdownsound = "npc/overwatch/cityvoice/f_confirmcivilstatus_1_spkr.w
 -- The skin DarkRP uses. Set to "default" to use the GMod default derma theme.
 GM.Config.DarkRPSkin = "DarkRP"
 GM.Config.currency = "$"
-GM.Config.currencyDecimalSeparator = "."
 GM.Config.currencyThousandSeparator = ","
 GM.Config.chatCommandPrefix = "/"
 GM.Config.F1MenuHelpPage = "https://darkrp.miraheze.org/wiki/Main_Page"
--- a/gamemode/config/config.lua
+++ b/gamemode/config/config.lua
@@ -325,6 +325,8 @@ GM.Config.lockdownsound = "npc/overwatch/cityvoice/f_confirmcivilstatus_1_spkr.w
 -- The skin DarkRP uses. Set to "default" to use the GMod default derma theme.
 GM.Config.DarkRPSkin = "DarkRP"
 GM.Config.currency = "$"
+GM.Config.currencyDecimalSeparator = "."
+GM.Config.currencyThousandSeparator = ","
 GM.Config.chatCommandPrefix = "/"
 GM.Config.F1MenuHelpPage = "https://darkrp.miraheze.org/wiki/Main_Page"
 GM.Config.F1MenuHelpPageTitle = "DarkRP Wiki"
--- a/gamemode/modules/base/sh_util.lua
+++ b/gamemode/modules/base/sh_util.lua
@@ -40,18 +40,14 @@ function DarkRP.formatMoney(n)
 
     n = tostring(math.abs(n))
 
-    if config.currencyDecimalSeparator ~= "." then
-        n = string.Replace(n, ".", config.currencyDecimalSeparator)
-    end
-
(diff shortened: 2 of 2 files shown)


Leo-GMprog:
Yeah!

EvanPole:
👍

FPtje:
Thanks for the PR. The functionality looks good, I'll  take a closer look at the code when I have time 👍

gamemode/config/config.lua
```diff
@@ -325,6 +325,8 @@ GM.Config.lockdownsound = "npc/overwatch/cityvoice/f_confirmcivilstatus_1_spkr.w
 -- The skin DarkRP uses. Set to "default" to use the GMod default derma theme.
 GM.Config.DarkRPSkin = "DarkRP"
 GM.Config.currency = "$"
+GM.Config.currencyDecimalSeparator = "."
```
FPtje, line 328: For float reasons, I don't think it's wise to have options related to decimal numbers. DarkRP officially doesn't support decimal numbers. The reason for that is float inconsistencies. It's quite hard to avoid rounding errors and money amounts looking like `$2.000000000001`

gamemode/modules/base/sh_util.lua
```diff
@@ -34,17 +34,24 @@ function DarkRP.formatMoney(n)
...
+    local dp = string.find(n, config.currencyDecimalSeparator, 1, true) or #n + 1
 
     for i = dp - 4, 1, -3 do
-        n = n:sub(1, i) .. "," .. n:sub(i + 1)
+        n = n:sub(1, i) .. config.currencyThousandSeparator .. n:sub(i + 1)
     end
 
     -- Make sure the amount is padded with zeroes
-    if n[#n - 1] == "." then
+    if n[#n - 1] == config.currencyDecimalSeparator then
```
FPtje, line 54: In light of not supporting decimals, I think this code is dead in DarkRP. It should be possible to remove it.

FPtje (changes requested):
Thanks for the contribution! I think the Decimal stuff is not something we should support. Working with decimal number amounts is probably a huge can of worms, given the fact that float calculations are imprecise. Let's not give the idea that this is supported in DarkRP.

Kefta:
Floats are well-defined in Lua and not supporting them in DarkRP seems silly considering all numbers in (g)lua are floats, thus that statement introduces abstraction breaks abound.

FPtje:
I'm not sure what you mean by introducing abstraction breaks, but not supporting decimal numbers for money is not something I decided on the spot yesterday. Code paths relating to money call a function `DarkRP.toInt` to make sure a number is floored.

The reason is that indeed that Lua numbers are floats. Floating point numbers are great for many reasons, but they're also imprecise. You have to be careful in dealing with that, and common wisdom is never to represent money as floats. See [this stack overflow answer](https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency) for a good explanation.

djtb2924:
I see the issue there. However, it's not related to decimal separator. If we don't accept floats, why are we handling it ?

FPtje:
Lua only has one type for numbers, which is floating point. That's fine in general, but there is imprecision when doing calculations with decimals (i.e. Non-whole numbers). So that's the part that DarkRP doesn't support.

djtb2924:
indeed, so the last lines of the functions are useless, as well as the dp (should always be #n + 1) ?

gamemode/modules/base/sh_util.lua
```diff
@@ -34,13 +34,16 @@ function DarkRP.formatMoney(n)
...
     if n <= -1e14 then return "-" .. attachCurrency(tostring(math.abs(n))) end
 
+    local config = GAMEMODE.Config
+
     local negative = n < 0
 
     n = tostring(math.abs(n))
-    local dp = string.find(n, "%.") or #n + 1
+
+    local dp = string.find(n, ".", 1, true) or #n + 1
```
FPtje, line 43: Ah, this is a little clearer than the escaped dot. I had to look up what `'%.` meant again.

FPtje (approved):
Thanks!

PR #3212, opened 2023-01-05
getDarkRPVar optimisation

maxmol:
Indexing the player table in getDarkRPVar has proven to be expensive due to the amount of times this function is called every frame. I suggest localizing Player.UserID to improve performance.

Before the change:
image
After:
image


FPtje:
Whoa! That's a huge difference! That's a major bottleneck fix right there!

This is awesome!

PR #3213, opened 2023-01-06
Fix getDarkRPVar returning nil when the var is false

Kefta:
Ternary-assumption bug. Probably went unnoticed since most cases that check bool vars do something like `if (not ply:getDarkRPVar("some_bool")) then`.

image


SneakySquid:
another banging pr from our man code_gs

keep up the good work you're nothing but a boon to this community!

FPtje:
Thanks!

PR #3214, opened 2023-01-13
FAdmin_CanExecuteCommand hook

doxzter:
It will be very useful for developers have possibility to prevent some commands from running in some conditions without editing gamemode files and rewriting them. IDK why this isn't there yet


AMD-NICK:
I think you're looking for something like FAdmin_CanTarget or CAMI.PlayerHasAccess.

Examples of uses: https://forum.gm-donate.net/t/fadmin-zapret-dlya-rp-profy/1824/2 and https://forum.gm-donate.net/t/1755/9

doxzter:
Yeah, really. Didn't know about these hooks, thx. So, this is already there :)

PR #3217, opened 2023-03-31
Proper checks for FAdmin.PlayerName()

doxzter:
Now it returns Console when ply is console

### DIFF
--- a/gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
+++ b/gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
@@ -18,7 +18,9 @@ FAdmin.Messages.MsgTypesByName = {
 function FAdmin.PlayerName(ply)
     if CLIENT and ply == LocalPlayer() then return "you" end
 
-    return isstring(ply) and ply or not IsValid(ply) and "unknown" or ply:EntIndex() == 0 and "Console" or ply:Nick()
+    if isstring(ply) then return ply end
+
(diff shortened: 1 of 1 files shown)


gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
```diff
@@ -18,7 +18,9 @@ FAdmin.Messages.MsgTypesByName = {
 function FAdmin.PlayerName(ply)
     if CLIENT and ply == LocalPlayer() then return "you" end
 
-    return isstring(ply) and ply or not IsValid(ply) and "unknown" or ply:EntIndex() == 0 and "Console" or ply:Nick()
```
FPtje, line 21: Ah, so it would return `unknown`, because `IsValid` returns false. Got it :+1:

gamemode/modules/fadmin/fadmin/messaging/sh_shared.lua
```diff
@@ -18,7 +18,9 @@ FAdmin.Messages.MsgTypesByName = {
...
+    if isstring(ply) then return ply end
+
+    return isentity(ply) and (ply:EntIndex() == 0 and "Console" or ply:Nick()) or "unknown"
```
FPtje, line 23: Hypothetically you could throw a non-player entity in here and have it error on `ply:Nick()`. It would breach the idea of getting the player name, though, so I think it's fine.

FPtje (approved):
Thanks!

Kefta:
I would just use GetName so that it can work with players and non-players.

PR #3218, opened 2023-04-18
Added DrawChatIndicator hook

Luiggi33:
Added a hook to disable the chat indicator.
While you can disable it via DarkRP modification, said method doesn't allow it to be activated or deactivated again on the fly.


FPtje:
Thanks!

PR #3220, opened 2023-04-30
Add fallbacks for DarkRP Vars

Luiggi33:
I suggest adding the ability to make easy fallbacks, so you won't need to add any checks.
It is only optional, so there is no need to update anything ofc

### DIFF
--- a/gamemode/modules/base/cl_entityvars.lua
+++ b/gamemode/modules/base/cl_entityvars.lua
@@ -7,10 +7,11 @@ local pmeta = FindMetaTable("Player")
 -- This function is made local to optimise getDarkRPVar, which is called often
 -- enough to warrant optimizing. See https://github.com/FPtje/DarkRP/pull/3212
 local get_user_id = pmeta.UserID
-function pmeta:getDarkRPVar(var)
+function pmeta:getDarkRPVar(var, fallback)
+    fallback = fallback or nil
     local vars = DarkRPVars[get_user_id(self)]
-    if vars == nil then return nil end
-    return vars[var]
+    if vars == nil then return fallback end
+    return vars[var] or fallback
 end
 
 --[[---------------------------------------------------------------------------
--- a/gamemode/modules/base/sh_interface.lua
+++ b/gamemode/modules/base/sh_interface.lua
@@ -394,6 +394,12 @@ DarkRP.PLAYER.getDarkRPVar = DarkRP.stub{
             description = "The name of the variable.",
             type = "string",
             optional = false
+        },
+        {
+            name = "fallback",
+            description = "The value to return if the DarkRPVar doesn't exist.",
+            type = "any",
+            optional = true
         }
     },
     returns = {
--- a/gamemode/modules/base/sv_entityvars.lua
+++ b/gamemode/modules/base/sv_entityvars.lua
@@ -65,11 +65,12 @@ end
 --[[---------------------------------------------------------------------------
 Get a DarkRPVar
 ---------------------------------------------------------------------------]]
-function meta:getDarkRPVar(var)
+function meta:getDarkRPVar(var, fallback)
+    fallback = fallback or nil
     local vars = DarkRPVars[self]
 
     vars = vars or {}
-    return vars[var]
+    return vars[var] or fallback
 end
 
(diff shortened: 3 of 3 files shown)


gamemode/modules/base/cl_entityvars.lua
```diff
@@ -7,10 +7,11 @@ local pmeta = FindMetaTable("Player")
 -- This function is made local to optimise getDarkRPVar, which is called often
 -- enough to warrant optimizing. See https://github.com/FPtje/DarkRP/pull/3212
 local get_user_id = pmeta.UserID
-function pmeta:getDarkRPVar(var)
+function pmeta:getDarkRPVar(var, fallback)
+    fallback = fallback or nil
```
Kefta, line 11: Change this to an explicit nil check, otherwise fallback will be forced to nil if it's false.

Kefta (changes requested):

gamemode/modules/base/cl_entityvars.lua
```diff
@@ -7,10 +7,11 @@ local pmeta = FindMetaTable("Player")
...
     local vars = DarkRPVars[get_user_id(self)]
-    if vars == nil then return nil end
-    return vars[var]
+    if vars == nil then return fallback end
+    return vars[var] or fallback
```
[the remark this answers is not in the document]
Luiggi33: Didn't think of that :P
Thanks

gamemode/modules/base/cl_entityvars.lua
Luiggi33: Thx

gamemode/modules/base/cl_entityvars.lua
```diff
@@ -8,10 +8,11 @@ local pmeta = FindMetaTable("Player")
 -- enough to warrant optimizing. See https://github.com/FPtje/DarkRP/pull/3212
 local get_user_id = pmeta.UserID
 function pmeta:getDarkRPVar(var, fallback)
-    fallback = fallback or nil
+    fallback = fallback ~= nil and fallback or nil
```
Kefta, line 11: This doesn't fix it. ``true and false or nil`` evaluates to nil. Copy the == nil check idiom on line 13.

gamemode/modules/base/cl_entityvars.lua
```diff
@@ -8,10 +8,11 @@ local pmeta = FindMetaTable("Player")
...
     local vars = DarkRPVars[get_user_id(self)]
     if vars == nil then return fallback end
-    return vars[var] or fallback
+    local results = vars[var]
+    return results ~= nil and results or fallback
```
Kefta, line 15: Here too.

Kefta (changes requested):

gamemode/modules/base/cl_entityvars.lua
Luiggi33: I would just cut that check out and make a check at the bottom for if results == nil and fallback ~= nil. Seems like the better option, if I am not mistaken

Kefta:
```Lua
function pmeta:getDarkRPVar(var, fallback)
    local vars = DarkRPVars[get_user_id(self)]
    if vars == nil then return fallback end
    
    local results = vars[var]
    if results == nil then return fallback end
	
    return results
end
```
This is the cleanest imo

Kefta (approved):

gamemode/modules/base/cl_entityvars.lua
```diff
@@ -1,16 +1,20 @@
 local DarkRPVars = {}
 
 --[[---------------------------------------------------------------------------
-interface"someString"
```
FPtje, line 4: This looks like something done by glualint :thinking: 

Not sure how it got there, and I'm too lazy to git blame it, because it's probably me :joy:

FPtje:
Thanks!

gamemode/modules/base/cl_entityvars.lua
Luiggi33: Thought so too, it happens :P

Kefta:
@FPtje Is there a reason why UserID is used for the key of the var table on the server, but the entity is used on the client?

PR #3221, opened 2023-04-30
Updated action version

Luiggi33:
The ubuntu version that is currently configured for the check-modified-subtree workflow, is deprecated. That is why said workflow only outputs "Waiting for a runner to pick up this job... " and doesn't work.
Optionally, it could be put to `ubuntu-latest` to always stay "up-to-date", which is currently 22.04.


FPtje:
Thanks!

PR #3222, opened 2023-05-29
Add teleportOverride to unArrest

Aws0mee:
I noticed while making my jail escaping system that there doesn't seem to be a proper way to unarrest someone without respawning them if the config is set to do so or vice versa. This addition should allow you to decide manually if a player should be respawned or not, and will fall back on the config value if one is not provided. This should be much cleaner than my previous solution which was changing the config value, unarresting the player, then setting it back.

### DIFF
--- a/gamemode/modules/police/sv_init.lua
+++ b/gamemode/modules/police/sv_init.lua
@@ -370,7 +370,10 @@ function DarkRP.hooks:playerUnArrested(ply, actor, teleportOverride)
     end
 
     gamemode.Call("PlayerLoadout", ply)
-    if teleportOverride or (GAMEMODE.Config.telefromjail and teleportOverride ~= false) then
+    -- teleportOverride can either be nil, false, or a vector. Nil means "do not
+    -- modify behavior", false means "do not teleport", and a vector means
+    -- "teleport to this place instead"
+    if (GAMEMODE.Config.telefromjail or teleportOverride ~= nil) and teleportOverride ~= false then
         local pos
         if isvector(teleportOverride) then
             pos = teleportOverride


Meowlan (approved):

gamemode/modules/police/sv_init.lua
```diff
@@ -107,12 +107,12 @@ function plyMeta:arrest(time, arrester)
     end
 end
 
-function plyMeta:unArrest(unarrester)
+function plyMeta:unArrest(unarrester, teleportOverride)
```
FPtje, line 110: This adds a boolean argument that cannot really be interpreted from the call site without a comment:

```lua
ply:unArrest(someOtherPly, true)
```

In the above snippet, it's hard to interpret the `true`. I think there might be a better solution.

gamemode/modules/police/sv_init.lua
```diff
@@ -370,7 +370,7 @@ function DarkRP.hooks:playerUnArrested(ply, actor)
     end
 
     gamemode.Call("PlayerLoadout", ply)
-    if GAMEMODE.Config.telefromjail then
+    if teleportOverride or (GAMEMODE.Config.telefromjail and teleportOverride ~= false) then
```
FPtje, line 373: With a boolean `teleportOverride` it is not possible to set a custom respawn position through this function. Though arguably this could be implemented as a `ply:SetPos` _after_ calling the `ply:unArrest()`.

gamemode/modules/police/sv_init.lua
```diff
@@ -358,7 +358,7 @@ function DarkRP.hooks:playerArrested(ply, time, arrester)
     umsg.End()
 end
 
-function DarkRP.hooks:playerUnArrested(ply, actor)
+function DarkRP.hooks:playerUnArrested(ply, actor, teleportOverride)
```
FPtje, line 361: This hook does quite a few different things that someone might want to override:

- Exiting a vehicle
- Waking up from sleep
- Setting  respawn time
- Give the player weapons through `PlayerLoadout`
- Teleport to a spawn position if the config is enabled

Your use case is the last option, but I can imagine future PRs adding more arguments to deal with the other behaviours. The same can then be done for the other hooks. It's a bit of a slippery slope, but I think it is a good indication that this might not be the best way to go here.

Instead, would it be an idea to override `DarkRP.hooks.playerUnArrested` to remove the behaviours you don't want? Then in a separate hook (or in the override), you can have your own teleporting implementation.

gamemode/modules/police/sv_init.lua
Aws0mee: I think this could be used to set a custom respawn after being unjailed if you simply force unarrest respawn off and then set the pos right after. This would allow people to do things like bail someone out and teleport the player to them after being bailed out for example.

gamemode/modules/police/sv_init.lua
Aws0mee: I see your point with that. I can't really think of a better solution for this.

gamemode/modules/police/sv_init.lua
Aws0mee: I see your point. I do think the respawning aspect is probably the most important as I had literally no clue those other things happened without reading the code. I've seen many different addons like jail escaping systems, bail npcs, etc. and all of them seem to have subpar workarounds for unarresting a player without respawning them. I suppose I could just keep this change for my server only, but I think it's something a lot of other people could benefit from. The problem is that I do want players to be respawned when they get unarrested, but in specific cases I want to be able to choose. For example, if a player breaks another player out I don't want the escaping player to be respawned.

FPtje:
You raise very good points. How about this alternative:

Run a hook called something like `playerUnarrestTeleportPosition`. The `DarkRP.hooks` implementation has what's currently there, namely returning a spawn pos if the config is set.

Then in your code, you can add a hook to that that just returns false, to prevent the base hook from firing.

It would solve the boolean blindness, be better documented, and allow full freedom of what happens. What do you think?

Aws0mee:
> You raise very good points. How about this alternative:
> 
> Run a hook called something like `playerUnarrestTeleportPosition`. The `DarkRP.hooks` implementation has what's currently there, namely returning a spawn pos if the config is set.
> 
> Then in your code, you can add a hook to that that just returns false, to prevent the base hook from firing.
> 
> It would solve the boolean blindness, be better documented, and allow full freedom of what happens. What do you think?

I'm not quite sure I understand your solution. Are you suggesting this hook will return a position or false to determine where the player will be respawned or if they will be respawned at all? I think this could potentially work, but I think it would be suboptimal since every use case I can think of would want to decide how to handle the unarrest from the function call. It seems like adding a hook in this situation would be inefficient since we would have to add the hook, call the unarrest function, do logic in the hook to check if this player is the one we wanted to set the respawn on, and then set the respawn/prevent it.

If you want to knock out two birds with one stone here and allow for custom respawn positions as well as preventing respawning entirely I still think the best solution would be to add a parameter to the function. We can instead change it to be a vector position or false to control custom respawn position or prevent the respawn completely.

FPtje:
Good point. You know the use case best, so that makes sense. Changing the argument to a vector/false makes sense. Let's go for that!

Aws0mee:
Seems to work perfectly for me. I tested with the config value set to true and false. I tested true, false, and with the Vector(0,0,0) and they all seem to work.

FPtje:
Rebased and merged in https://github.com/FPtje/DarkRP/commit/710f1108d0667f0c3b52c60c6c3ceed3eefe17f5! Thanks for the contribution!

I've made some modifications in https://github.com/FPtje/DarkRP/commit/8fd1a9981753ec03936ea6b75e9caf9ecb373579 and https://github.com/FPtje/DarkRP/commit/f823c326ad7f533967370c69fa5a014677509bda to do the following:

- Do not call `PlayerSelectSpawn` when setting `teleportOverride` to a vector
- Allow `teleportOverride` to work even when `GAMEMODE.Config.telefromjail` is `false`

GitHub doesn't recognize the merge due to pushing to your branch _after_ pushing master.

PR #3229, opened 2023-07-19
Changed pronouns to be more neutral

cyanpone:
Changes all the pronouns I could find to be gender-neutral.


FPtje:
Thanks! Singular they works way better than `his/her` :+1:

PR #3231, opened 2023-07-28
don't call player.GetAll twice

Aws0mee:
noticed this while reading some of the code, might as well only call it once here


FPtje:
Thanks!

PR #3127, opened 2021-06-03
Remove the SteamID64/AccountID workaround

FlorianLeChat:
In the next update (scheduled for **June 9**) and currently implemented in the development branches, this issue is fixed.

Sources:
- https://github.com/Facepunch/garrysmod-issues/issues/2614
- https://commits.facepunch.com/381723
- https://docs.google.com/document/u/1/d/e/2PACX-1vTzEJEpdEje8e-FgsbyWGydu_Ez7p82MwOUPmRlUAAJ-KpkNJhHctyadZosfUYjVTz26KGip7bI7M9T/pub

**Note**: If this pull request is merge, then [this](https://github.com/FPtje/darkrpmodification/blob/b57df6bc662a3b1dad92aac773f100d737749018/lua/darkrp_config/disabled_defaults.lua#L179) line should also be deleted.

PR #3233, opened 2023-08-09
Remove the SteamID64/AccountID workaround

FlorianLeChat:
This pull request is basically inherited from #3127, but this time these two functions will always return a value whatever the situation (even in a `-multirun` environment) since yesterday's last update:
- https://store.steampowered.com/news/app/4000/view/3664293140704541926
- https://wiki.facepunch.com/gmod/Player:SteamID64~diff:551389
- https://wiki.facepunch.com/gmod/Player:AccountID~diff:551390

I did some testing with these changes and didn't see any issues as everything is now done internally by Garry's Mod, so it's pretty much a transparent change.

### DIFF
--- a/gamemode/modules/workarounds/sh_workarounds.lua
+++ b/gamemode/modules/workarounds/sh_workarounds.lua
@@ -86,25 +86,6 @@ timer.Simple(3, function()
 ]])
 end)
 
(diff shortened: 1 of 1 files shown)


FPtje:
Thanks! Crazy that things like this are still being fixed in gmod.

PR #3236, opened 2023-10-24
Update sh_language.lua

Argus-cpp:
Added support for localized darkrp notify for players


Argus-cpp:
I accidentally closed this =/

PR #3238, opened 2023-12-16
RemoveCcalls to SetCommand

### LINKED ISSUE #3237: Voting menu causes error, SetCommand was removed
### Description of the bug
The vote menu is broken and triggers an error after the newest gmod update. As Rubat stated [here](https://wiki.facepunch.com/gmod/Panel:SetCommand%7Ehistory), the function was removed for some reason.

image

Whenever somebody tries to become a vote job this bug will occur.

### Lua errors
```lua
[DarkRP] gamemodes/darkrp/gamemode/modules/voting/cl_voting.lua:74: attempt to call method 'SetCommand' (a nil value)
  1. Function - gamemodes/darkrp/gamemode/modules/voting/cl_voting.lua:74
   2. unknown - lua/includes/modules/usermessage.lua:80
   ```

He is not. I am simply posting this here for the sake of documenting this bug. I will be fixing it in a PR.

Aws0mee:
Remove SetCommand in the voting menu. fixes #3237


FPtje:
Thanks!

FPtje:
I don't have the ability to do a workshop update right now sadly

Aws0mee:
> Thanks!

You're welcome 😃
> I don't have the ability to do a workshop update right now sadly

That's fine. Hopefully other server owners will be able to find my issue #3237 and realize the solution is to update to this PR. Also, I don't think it's a huge issue. My players first reported this to me about a day or two ago, so if other server owners haven't figured it out by now they likely don't care very much.

Aws0mee:
I'm honestly a little confused as to why it was even removed I don't think it was deprecated, but I might be mistaken. Either way the voting menu should work fine without it, so we're all set I believe.

FlorianLeChat:
> I'm honestly a little confused as to why it was even removed I don't think it was deprecated, but I might be mistaken. Either way the voting menu should work fine without it, so we're all set I believe.

I suspect they were removed at the same time as when Rubat fixed some "unspecified security vulnerabilities", so maybe that has something to do with it.

Aws0mee:
> > I'm honestly a little confused as to why it was even removed I don't think it was deprecated, but I might be mistaken. Either way the voting menu should work fine without it, so we're all set I believe.
> 
> I suspect they were removed at the same time as when Rubat fixed some "unspecified security vulnerabilities", so maybe that has something to do with it.

Maybe, they seem to be hush hush about security vulnerabilities whenever there is one.

PR #3239, opened 2023-12-31
use an assert on storeMoney amount check

lionbryce:
If we don't allow people to go into the negatives, we should throw an error when something goes wrong.

It'd likely be worthwhile to remove the "amount > 0" check entirely as this is 1 of the 2 main methods of duping money in DarkRP (IE: going into the negatives then rejoining) though this commit is just meant to give a stack trace when something has gone wrong

also corrected allowing setmoney letting them go negative when that's not an option here

I should note, this doesn't technically "fix" any issues with DarkRP itself other than /setmoney allowing people to temporarily go negative


lionbryce:
I'd recommend looking at #3240 before this one

PR #3243, opened 2024-01-14
GM.NoLicense is empty on the client

sobri3tas:
weaponchecker prints GM.NoLicense weapons because its not shared and PrintWeapons() runs only on clients


FPtje:
Thanks for the PR! I don't have the ability to properly review this PR at the moment, as my laptop is in for repairs. I'll take a look as soon as I am able.

FPtje:
I just verified the fix. It indeed fixes the problem. Thanks!

This problem wouldn't appear if you used the darkrpmod though. I guess that's why no one hit it before.

PR #3244, opened 2024-01-15
Add support for all hook libraries that exist not just ulib/ulx

Srlion:
This will make it work with any hook library not just ulx/ulib's library, like my hook library https://github.com/Srlion/Hook-Library

### DIFF
--- a/gamemode/modules/chat/sv_chat.lua
+++ b/gamemode/modules/chat/sv_chat.lua
@@ -124,182 +124,36 @@ function GM:canChatCommand(ply, cmd, ...)
     return false
 end
 
-GM.OldChatHooks = GM.OldChatHooks or {}
+GM.OldHookCall = GM.OldHookCall or hook.Call
 
-local function callHooks(hooks, canReturn, ...)
-    local text
+local GM = GM
+function hook.Call(name, gm, ply, text, teamonly, ...)
+    if name == "PlayerSay" then
+        local dead = not ply:Alive()
 
-    for id, f in pairs(hooks) do
-        local isString
+        local text2 = text
+        local callback
+        local DoSayFunc
 
-        -- ULib hook tables are {fn = func, isstring = istring(id)}
-        if istable(f) then
-            isString = f.isstring
-            f = f.fn
-        else
-            isString = isstring(id)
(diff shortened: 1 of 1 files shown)


FPtje:
Thanks for the PR! It looks like you've found some of DarkRP's ugliest, but also some of its most revised code. Hooks are a huge pain point in Gmod. The default system is lacking, and alternative systems very often break some niche edge case that some addon somewhere ends up depending on. Not to mention the amount of times some addon incorrectly just always returns a value in some hook, silently breaking everything else. Ugh

I see your addon is a bit more explicit about returning values, and that your code in this PR is a lot shorter and more elegant than what is currently there. I think that's really nice.

One huge worry of mine, though, is these edge cases. Between this PlayerSay thing, and the bad hooks I've disabled in the workarounds files, there's a big history of fixing things w.r.t. other addons. How do we know that this change would not regress these fixes? This is an open question, and a hard one. Maybe the answer (partially) lies in how you tested this change.

Thanks again!

Srlion:
I checked workarounds and found 2 for PlayerSay

1. CAC is dead
2. ULXMeCheck and tested if it gets removed or not and it indeed gets removed:

```
FAdminChatCommands      =       function: 0xf0bbbd72
FAdmin_Chatmute =       function: 0xf0d8ef2a
FSpectate       =       function: 0xf0a8aba2
ULXGimpCheck    =       function: 0xf0ce99fa
ULXLogSay       =       function: 0xf0c93ae2
ULXMeCheck      =       function: 0xf0d90c6a
ULib_saycmd     =       function: 0xf0db4ff2
ulxPlayerSay    =       function: 0xf0ceade2
------------------
FAdminChatCommands      =       function: 0xf0bbbd72
FAdmin_Chatmute =       function: 0xf0d8ef2a
FSpectate       =       function: 0xf0a8aba2
ULXGimpCheck    =       function: 0xf0ce99fa
ULXLogSay       =       function: 0xf0c93ae2
ULib_saycmd     =       function: 0xf0db4ff2
ulxPlayerSay    =       function: 0xf0ceade2
```

FPtje:
Thanks for this PR! I see no objection to merging it. Let's merge it and see if people start complaining. If they don't for long enough, I can upload it to workshop :+1:

Srlion:
Thank you <333

PR #3245, opened 2024-01-24
Update mysqlite.lua

keksie:
Fixxes yesterdays modification of `file.Exists` whereafter mysqloo ceased to function

### DIFF
--- a/gamemode/libraries/mysqlite/mysqlite.lua
+++ b/gamemode/libraries/mysqlite/mysqlite.lua
@@ -123,15 +123,18 @@ local function loadMySQLModule()
     if moduleLoaded or not MySQLite_config or not MySQLite_config.EnableMySQL then return end
 
     local moo, tmsql = util.IsBinaryModuleInstalled("mysqloo"), util.IsBinaryModuleInstalled("tmysql4")
-
+    
     if not moo and not tmsql then
-        error("Could not find a suitable MySQL module. Supported modules are MySQLOO and tmysql4.")
+        error("Could not find a suitable MySQL module. Please either:\n  - Install tmysql. It can be obtained from https://github.com/SuperiorServers/gm_tmysql4\n  - Install MySQLOO. It can be obtained from https://github.com/FredyH/MySQLOO\nDue to this error, MySQL is disabled. This means that SQLite is used instead to store data.")
+
     end
     moduleLoaded = true
 
-    require(moo and tmsql and MySQLite_config.Preferred_module or
-            moo and "mysqloo"                                  or
-            "tmysql4")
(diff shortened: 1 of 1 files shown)


FPtje:
What's this Gmod breaking things recently?

FlorianLeChat:
Rubat's changes only affect beta branches at this time (https://github.com/Facepunch/garrysmod-issues/issues/5674#issuecomment-1908148281).

keksie:
> Rubat's changes only affect beta branches at this time ([Facepunch/garrysmod-issues#5674 (comment)](https://github.com/Facepunch/garrysmod-issues/issues/5674#issuecomment-1908148281)).

For matter of fact, those changes also affected dedicated servers running off main branch

robotboy655:
@keksie No they didn't.

@FPtje `file.Exists` was never meant to support wildcards, it was a side effect on implementation of a fix for file detection in `LUA` paths that was added last year.

Here's also a good reasoning for why you shouldn't check for binary module installation the way you do (besides the fact of using an undocumented side-effect of a relatively "recent" change) https://github.com/Facepunch/garrysmod-issues/issues/5705

FPtje:
I'm honestly not up to date on what happens in Gmod code. This util function definitely didn't exist when I wrote this check.

@robotboy655 if you want to deprecate things that's fine, I'm willing to update everything. If I have to find out it's broken through pull requests from people saying their server is suddenly broken, then I do not consider myself responsible.

I don't think I'll have time to look at this today. I will as soon as I get the chance. To prevent random server breakage in the future, please give me some info channel where I can find backwards incompatible changes on time.

robotboy655:
I honestly didn't even know it was a possibility when making changes, but in hindsight it makes sense.

There are always these edge cases where people do unexpected things, like so many cases of providing invalid input types to functions while those functions did nothing before, and are updated to throw errors like they always should've.

FPtje:
I can empathize with unexpected breakages, this does seem like an edge case.

FPtje:
Thanks for the PR. This PR has sadly been superseded by a PR on the original repo: https://github.com/FPtje/MySQLite/pull/17

and its subsequent merge into DarkRP. Nonetheless, thanks for the contribution.

FPtje:
I've now run the commands to merge MySQLite to FPP and DarkRP. Both have been pushed to master. It looks like FPP was further behind on some updates. Both FPP and DarkRP have been pushed to the workshop.

That means https://github.com/FPtje/DarkRP/pull/3244 also got pushed to the workshop, so CC @Srlion. That's a bit earlier than I'd hoped, but this is a good enough reason to push to workshop.

PR #3248, opened 2024-03-06
rewrited umsg to net

IP1A:


Kefta:
Changes from umsg to net have not been accepted in the past in order to not break addons that hook into DarkRP umsgs.

FPtje:
Thanks for the contribution. @Kefta is tight, legacy means changing user messages will start breaking things. As for the optimization, I dont really see the use of it if you still have to copy these vectors. Besides, these functions are not called often enough to warrant micro optimizations.

Kefta:
The min vector doesn't have to be copied since it's a constant, but it's a pretty negligible optimisation.

Issue #3252, opened 2024-03-11
After save changes in any files gmod server stop work


Bacring:
[//]: # (DO NOT REMOVE THIS TEMPLATE.)
[//]: # (GITHUB ISSUES ARE **NOT** FOR HELP)
[//]: # (GO TO THE DISCORD INSTEAD: https://darkrp.page.link/discord )

### Description of the bug
[//]: # (Describe the issue as accurately as possible)
When i edit any lua file in gamemodes/darkrp folder then my server freeze and i can't write in server console anymore
![image](https://github.com/FPtje/DarkRP/assets/101855368/19167d82-bec8-4454-837d-43f3703d17db)
![image](https://github.com/FPtje/DarkRP/assets/101855368/938a1d77-b694-44e4-afd8-507d50cc7ac5)
![image](https://github.com/FPtje/DarkRP/assets/101855368/0a96a9dc-09fe-4278-988f-ba995b89dd17)

### How to make the bug happen
[//]: # (Knowing how to make a bug happen allows the developers to figure out)
[//]: # (what the cause of the problem is and whether a certain fix solves it)
When i edit some files in darkrp,
before in prev years if i save changes then server freeze for 5 seconds and server continue work, but now after i enter/save changes in lua file darkrp gamemode then server freezing and go down, see screenshots
with other gamemodes when i edit files then i don't have this problem
### Lua errors
[//]: # (Note that errors on server startup are more important than other ones)
[//]: # (because they can cause other errors. Always look in the startup log of the server!)

### Why the developer of DarkRP is responsible for this issue
[//]: # (By posting on Github, you ask the developers of DarkRP to solve the problem)
[//]: # (It may seem obvious to you, but you have to make clear why they are the right people to look at it)

PR #3253, opened 2024-03-11
Fixed infinite loop

robotboy655:
Fixes https://github.com/FPtje/DarkRP/issues/3252

Culprit https://github.com/FPtje/DarkRP/commit/1b778981033ee472dea57820bd0e99958e68d19a

The root issue is that the `GM` table is recreated during auto refresh, so the value of `GM.OldHookCall` disappears at the time the code is run.

This is a quick and dirty fix, since I already invested the time to look into this as it was initially reported to me as a bug with the game.

### DIFF
--- a/gamemode/modules/chat/sv_chat.lua
+++ b/gamemode/modules/chat/sv_chat.lua
@@ -124,7 +124,7 @@ function GM:canChatCommand(ply, cmd, ...)
     return false
 end
 
-GM.OldHookCall = GM.OldHookCall or hook.Call
+g_DarkRPOldHookCall = g_DarkRPOldHookCall or hook.Call
 
 local GM = GM
 function hook.Call(name, gm, ply, text, teamonly, ...)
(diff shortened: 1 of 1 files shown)


FPtje:
Oh wow, I'll be honest. I did not think that was a DarkRP regression, but a Gmod issue instead.

Thanks for fixing this!

Bacring:
problem still have and again shut down my server, after i download again this darkrp gamemode and try save jobs.lua

FPtje:
What version of DarkRP do you use?

Bacring:
image

Bacring:
i see fix, i remove old folder darkrp and download again from github, and try change jobs.lua and it's shut down bug still worked

Bacring:
image

FPtje:
The code in your screenshot is old code.

Bacring:
> The code in your screenshot is old code.

see files, i think you not update file, can you share screen how it need look

Bacring:
yes i download works, but when you changes hud bug
image

FPtje:
That often happens when there's some error in the code. By the way, you're supposed to use the darkrpmod instead of editing the DarkRP files directly: https://github.com/FPtje/darkrpmodification

Bacring:
Okay, but please make like it was before, save changes and loading money job info

PR #3255, opened 2024-03-17
Use the improved Iterator functions for players & ents in loops

Luiggi33:


FlorianLeChat:
These two new functions [may suffer from a client-side synchronization issue](https://github.com/Facepunch/garrysmod-issues/issues/5800), we'll probably have to wait and see if this is confirmed with other reports

Luiggi33:
I saw that aswell, the chance is low but shouldn't be ignored

FPtje:
Thanks for the PR! Let's wait for upstream to make sure the function is stable

FPtje:
The linked issue doesn't seem to be moving much. I feel rather uneasy about meeting this, since it could display the problems of the other issue.

I'm sure that the issue would be found more quickly if I merge this, but that would go at the cost of the server owners and players.

Is anyone actively investigating the issue?

Luiggi33:
I haven't had the time to properly check it out, but this seems to be some cleanup related issue?
On the server I'm maintaining, I switched everything to the iterators and have yet to report an instance of this bug. I don't expect to run into this issue tho, because we don't use the gmod map cleanup anyways

FPtje:
It looks like the mentioned issue has no reproducible case just yet. It also looks like those iterator is sensitive to addons returning and/or throwing errors in some hooks: https://wiki.facepunch.com/gmod/ents.Iterator

Looks like it's too much of a risk to merge this. Thanks again for the PR, but I'll have to deny it.

PR #3256, opened 2024-03-20
[ Maintenance ] Reworked Dependency Checker

PhoneDroid:
*I used far too much time to get this just right, god I love shell scripting :skull:*

## Description

I reworked the dependency-modification-checking workflow / script

#### Workflow

- Now uses checkout v4
- Runs on latest ubuntu instead

#### Script

- Added various info comments
- Added info messages ( echo )
- General refactoring / code rework
- Reworked to use GitHub summary for better visualization
  - Prints out markdown as seen in the below screenshots
  - Links to the repo in question
  - Lists what files were problematic

image
image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,33 @@
+
+################################################################################
+#                                                                              #
+#                             Ｅｄｉｔｏｒｃｏｎｆｉｇ                             #
+#                                                                              #
+#                 This file defines basic file properties that                 #
+#                 editors / IDEs use when you edit documents.                  #
+#                                                                              #
+#                 Check https://EditorConfig.org for details.                  #
+#                                                                              #
+################################################################################
+
+
+root = true
+
+
+[*]
+
+end_of_line = lf
+
+
+[*.{editorconfig,json,yml,lua,txt,md}]
+
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+charset = utf-8
+
+
+[*.md]
+
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,23 +1,8 @@
-
-################################################################################
-#                                                                              #
-#                             Ｅｄｉｔｏｒｃｏｎｆｉｇ                             #
-#                                                                              #
-#                 This file defines basic file properties that                 #
-#                 editors / IDEs use when you edit documents.                  #
-#                                                                              #
-#                 Check https://EditorConfig.org for details.                  #
-#                                                                              #
-################################################################################
-
-
 root = true
-
 
 [*]
 
 end_of_line = lf
-
 
 [*.{editorconfig,json,yml,lua,txt,md}]
 
@@ -26,8 +11,3 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 charset = utf-8
-
-
-[*.md]
-
-trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 
 end_of_line = lf
 
-[*.{editorconfig,json,yml,lua,txt,md}]
+[*.{editorconfig,json,yml,lua,txt,md,sh}]
 
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.github/scripts/check-modified-subtree.sh
+++ b/.github/scripts/check-modified-subtree.sh
@@ -51,8 +51,7 @@ function check {
 
 
     # Check if any matching files were modified
-    local matched
-    matched=$(
+    local matched=$(
         echo "$files" |
         grep --perl-regexp "$path"
     )
--- a/.github/scripts/check-modified-subtree.sh
+++ b/.github/scripts/check-modified-subtree.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 #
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,33 @@
...
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+charset = utf-8
+
+
+[*.md]
+
+trim_trailing_whitespace = false
```
FPtje, line 33: I think this should be on for markdown as well

.editorconfig
FPtje, line 1: I'm a bit on the fence on whether to include this, but it could prevent some basic glualint warnings

.github/scripts/check-modified-subtree.sh
```diff
@@ -1,46 +1,123 @@
...
+#   Template for the GitHub summary
 
-if echo "$DIFFED_FILES" | grep -qP "^gamemode/libraries/sh_cami.lua"; then
-    echo "::error::The CAMI library has been edited. Please submit a PR to https://github.com/glua/CAMI instead!"
-    FAILED=true
-fi
+template="
+    > [!WARNING]
+    > Files from **[{project}]({repository})** have been modified!  
+    > ➞ Changes should only be made in it's repository
```
FPtje, line 32: "it's" should be "its", because the apostrophe goes when talking about possession.

Though perhaps a different wording altogether could explain it better, e.g.:

> These files are in the DarkRP repository, but they originally come from {url}. The changes should be applied there instead.

.github/scripts/check-modified-subtree.sh
```diff
@@ -1,46 +1,123 @@
...
+    local name="$1"
+    local path="^$2"
+    local repo="$3"
+
+    
+    #   Check if any matching files were modified
+    
+    local matched=$( 
+        echo "$files" |
+        grep --perl-regexp "$path"
```
FPtje, line 56: I like that you're using the full name for the options, it's very readable 👍

.github/scripts/check-modified-subtree.sh
```diff
@@ -1,46 +1,123 @@
...
+
+if [[ "$failed" = true ]] ; then
+    echo 'Some dependencies have been modified!'
     exit 1
 fi
+
+
+echo 'No dependencies have been modified.'
+
+echo "**No dependencies have been modified**" >> $GITHUB_STEP_SUMMARY
```
FPtje, line 121: I prefer this check not posting any comments on success, as that would be clutter that requires some thinking to process. The Green checkmark should be good enough

FPtje (changes requested):
Thanks! This is much appreciated. I have some comments, but overall I'm very happy with this.

.editorconfig
PhoneDroid: Well in Markdown, trailing double spaces act the same as `<br>` tags, so enabling  
it would trip people up, I'm fine with explicitly typing `<br>` in Markdown just FYI

FPtje: TIL, my editor automatically removes trailing whitespace, so it would be quite fragile. Explicit br is preferred.

.editorconfig
PhoneDroid: Well an editorconfig file is meant to give a users IDE - either through native 
support or via an extension - some basic formatting information / override  
( on project level ) their settings to prevent problems such as conflicting file  
endings for different operating system - for example - in commits

.github/scripts/check-modified-subtree.sh
PhoneDroid: I removed the GitHub summary message~

.github/scripts/check-modified-subtree.sh
PhoneDroid: Fixed the typo.

.editorconfig
PhoneDroid: Sure. What editor are you using btw?

PhoneDroid:
In case you prefer Discord for questions, my username is the same as on here~

.editorconfig
FPtje: VS Code, but I copied over my settings from back when I used Sublime Text. I really like the trim trailing settings, so I might have turned it on manually.

PhoneDroid: Could you for referrence install the editoconfig extension and see if it overrides your setting?

PhoneDroid: image

FPtje: Editorconfig works, but that's because I already have this extension installed: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

I reckon it won't work without this extension.

FPtje: Oh, btw the "TIL" is about double space meaning the same as `<br>`. It's a bit odd to parse spaces like that.

PhoneDroid: Well yeah ( in VSCode since it's not native ) tho in general if a developer  
doesn't have the tools used in a project they will have problems~

PhoneDroid: > Oh, btw the "TIL" is about double space meaning the same as `<br>`. It's a bit odd to parse spaces like that.

Well since markdown is a simplified version of writing html they probably  
wanted a way to break lines without having the user know what a break tag is~

FPtje (approved):
Sweet stuff! Thanks a lot for the contribution, and thanks for discussing it on Discord!

PR #3258, opened 2024-03-22
[ Feature ] Dependency Updater

PhoneDroid:
Created updater script & workflow.

Description:
- `Update-Dependency.sh` exports the `update` function
- `Update-Dependencies.sh` contains the export definitions & optional push logic
- Scripts can be tested with a `dry`-run variable
- Workflow sets github name / email used for committing
- Commits are per dependency and contain the their name
- update parameters are guarded
- Any realistic error case is logged / guarded
- script should exit if any update result is a failure
- script should not exit if anything inside the update function fails ( handled failures )
- update function allows to specify the remote repo name & url , the DarkRP 
( local ) folder/file as well as an optional folder in the remote repository
- Only commits if dependency had any changes
- Only pushes if there are any new commits

*Should be compatible with the current dependency structure and flexible for future changes.*

#### No Params / No Remote
image

#### No Repository
image

#### No Target
image

#### Actual Run
*Some dependencies seem to be actually out of date*
image
image
image

### DIFF
--- a/.github/scripts/Update-Dependencies.sh
+++ b/.github/scripts/Update-Dependencies.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+#
+#   Update a dependency from a remote repository.
+#
+
+echo '>>> Updating dependencies.'
+
+set -o errexit # Exit for any error
+
+dry=false
+
+
+source "$( dirname "$0" )/Update-Dependency.sh"
+
+
+update  "dry=${dry}"                          \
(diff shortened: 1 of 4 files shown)


FPtje:
Thanks for the contribution, though I think this feature goes a bit too far. Generally I update other repos manually. I don't always do this immediately. One example where I wait, is when e.g. FPP has a risky change, that should be on FPP itself first before I import it into DarkRP.

I prefer to stick to this workflow for now.

PR #3260, opened 2024-05-05
Don't clear the original owner nw var unless they're the one that reconnected

lionbryce:
Fixes:
Person A fallback to Person B, disconnects
Person B fallback to Person C, disconnect
Person B reconnects gets full ownership with no trace of A

edit: I completely forgot fpp had its own repo, cheers


FPtje:
Thabks for the PR! I see you already moved it to FPP. Thanks!

PR #3262, opened 2024-07-14
Fix: Workaround for localized texts in DarkRP not loading correctly

jjyao88:
In DarkRP, some in-game notifications are localized and sent from the server.

However, during server initialization, the convar `gmod_language` doesn't get read correctly. Even if we set `gmod_language [geo code]` in server.cfg, it returns an empty string at that time, causing all server-side localized texts to be in English instead of the desired language.

To fix this, I've proposed a workaround that delays reading the convar until the server is fully loaded.

And there's a minor concern that the client's language might still be different from the server, leading to incorrect localized texts. I think all localized texts should be handled on the client-side, not the server.

### DIFF
--- a/gamemode/modules/animations/sh_animations.lua
+++ b/gamemode/modules/animations/sh_animations.lua
@@ -1,7 +1,7 @@
 local Anims = {}
 
 -- Load animations after the languages for translation purposes
-hook.Add("loadCustomDarkRPItems", "loadAnimations", function()
+hook.Add("Think", "loadAnimations", function()
     Anims[ACT_GMOD_GESTURE_BOW] = DarkRP.getPhrase("bow")
     Anims[ACT_GMOD_TAUNT_MUSCLE] = DarkRP.getPhrase("sexy_dance")
     Anims[ACT_GMOD_GESTURE_BECON] = DarkRP.getPhrase("follow_me")
@@ -11,6 +11,7 @@ hook.Add("loadCustomDarkRPItems", "loadAnimations", function()
     Anims[ACT_GMOD_GESTURE_AGREE] = DarkRP.getPhrase("thumbs_up")
     Anims[ACT_GMOD_GESTURE_WAVE] = DarkRP.getPhrase("wave")
     Anims[ACT_GMOD_TAUNT_DANCE] = DarkRP.getPhrase("dance")
+    hook.Remove("Think", "loadAnimations")
 end)
 
 function DarkRP.addPlayerGesture(anim, text)
--- a/gamemode/modules/animations/sh_animations.lua
+++ b/gamemode/modules/animations/sh_animations.lua
@@ -1,7 +1,7 @@
 local Anims = {}
 
 -- Load animations after the languages for translation purposes
-hook.Add("Think", "loadAnimations", function()
+hook.Add("loadCustomDarkRPItems", "loadAnimations", function()
     Anims[ACT_GMOD_GESTURE_BOW] = DarkRP.getPhrase("bow")
     Anims[ACT_GMOD_TAUNT_MUSCLE] = DarkRP.getPhrase("sexy_dance")
     Anims[ACT_GMOD_GESTURE_BECON] = DarkRP.getPhrase("follow_me")
@@ -11,7 +11,6 @@ hook.Add("Think", "loadAnimations", function()
     Anims[ACT_GMOD_GESTURE_AGREE] = DarkRP.getPhrase("thumbs_up")
     Anims[ACT_GMOD_GESTURE_WAVE] = DarkRP.getPhrase("wave")
     Anims[ACT_GMOD_TAUNT_DANCE] = DarkRP.getPhrase("dance")
-    hook.Remove("Think", "loadAnimations")
 end)
 
 function DarkRP.addPlayerGesture(anim, text)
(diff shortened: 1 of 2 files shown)


jjyao88:
I just noticed that all pose names in animation menu are not translated as well.

https://github.com/FPtje/DarkRP/blob/c7312a411843e8e644cdd898d56ed49fca39616a/gamemode/modules/animations/sh_animations.lua#L4-L14

I modified the event name to `Think` by myself in the above code and it seems to work now.

gamemode/modules/animations/sh_animations.lua
```diff
@@ -1,7 +1,7 @@
 local Anims = {}
 
 -- Load animations after the languages for translation purposes
-hook.Add("loadCustomDarkRPItems", "loadAnimations", function()
+hook.Add("Think", "loadAnimations", function()
```
FPtje, line 4: I'm not sure I understand the problem. Looking quickly at the rest of the file, it seems that the server only uses `Anims` to see whether an animation is registered. It ignores the actual translation. The client does show the animation in the menu.

Other than that, instead of creating and removing a think hook, on can run a `timer.Simple(0,...)`. Note that either option will run the code quite a bit later than now.

gamemode/modules/language/sh_language.lua
```diff
@@ -1,10 +1,18 @@
 local rp_languages = {}
-local selectedLanguage = GetConVar("gmod_language"):GetString() -- Switch language by setting gmod_language to another language
+local selectedLanguage = "en" -- Switch language by setting gmod_language to another language
 
 cvars.AddChangeCallback("gmod_language", function(cv, old, new)
     selectedLanguage = new
 end)
 
+hook.Add("Think", "DarkRPSetLanguage", function()
```
FPtje, line 8: Shouldn't this be covered by the `cvars.AddChangeCallback` call above? After all, the moment the car goes from empty to set, that function should trigger.

Maybe the empty string check can be moved there instead. That would remove the need of the separate hook.

gamemode/modules/language/sh_language.lua
```diff
@@ -1,10 +1,18 @@
 local rp_languages = {}
-local selectedLanguage = GetConVar("gmod_language"):GetString() -- Switch language by setting gmod_language to another language
+local selectedLanguage = "en" -- Switch language by setting gmod_language to another language
```
FPtje, line 2: I think even if the convar is empty, we should still try to use it. That way it will still work if some other server managed to configure it differently.

FPtje (commented):
Thanks for the PR! I have to be honest, this is the first time I hear of this bug, and the language system has not changed in a long time. I think perhaps there may be something off in your setup. Maybe your `server.cfg` sets it, but some other config or script unsets it.

jjyao88:
> Thanks for the PR! I have to be honest, this is the first time I hear of this bug, and the language system has not changed in a long time. I think perhaps there may be something off in your setup. Maybe your `server.cfg` sets it, but some other config or script unsets it.

To confirm this, I have reinstalled the dedicated server on my non-English Windows. Despite only enabling DarkRP and my language file and setting my `server.cfg`, in-game notifications are still appearing in English.

gamemode/modules/language/sh_language.lua
jjyao88: Because the `cvars.AddChangeCallback` above doesn't work at all no matter what language codes I changed from server-side. This could be the bug directly from Facepunch...

So, I have to create a separate hook to deal with this.

gamemode/modules/animations/sh_animations.lua
jjyao88: Never mind, I'm not able to reproduce this issue. This may be due to my misconfiguration on my part. I'll revert this change.

FlorianLeChat:
I've been running a French DarkRP server for a decade and have strangely never experienced this issue, is your dedicated server on the stable branch or on a development branch?

gamemode/modules/language/sh_language.lua
jjyao88: Seems `gmod_language` only works for client side by design.

I haven't found a way to make `cvars.AddChangeCallback("gmod_language", ...)` work on server side.

FlorianLeChat:
My dedicated server uses Debian 12 configured in French. My Gmod server isn't virtualized or otherwise using Docker.

> Protocol version 24
Exe version 2023.06.28 (garrysmod)
Exe build: 19:02:32 Mar 21 2024 (9246) (4000)
GMod version 2024.03.21, branch: unknown
Linux 32bit Dedicated Server

gamemode/modules/language/sh_language.lua
```diff
@@ -5,6 +5,14 @@ cvars.AddChangeCallback("gmod_language", function(cv, old, new)
     selectedLanguage = new
 end)
 
+hook.Add("Think", "DarkRPSetLanguage", function()
+    gmodLanguage = GetConVar("gmod_language"):GetString()
```
FPtje, line 9: I noticed that this introduces a new global variable, by the way.

FPtje:
The language setting probably works for _most_ people without this pull request, since @FlorianLeChat mentioned it working for them. However, I also think this workaround is harmful for those where it does not work. I just pushed some changes, so this is good for merging.

Thanks!

PR #3264, opened 2024-07-17
Added custom default delay between entities spawning

HavannaFR:
Something that has existed for a long time for shipments but not for entities


FPtje:
Thanks for the PR! I'll take a closer look when I have time.

FPtje (approved):
This looks good. Thanks!

PR #3265, opened 2024-08-08
Add validity checks to dropDRPWeapon

RubikSW:
Prevent item duplication


entities/weapons/weapon_cs_base2/sv_commands.lua
```diff
@@ -1,5 +1,8 @@
 local meta = FindMetaTable("Player")
 function meta:dropDRPWeapon(weapon)
+    if !meta:IsValid() or !meta:IsPlayer() then return end
```
FPtje, line 3: This use of `meta` doesn't seem to be what you intend it to be. Do you mean to use `self` instead of `meta`?

FPtje (commented):
Thanks for the pull request!

Could you explain what the duplication exploit is and how this would fix it?

PR #3266, opened 2024-09-13
Isolate PreferredJobModel

Be1zebub:
Added column server to darkp_playermodels.  
Now changing the preferred model on server A will not affect the others.

also check job exists & model column not outdated - to display actual info in ui & network less data.

### DIFF
--- a/gamemode/modules/base/cl_jobmodels.lua
+++ b/gamemode/modules/base/cl_jobmodels.lua
@@ -1,31 +1,21 @@
 -- Create a table for the preferred playermodels
 --
--- Note: the server column wasn't always there, so players may not have a value
--- for it, even though the code in this file always adds it. That is why it is
--- added as a NULLable column.
-sql.Query([[CREATE TABLE IF NOT EXISTS darkp_playermodels(
-    jobcmd VARCHAR(45) NOT NULL PRIMARY KEY,
-    model VARCHAR(140) NOT NULL,
-    server TEXT NULL
+-- Note: in DarkRP before 2024-09, there was a different table called
+-- `darkp_playermodels` (note the misspelling of "darkp"). This table was
+-- missing the server column, meaning that preferred job models would persist
+-- across multiple servers. To make preferred job models store per server, this
+-- new table (without the spelling mistake) was created.
+--
+-- See the original issue to create the player model preference feature:
+-- https://github.com/FPtje/DarkRP/issues/979 and the subsequent refactor at
+-- https://github.com/FPtje/DarkRP/pull/3266
+sql.Query([[CREATE TABLE IF NOT EXISTS darkrp_playermodels(
+    server TEXT NOT NULL,
+    jobcmd TEXT NOT NULL,
+    model TEXT NOT NULL,
+    PRIMARY KEY (server, jobcmd)
 );]])
 
-
--- Migration: the `server` column was added in 2024-09. With the above query
--- only creating the table if not exists, the table may not have the `server`
--- column if it was created earlier. This will add it in retrospect.
-local tableInfo = sql.Query("PRAGMA table_info(darkp_playermodels)")
-local serverColumnExists = false
-
-for _, info in ipairs(tableInfo) do
-    if info.name == "server" then
-        serverColumnExists = true
-        break
-    end
-end
-
-if not serverColumnExists then
-    sql.Query("ALTER TABLE darkp_playermodels ADD COLUMN server TEXT;")
-end
 
 local preferredModels = {}
 
@@ -37,7 +27,7 @@ function DarkRP.setPreferredJobModel(teamNr, model)
     local job = RPExtraTeams[teamNr]
     if not job then return end
     preferredModels[job.command] = model
-    sql.Query(string.format([[REPLACE INTO darkp_playermodels VALUES(%s, %s, %s);]], sql.SQLStr(job.command), sql.SQLStr(model), sql.SQLStr(game.GetIPAddress())))
+    sql.Query(string.format([[REPLACE INTO darkrp_playermodels(server, jobcmd, model) VALUES(%s, %s, %s);]], sql.SQLStr(game.GetIPAddress()), sql.SQLStr(job.command), sql.SQLStr(model)))
 
     net.Start("DarkRP_preferredjobmodel")
         net.WriteUInt(teamNr, 8)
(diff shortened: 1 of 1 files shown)


gamemode/modules/base/cl_jobmodels.lua
```diff
@@ -42,11 +44,22 @@ local function sendModels() -- run after the jobs have loaded
...
+local function jobHasModel(job, model)
+    return istable(job.model) and table.HasValue(job.model, model) or job.model == model
+end
+
+timer.Simple(0, function()
+    -- run after the jobs have loaded
+    local models = sql.Query([[SELECT jobcmd, model, server FROM darkp_playermodels;]])
+
     for _, v in ipairs(models or {}) do
+        if v.server and v.server ~= game.GetIPAddress() then continue end
```
FPtje, line 56: This could be made somewhat more efficient by placing the server check in the query rather than in the code. That way, there is less data sent between SQLite and Lua, and you don't need to walk over as much.

Could you change the select query to be as follows?

```sql
    local models = sql.Query(string.format([[SELECT jobcmd, model FROM darkp_playermodels WHERE server = %s;]], sql.SQLStr(game.GetIPAddress())))
```

gamemode/modules/base/cl_jobmodels.lua
```diff
@@ -42,11 +44,22 @@ local function sendModels() -- run after the jobs have loaded
...
+
+        local job = DarkRP.getJobByCommand(v.jobcmd)
+        if job == nil or jobHasModel(job, v.model) == false then continue end
```
FPtje, line 59: Nit

```suggestion
        if job == nil or not jobHasModel(job, v.model) then continue end
```

FPtje (commented):
Thanks for the PR! I think it's a good idea to store preferred models per server. One tricky thing with this code is that the sqlite table needs to be compatible with both the old code _and_ the new code. This is because different servers will have different versions of the code.

I think that is covered with your current way of making the table and then altering it. Thanks! There are some requests, but overall I think this is quite a nice PR!

gamemode/modules/base/cl_jobmodels.lua
Be1zebub: better todo it on lua side, because it will use old rows without `server` column.  
if do this on the sql side, as you suggested, then we should add clearing table of old data without the server column, because they will be useless.

FPtje: Good point! Maybe it's still possible to place the condition in the query by using `WHERE server IS NULL OR server = %s`?

Be1zebub: lgtm

gamemode/modules/base/cl_jobmodels.lua
```diff
@@ -1,10 +1,31 @@
...
+-- for it, even though the code in this file always adds it. That is why it is
+-- added as a NULLable column.
 sql.Query([[CREATE TABLE IF NOT EXISTS darkp_playermodels(
     jobcmd VARCHAR(45) NOT NULL PRIMARY KEY,
-    model VARCHAR(140) NOT NULL
+    model VARCHAR(140) NOT NULL,
+    server TEXT NULL
 );]])
 
-sql.Query("ALTER TABLE darkp_playermodels ADD COLUMN IF NOT EXISTS server VARCHAR(21);")
```
FPtje: When testing this, I found out that the `IF NOT EXISTS` construct does not exist in SQLite. You have to `PRAGMA table_info(...)` to find out whether the column exists.

gamemode/modules/base/cl_jobmodels.lua
```diff
@@ -14,7 +37,7 @@ function DarkRP.setPreferredJobModel(teamNr, model)
     local job = RPExtraTeams[teamNr]
     if not job then return end
     preferredModels[job.command] = model
-    sql.Query(string.format([[REPLACE INTO darkp_playermodels VALUES(%s, %s);]], sql.SQLStr(job.command), sql.SQLStr(model)))
```
FPtje, line 17: There's another issue. After the table is migrated to have three columns, the old version of this query will fail with this error: `table darkp_playermodels has 3 columns but 2 values were supplied`.

Because SQLite errors are silent (the query just returns `false`, and you have to print `sql.LastError()` to get the error,  job models are silently no longer saved.

This would not have a problem if the query was `REPLACE INTO darkp_playermodels(jobcmd, model) VALUES(%s, %s);`, but this is not something we can fix in hindsight :thinking:

FPtje:
It looks like it's going to be very challenging to make this code backwards compatible: when players join servers with the old version, saving preferred job models would sadly break.

Since the old table has a spelling error (`darkp` instead of `darkrp`), maybe we can get away with creating a _new_ table `darkrp_playermodels` which has column `SERVER TEXT NOT NULL`, and then also get data from `darkp_playermodels` if it exists. What do you think?

gamemode/modules/base/cl_jobmodels.lua
```diff
@@ -1,9 +1,32 @@
 -- Create a table for the preferred playermodels
+--
+-- Note: the server column wasn't always there, so players may not have a value
+-- for it, even though the code in this file always adds it. That is why it is
+-- added as a NULLable column.
 sql.Query([[CREATE TABLE IF NOT EXISTS darkp_playermodels(
     jobcmd VARCHAR(45) NOT NULL PRIMARY KEY,
-    model VARCHAR(140) NOT NULL
+    model VARCHAR(140) NOT NULL,
+    server TEXT NULL
```
FPtje, line 9: One thing I realized as well: with `jobcmd` still being the primary key, two different servers with the same `jobcmd` would not be able to have two rows in the table.

FPtje:
Alright, I made a last change to the PR, to make an entirely _new_ table for the jobs. It will still use models from the old table, but those will be overwritten by the new table.

I also ran some manual tests to make sure it works right. This code is remarkably tricky to get working perfectly. Those invisible SQLite errors are a real stinger.

Let me know what you think. If this is alright with you I'll merge it :+1:

Be1zebub:
I thought about transferring the old data to a new table, but this method is also good.  
lgtm, but i think something like this would better:
```lua
-- pseudo code
if oldTable.exists() then
    for _, row in ipairs(oldTable.select()) do
        newTable.insert(row)
    end
    
    oldTable.delete()
end
```

FPtje:
That's an interesting idea, though I worry that it would break preferred models if players go back to a server running the old code. If the table gets deleted, servers running an older DarkRP will no longer be able to read it.

DarkRP has always had the issue of servers not caring to update, which means for some things, backwards compatibility needs to be kept indefinitely. By only reading from the table, nothing breaks when hopping between servers with the latest version and the old one.

Be1zebub:
makes sence.
is there a reason to store db on clientside and send it to the server, and not vice versa?

FPtje:
That's a really good question. I'm not quite sure 😅
Maybe my idea back then was that most servers would at least have the basic jobs in common, and that sharing the model setting across servers would then make sense. Nowadays it doesn't anymore though.

PR #3270, opened 2024-12-15
Delay removal of Players' DarkRP Vars

Srlion:
On server-side `EntityRemoved` the player entity is still valid, not having the delay makes some hooks not work with `ply:getDarkRPVar` inside it.

On client-side it will always fail because `PlayerDisconnected` and `EntityRemoved` are called on the same tick but the former runs first so when it reaches the client, any `EntityRemoved` hook that calls `ply:getDarkRPVar` will just return `nil`

5 seconds is overkill for client-side, but just in case gmod decides to delay EntityRemoved call and some bad networking happens. (future proof lol)

### DIFF
--- a/gamemode/modules/base/cl_entityvars.lua
+++ b/gamemode/modules/base/cl_entityvars.lua
@@ -81,8 +81,31 @@ timer.Simple(0, fp{RunConsoleCommand, "_sendDarkRPvars"})
 
 net.Receive("DarkRP_DarkRPVarDisconnect", function(len)
     local userID = net.ReadUInt(16)
-    timer.Simple(5, function()
+    local ply = Player(userID)
+
+    -- If the player is already gone, then immediately clear the data and move on.
+    if not IsValid(ply) then
         DarkRP.ClientsideDarkRPVars[userID] = nil
+        return
+    end
+    -- Otherwise, we need to wait until the player is actually removed
+    -- clientside. The net message may come in _much_ earlier than the message
+    -- that the player disconnected and should therefore be removed.
+    local hook_name = "darkrp_remove_darkrp_var_" .. userID
+
+    hook.Add("EntityRemoved", hook_name, function(ent)
+        -- NOTE: ent:UserID() will return -1 in this hook, so there is no use to
+        -- compare UserIDs. That also means that getting DarkRPVars in the
+        -- EntityRemoved hook is futile, as the lookup of -1 in
+        -- DarkRP.ClientsideDarkRPVars wil fail.
+        if ent ~= ply then return end
+        hook.Remove("EntityRemoved", hook_name)
+
+        -- Placing this in a timer allows for the rest of the hook runners to
+        -- still use the DarkRPVars until the entity is _really_ gone.
+        timer.Simple(0, function()
+            DarkRP.ClientsideDarkRPVars[userID] = nil
+        end)
     end)
 end)
 
--- a/gamemode/modules/base/cl_entityvars.lua
+++ b/gamemode/modules/base/cl_entityvars.lua
@@ -8,7 +8,15 @@ local pmeta = FindMetaTable("Player")
 -- enough to warrant optimizing. See https://github.com/FPtje/DarkRP/pull/3212
 local get_user_id = pmeta.UserID
 function pmeta:getDarkRPVar(var, fallback)
-    local vars = DarkRP.ClientsideDarkRPVars[get_user_id(self)]
+    local user_id = get_user_id(self)
+
+    -- Special case: when in the EntityRemoved hook, UserID returns -1. In this
+    -- case, hope that we still have a stored userID lying around somewhere.
+    if user_id == -1 then
+        user_id = self._darkrp_stored_user_id_for_entity_removed_hook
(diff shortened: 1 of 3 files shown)


FPtje:
For the server, wouldn't it be better to make the function based on the GAMEMODE function instead of the hook? That way it is executed last without timers

Srlion:
I don't think that it's worth the hassle of doing some code refactoring, timers will cause no issues at all and there will be no performance penalty.

FPtje:
The performance penalty is not what I'm worried about. Timers can solve some issues, but over the years I've also seen them cause the craziest other issues.
Especially the 5 second timer is quite arbitrary, and it would be nice to look for a different solution.

FPtje:
It turns out it can: using a timer in `EntityRemoved` causes the entity to become `NULL`. After all, the entity has been removed.

```
lua_run hook.Add("EntityRemoved", "timer_test", function(e) if not e:IsPlayer() then return end print("BEFORE TIMER:", e) timer.Simple(0, function() print("AFTER TIMER", e) end) end)

BEFORE TIMER:	Player [2][Bot01]
Dropped Bot01 from server (Kicked by (FPtje) Falco (test))
AFTER TIMER	Player [NULL]
```

The code in this PR "works", because it's not necessarily removing anything from the table at all.

Srlion:
Oops, the only issue that can could happen, is if you loop through the `ServerDarkRPVars` or `ServerPrivateDarkRPVars` tables without checking if player is valid or not. (DarkRP doesn't, but someone for weird reason could be doing it) But still the player NULL object is unique and can only be obtained if you saved it in a table.

I can change it to the end of `EntityRemoved` for both server and clien,t but that would require some refactoring if you don't mind.

FPtje:
It looks like `ply:GetUserID()` returns `-1` clientside when a player is removed. I think that's why the net message exists in the first place :thinking: :disappointed:

FPtje:
I have an alternative fix here:

https://github.com/FPtje/DarkRP/compare/fp/alternative-fix-for-hook-calling-problem?expand=1

It still won't allow getting DarkRPVars clientside in `EntityRemoved`, but I think changing that would have too big a risk to break addons.

Srlion:
> It looks like `ply:GetUserID()` returns `-1` clientside when a player is removed. I think that's why the net message exists in the first place 🤔 😞

https://github.com/Facepunch/garrysmod-issues/issues/6117
I solve it this way:
```lua
hook.Add("OnEntityCreated", "SAM.Commands.Arguments.Player", function(ent)
    if not isplayer(ent) then return end
    if not IsValid(ent) then return end

    ent.__sam_user_id = ent:UserID()
end)
```

You could assign the user_id to them as `__darkrp_user_id` till a fix is implemented

FPtje:
To be very honest, with this hook being there since pretty much the beginning of GMod 10 in 2006, I reckon this behavior has been known and unfixed for basically forever.

Maybe a different question to ask is whether you _really_ need your `getDarkRPVar` in `EntityRemoved`. What can you tell me about your use case?

Srlion:
No one reported that bug that's why it never got fixed, Rubat usually fixes small bugs like this.

I process command notifications on the client side, and some commands run asynchronously. A rare issue can occur if the player targeted by a command—or the admin who initiated the command—leaves the game before the command finishes. When the command task completes, the player’s entity may no longer exist, causing the player’s name to not appear in the notification sent to clients. To prevent this, I store each player’s name for 20 seconds after they disconnect using the EntityRemoved event. This ensures that if a command takes a significant amount of time to complete, the player’s name is still available, and the notification can display it correctly, avoiding this rare issue. (It's a really simple stupid use case lol)

It's not a big deal really if you don't implement the fix, it will just show their steam name lol

FPtje:
Thanks for the info. I think my solution relies less on timers, especially the 5 second one could go, so I would prefer that. 

Beyond that we would need a separate fix for the user ID thing. I'll take a look at that as well in a moment

FPtje:
That makes sense. Is it OK if I push the changes of `fp/alternative-fix-for-hook-calling-problem` before merging?

FPtje:
The fix seems to indeed be working:

```
] lua_run_cl hook.Add("EntityRemoved", "a", function(e) if not e:IsPlayer() then return end print("REMOVE", e:getDarkRPVar("money"), e._darkrp_stored_user_id_for_entity_removed_hook, type(e:UserID())) end)
] bot
] fadmin kick bot
REMOVE	500	9	number
```

What an adventure! Thanks! I've added some comments to this PR, so you get the credit for the solution :+1:

Srlion:
Wouldn't it be more viable to go with your solution to avoid using timers to avoid having 1% chance that someone loops through the table without checking if player is valid or not?

FPtje:
Ah, sorry. I meant to merge my solution into this PR

Srlion:
No worries, glad you didn't close it for such a simple thing lol. Great to see someone maintain such thing for a very long time with awesome work, keep it up!

FPtje:
Thanks! And thanks again for the solution. I've rebased my changes on top of your commits, so your commits end up in the history :+1:

PR #3272, opened 2024-12-25
localize weapon class

Aws0mee:
Noticed wep:GetClass() was called 3 times in a row while reading the weapon checker code.


FPtje:
Thanks!

PR #3273, opened 2024-12-29
Fix server console database checking

Seef0x:


FPtje:
Thanks for the PR! It looks like the IsValid check should already be covered by the IsSuperAdmin check and the getworld check.

I think I'm missing some context. What problem does this solve?

Seef0x:
Sorry for for lack of detail!

It fixes a lua error if u try to check db with server console

image

Seef0x:
Hmm
Display func will still send a net

FPtje:
Ah, thanks for the detail! Interesting, it looks like the `ply ~= game.GetWorld()` check doesn't work. That one should exit the if-statement before it gets to the superadmin check.

Seef0x:
World check isnt for singleplayer mod?

FPtje:
Did something change in console commands, that instead of ply being `Entity(0)`, it is now `NULL`?

Seef0x:
Hmm no
image

FPtje:
Yeah in con commands, the ply parameter

Seef0x:
Rip
image

FPtje:
Thanks!

PR #3278, opened 2025-03-24
Don't allow people to holster weapons given via giveweapon.

liondadev:
## PR Type
Exploit Fix

## Details
- Make all weapons given via `fadmin giveweapon` not holsterable
- Add new command, `/giveholsterableweapon`, that allows players to holster the weapon they're given


liondadev:
Wrong repo, mb :cry:

PR #3280, opened 2025-06-06
Optimization code

animerem:
Changed pairs to ipairs (where it was possible)

### DIFF
--- a/entities/entities/letter/init.lua
+++ b/entities/entities/letter/init.lua
@@ -48,7 +48,7 @@ function ENT:Use(ply)
             umsg.Vector(self:GetPos())
             local numParts = self.numPts
             umsg.Short(numParts)
-            for _, b in pairs(self.Parts) do umsg.String(b) end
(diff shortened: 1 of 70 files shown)


FPtje:
Thanks for the optimizations. The thing with pairs to ipairs is that it can be very hard to check that something _really_ is always indexed without holes from 1..n. A particularly sneaky example is the job table. It looks like this table can be looped over with ipairs, and on vanilla darkrp it _can_, but addons typically delete vanilla jobs and add their own, with the consequence of holes in the keys of the table.

When ipairs doesn't work, it doesn't throw an error, it just iterates over some, or even none of the elements of the table. This leads to very hard to debug problems. Things like only seeing _some_ jobs in the F4 menu, or messages affecting only some players.

When choosing ipairs, you gain a tiny bit of faster code, but you lose in the greatly increased risk of subtle silent failure. For me, that risk is too big. For that reason I only accept pairs to ipairs if a really good argument can be made that it is indeed safe.

PR #3289, opened 2025-10-18
Replacing a hardcoded value with an initialized variable for unknown PVars

Anteik:
Changing DARKRP_ID_BITS causes problems reading network messages


FPtje:
Thanks!

PR #3292, opened 2025-11-08
fix: job switch failing when target job slot is full

DrDuggi:
Corrects the order of team changes when switching jobs between players.

The current issue (example):
Mafiosi tries to /switchjob with the Mafiaboss
-> Fails because the job slot is already full

Now it works like this:
Mafiosi tries to /switchjob with the Mafiaboss:
-> Mafiaboss switches to Mafiosi first (freeing the slot), then Mafiosi switches to Mafiaboss. No more issues.


Kapppa:
You have just inverted the problem. Now it doesn't work the other way round.

PR #3293, opened 2025-11-25
Add money pile limit

GomeyDoge:
Same implementation as the cheque limit. Just a simple config option to prevent players spamming money piles all over the map.


FPtje:
Let's see what copilot says, just for the fun of it

PR #3295, opened 2026-02-13
OnChangedTeam called Clientside

loliebop:
OnChangedTeam is now triggered on the client for all players, even if you are not the one changing jobs.

Before this merge request, the OnChangedTeam hook was only called on the client for the player who actually changed their job.

### DIFF
--- a/gamemode/modules/base/cl_gamemode_functions.lua
+++ b/gamemode/modules/base/cl_gamemode_functions.lua
@@ -75,9 +75,11 @@ function GM:teamChanged(before, after)
 end
 
 local function OnChangedTeam(um)
-    local oldTeam, newTeam = um:ReadShort(), um:ReadShort()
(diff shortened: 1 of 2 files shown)


FPtje:
Thanks!

I have a bit of a worry that some addons actually depend on it being being for the local player only, and would then break because the hook fires for other teams 🤔

This is a bit of a downside of these things. Some things are super hard to change without breaking something somewhere.

Kapppa:
> Thanks!
> 
> I have a bit of a worry that some addons actually depend on it being being for the local player only, and would then break because the hook fires for other teams 🤔
> 
> This is a bit of a downside of these things. Some things are super hard to change without breaking something somewhere.

This is also meant to be SERVER only anyway. https://wiki.facepunch.com/gmod/GM:OnPlayerChangedTeam

FPtje:
Given previous experience of changing usermessages causing breakages, I'm going to be on the careful side and deny the pull request. I appreciate the effort, though!

PR #3299, opened 2026-04-25
Patch NaN Exploit

GomeyDoge:
This fixes an exploit where a player can have NaN money to bypass price checks.
Highly recommended to fix this immediately!

### DIFF
--- a/gamemode/modules/base/sh_util.lua
+++ b/gamemode/modules/base/sh_util.lua
@@ -369,7 +369,7 @@ Like tonumber, but makes sure it's an integer
 ---------------------------------------------------------------------------]]
 function DarkRP.toInt(value)
     value = tonumber(value)
-    return (value and value == value) and math.floor(value)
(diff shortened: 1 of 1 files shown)


GomeyDoge:
This exploit is being actively used on servers.

FPtje:
Thanks for the input! I don't think this is a good solution, though, because what should that function return instead when it gets a NaN? In your patch it will just return false, which is not quite helpful.

Perhaps the better question here is: where is that NaN coming from?

GomeyDoge:
I fixed it in toInt because in sv_money.lua addMoney function the check is:

```
    amount = DarkRP.toInt(amount)
    if not amount then return false end
```

The exploit caused by addons that use floats/doubles. I don't want to point fingers here but I can show you on discord if you want to see.

FPtje:
Fair enough. It looks like it's a parsing function anyway. Thanks!

PR #3302, opened 2026-05-22
Avoid repeated CurTime calls in AFK timer

PrinceOfCookies:
## Summary

Small optimization for the AFK timer.

This caches `CurTime()` once before looping over players, instead of calling it for every player in the loop.

Also checks bots before checking the AFK DarkRPVar.

## Changes

Old version:
- called `CurTime()` for every player
- checked `getDarkRPVar("AFK")` before `IsBot()`

New version:
- calls `CurTime()` once before the loop
- reuses the cached time for every player check
- checks `IsBot()` before `getDarkRPVar("AFK")`

Behavior should stay the same.

## Benchmark

Tested clientside with a fake player list:
- 128 fake players
- 100,000 timer runs
- timing done with `SysTime()`

| Run | Old | New |
|---:|---:|---:|
| 1 | 0.4467s | 0.0327s |
| 2 | 0.4580s | 0.0327s |
| 3 | 0.5357s | 0.0431s |
| 4 | 0.5033s | 0.0395s |

| Old avg | New avg | Improvement |
|---:|---:|---:|
| 0.4859s | 0.0370s | 92.4% faster |

### DIFF
--- a/gamemode/modules/afk/sv_afk.lua
+++ b/gamemode/modules/afk/sv_afk.lua
@@ -82,8 +82,10 @@ end
 hook.Add("KeyPress", "DarkRPKeyReleasedCheck", AFKTimer)
 
 local function KillAFKTimer()
+    local curTime = CurTime()
+
     for _, ply in ipairs(player.GetAll()) do
-        if ply.AFKDemote and CurTime() > ply.AFKDemote and not ply:getDarkRPVar("AFK") and not ply:IsBot() then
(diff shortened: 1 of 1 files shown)


FPtje (commented):
Thanks for the contribution! The PR explanation and code look good from a high level view. I'll take a more in depth look when I get back to my laptop.

FPtje (commented):
Looks like it works. Thanks!

PR #3304, opened 2026-07-08
Fix vote cleanup edge cases

ShiroKSH:
## Summary
- send the vote id with the same usermessage type that the client reads when closing vote panels
- make question destruction a no-op when the question has already gone away
- pass the actual bad argument into the Windows os.date workaround error helper

## Why
These are small cleanup paths that could currently either leave the client reading the wrong usermessage type or throw a Lua error while handling an already-removed question / invalid os.date argument.

## Checks
- git diff --check

### DIFF
--- a/gamemode/modules/voting/sv_votes.lua
+++ b/gamemode/modules/voting/sv_votes.lua
@@ -40,7 +40,7 @@ function Vote:handleEnd()
     win = win or self.yea > self.nay and 1 or self.nay > self.yea and -1 or 0
 
     umsg.Start("KillVoteVGUI", self:getFilter())
-        umsg.String(self.id)
+        umsg.Short(self.id)
     umsg.End()
 
     Votes[self.id] = nil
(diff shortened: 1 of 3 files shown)


gamemode/modules/voting/sv_votes.lua
```diff
@@ -40,7 +40,7 @@ function Vote:handleEnd()
     win = win or self.yea > self.nay and 1 or self.nay > self.yea and -1 or 0
 
     umsg.Start("KillVoteVGUI", self:getFilter())
-        umsg.String(self.id)
+        umsg.Short(self.id)
```
FPtje: I'm quite wary of changing usermessages. While a short would be way more efficient, addons other than DarkRP have come to depend on these usermessages. It wouldn't be the first time I broke addons by changing usermessages. In fact at some point I wanted to replace them all with net messages. Made a lot of people angry. This is a good example of [Hyrum's law](https://www.hyrumslaw.com)

Could you revert this particular change?

FPtje (changes requested):
Looking into it now, only one change request, the rest looks good!

gamemode/modules/voting/sv_votes.lua
ShiroKSH: > I'm quite wary of changing usermessages. While a short would be way more efficient, addons other than DarkRP have come to depend on these usermessages. It wouldn't be the first time I broke addons by changing usermessages. In fact at some point I wanted to replace them all with net messages. Made a lot of people angry. This is a good example of [Hyrum's law](https://www.hyrumslaw.com)
> 
> Could you revert this particular change?

done

PR #3305, opened 2026-07-08
Guard stale FAdmin action targets

ShiroKSH:
## Summary
- validate players/entities inside delayed FAdmin action callbacks before using them
- reject malformed TPToPos coordinates before constructing vectors
- avoid running scoreboard warrant/wanted commands after the selected player becomes invalid
- return early when UnRestrictWeapon is called without a weapon argument

## Why
Several admin actions keep player or entity references across timers or UI prompts. If the player disconnects, the target disappears, or the command is malformed before the callback runs, the old reference could be used directly and throw a Lua error.

## Checks
- git diff --check

### DIFF
--- a/gamemode/modules/fadmin/fadmin/playeractions/teleport/sv_init.lua
+++ b/gamemode/modules/fadmin/fadmin/playeractions/teleport/sv_init.lua
@@ -22,13 +22,21 @@ local function TPToPos(ply, cmd, args)
 
     local x, y, z = string.match(args[1] or "", "([-0-9\\.]+),%s?([-0-9\\.]+),%s?([-0-9\\.]+)")
     local vx, vy, vz = string.match(args[2] or "", "([-0-9\\.]+),%s?([-0-9\\.]+),%s?([-0-9\\.]+)")
-    local pos = Vector(tonumber(x), tonumber(y), tonumber(z))
-    local vel = Vector(tonumber(vx), tonumber(vy), tonumber(vz))
 
     if not args[1] or not x or not y or not z then return false end
+    x, y, z = tonumber(x), tonumber(y), tonumber(z)
+    if not x or not y or not z then return false end
+
+    local pos = Vector(x, y, z)
+    local vel
+
+    if vx and vy and vz then
+        vx, vy, vz = tonumber(vx), tonumber(vy), tonumber(vz)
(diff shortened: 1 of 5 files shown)


gamemode/modules/fadmin/fadmin/playeractions/teleport/sv_init.lua
```diff
@@ -113,7 +122,14 @@ local function Bring(ply, cmd, args)
 
             zapEffect(target)
 
-            if PHYSGUN then timer.Simple(0.5, function() target:Give("weapon_physgun") target:SelectWeapon("weapon_physgun") end) end
+            if PHYSGUN then
+                timer.Simple(0.5, function()
+                    if not IsValid(target) then return end
```
FPtje, line 127: Nice finds with these timers. Easy to forget to check validity 👍

gamemode/modules/fadmin/fadmin/restrict/sv_restrict.lua
```diff
@@ -31,6 +31,7 @@ end
 local function UnRestrictWeapons(ply, cmd, args)
     if not FAdmin.Access.PlayerHasPrivilege(ply, "Restrict") then FAdmin.Messages.SendMessage(ply, 5, "No access!") return false end
     local Weapon = args[1]
+    if not Weapon then return false end
```
FPtje: I feel like this should come with a message, like `FAdmin.Messages.SendMessage(ply, 5, "Please provide a weapon!")`. Could you add that?

FPtje (commented):
Thanks for the contribution!

gamemode/modules/fadmin/fadmin/restrict/sv_restrict.lua
ShiroKSH: > I feel like this should come with a message, like `FAdmin.Messages.SendMessage(ply, 5, "Please provide a weapon!")`. Could you add that?

Thanks! Added in https://github.com/FPtje/DarkRP/commit/898ff75a991cb6ac5e5caeccb706389c103d2896: [sv_restrict.lua](https://github.com/FPtje/DarkRP/blob/898ff75a/gamemode/modules/fadmin/fadmin/restrict/sv_restrict.lua#L34) now tells the caller that a weapon argument is required before returning.

PR #3306, opened 2026-07-08
Guard spawned entity failure paths

ShiroKSH:
## Summary
- validate spawned weapon/shipment entities before writing fields or placing them
- guard physics wake/mass calls when physics initialization fails
- preserve the original spawned weapon state when /makeshipment cannot create a crate
- skip delayed money printer explosions after the printer has already been removed
- avoid touching food owner counters after the owner entity is gone

## Why
These paths mostly depend on configurable entity classes/models. When creation or physics initialization fails, the code could continue into method calls on invalid entities or physics objects. The changes keep the normal success path the same and bail out earlier on failed creation paths.

## Checks
- git diff --check

### DIFF
--- a/entities/entities/food/init.lua
+++ b/entities/entities/food/init.lua
@@ -50,5 +50,7 @@ end
 
 function ENT:OnRemove()
     local ply = self:Getowning_ent()
+    if not IsValid(ply) then return end
+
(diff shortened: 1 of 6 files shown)


FPtje (approved):
LGTM! Will run one last test on these PRs, running them against my gmod instance before merging 👍

PR #3307, opened 2026-07-08
Guard tip jar and client stale state

ShiroKSH:
## Summary
- ignore invalid players before storing active tip jar donations
- remove active donation cleanup callbacks from the player they were attached to
- reject zero-value tip jar donations
- refresh cached LocalPlayer references before hitman and HUD drawing code uses them

## Why
Tip jar and HUD code can see stale player references during reconnects, entity removal, or malformed client updates. These guards avoid Lua errors and no-op updates without changing the normal donation or drawing paths.

## Checks
- git diff --check

### DIFF
--- a/gamemode/modules/hitmenu/cl_init.lua
+++ b/gamemode/modules/hitmenu/cl_init.lua
@@ -45,7 +45,9 @@ hook.Add("EntityRemoved", "hitmenu", function(ent)
 end)
 
 hook.Add("HUDPaint", "DrawHitOption", function()
-    localplayer = localplayer or LocalPlayer()
+    localplayer = IsValid(localplayer) and localplayer or LocalPlayer()
+    if not IsValid(localplayer) then return end
+
     hudText = hudText or GAMEMODE.Config.hudText
     local x, y
     local ply = localplayer:GetEyeTrace().Entity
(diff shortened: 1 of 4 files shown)


gamemode/modules/hitmenu/cl_init.lua
```diff
@@ -45,7 +45,9 @@ hook.Add("EntityRemoved", "hitmenu", function(ent)
 end)
 
 hook.Add("HUDPaint", "DrawHitOption", function()
-    localplayer = localplayer or LocalPlayer()
+    localplayer = IsValid(localplayer) and localplayer or LocalPlayer()
+    if not IsValid(localplayer) then return end
+
```
FPtje: LocalPlayer will never invalid. Could you revert the LocalPlayer IsValid checks in this file and cl_hud.lua?

FPtje (commented):
Also nice! Just one change request

gamemode/modules/hitmenu/cl_init.lua
ShiroKSH: > LocalPlayer will never invalid. Could you revert the LocalPlayer IsValid checks in this file and cl_hud.lua?

done <3

